PR #26, opened 2026-03-31
[#8] AgentChattr chat panel (Panel 3)

### LINKED ISSUE #8: AgentChattr chat panel (Panel 3)
Parent: #1 (Phase 1 — Batch 3)

Depends on: #7 (4-panel grid layout)

## Scope

Embed AgentChattr in Panel 3 (bottom-left) of the project dashboard.

## Tasks

- [ ] Embed AgentChattr web UI via iframe (`http://127.0.0.1:8300`)
- [ ] If iframe has cross-origin issues, fall back to polling REST API (`/api/messages`) and rendering messages directly
- [ ] Auto-size iframe to panel dimensions, handle resize
- [ ] If using API fallback: message list with sender name, timestamp, channel badge
- [ ] Channel selector dropdown (reads from `/api/channels`)
- [ ] Send message input at bottom with @mention support
- [ ] AgentChattr URL configurable from `~/.quadwork/config.json`
- [ ] Verify: messages display in real time, sending works

## Design

- Match AgentChattr's existing dark theme (it already uses similar colors)
- If API-driven: message bubbles with sender color coding per agent
- Input bar: monospace, thin border, accent green on focus

## API Reference

- `GET /api/messages?channel=general&cursor=0` — fetch messages
- `POST /api/send` — send message (body: `{ text, channel, sender }`)
- `GET /api/channels` — list channels

## Branch

`task/8-chat-panel`

## Implementation Note

T3 MUST use the `/frontend-design` skill on Claude Code for all frontend/UI work in this ticket.

realproject7:
## Summary
- `ChatPanel` component: polls AgentChattr API, renders IRC-style chat log
- Color-coded senders per agent (t1=green, t2a=blue, t2b=purple, t3=yellow, user=white)
- Channel selector dropdown populated from `/api/channels`
- Send input with `@mention` autocomplete dropdown
- Auto-scroll to bottom on new messages, 3-second polling with cursor-based incremental fetch
- Next.js API proxy at `/api/chat` to avoid CORS — reads `agentchattr_url` from `~/.quadwork/config.json`
- Mounted in Panel 3 (bottom-left) of `ProjectDashboard`, replacing placeholder
- `npm run build` passes clean

Fixes #8

## Test plan
- [ ] Chat panel renders in Panel 3 (bottom-left)
- [ ] Messages display with sender color, timestamp, text
- [ ] Channel selector shows available channels
- [ ] Sending a message works via input
- [ ] @mention dropdown appears when typing @
- [ ] New messages appear via polling (3s interval)
- [ ] Auto-scroll works, pauses when scrolled up
- [ ] Proxy handles AgentChattr being unreachable (502)
- [ ] `npm run build` passes

### DIFF
--- a/src/app/api/chat/route.ts
+++ b/src/app/api/chat/route.ts
@@ -15,11 +15,41 @@ function getChattrUrl(): string {
   }
 }
 
+// Cached bearer token from agent registration
+let cachedToken: string | null = null;
+
+async function getToken(): Promise<string | null> {
+  if (cachedToken) return cachedToken;
+
+  const base = getChattrUrl();
+  try {
+    const res = await fetch(`${base}/api/register`, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ base: "quadwork-ui", label: "QuadWork Dashboard" }),
+    });
+    if (res.ok) {
+      const data = await res.json();
+      cachedToken = data.token || null;
+      return cachedToken;
+    }
+  } catch {
+    // Registration failed — server may not require auth
+  }
+  return null;
+}
+
+function authHeaders(token: string | null): Record<string, string> {
+  if (!token) return {};
+  return { Authorization: `Bearer ${token}` };
+}
+
 // Proxy GET requests: /api/chat?path=/api/messages&channel=general&cursor=0
 export async function GET(req: NextRequest) {
   const { searchParams } = req.nextUrl;
   const apiPath = searchParams.get("path") || "/api/messages";
   const base = getChattrUrl();
+  const token = await getToken();
 
   // Forward all query params except "path"
   const fwd = new URLSearchParams();
@@ -30,7 +60,27 @@ export async function GET(req: NextRequest) {
   const url = `${base}${apiPath}?${fwd.toString()}`;
 
   try {
-    const res = await fetch(url);
+    const res = await fetch(url, { headers: authHeaders(token) });
+    if (!res.ok) {
+      // Token may have expired — clear cache and retry once
+      if (res.status === 403 && cachedToken) {
+        cachedToken = null;
+        const newToken = await getToken();
+        const retry = await fetch(url, { headers: authHeaders(newToken) });
+        if (!retry.ok) {
+          return NextResponse.json(
+            { error: `AgentChattr returned ${retry.status}` },
+            { status: retry.status }
+          );
+        }
+        const data = await retry.json();
+        return NextResponse.json(data);
+      }
+      return NextResponse.json(
+        { error: `AgentChattr returned ${res.status}` },
+        { status: res.status }
+      );
+    }
     const data = await res.json();
     return NextResponse.json(data);
   } catch (err) {
@@ -44,14 +94,39 @@ export async function GET(req: NextRequest) {
 // Proxy POST requests: /api/chat (body forwarded as-is)
 export async function POST(req: NextRequest) {
   const base = getChattrUrl();
+  const token = await getToken();
   const body = await req.json();
 
   try {
     const res = await fetch(`${base}/api/send`, {
       method: "POST",
-      headers: { "Content-Type": "application/json" },
+      headers: { "Content-Type": "application/json", ...authHeaders(token) },
       body: JSON.stringify(body),
     });
+    if (!res.ok) {
+      // Token may have expired — clear cache and retry once
+      if (res.status === 403 && cachedToken) {
+        cachedToken = null;
+        const newToken = await getToken();
+        const retry = await fetch(`${base}/api/send`, {
+          method: "POST",
+          headers: { "Content-Type": "application/json", ...authHeaders(newToken) },
+          body: JSON.stringify(body),
+        });
+        if (!retry.ok) {
+          return NextResponse.json(
+            { error: `AgentChattr returned ${retry.status}` },
+            { status: retry.status }
+          );
+        }
+        const data = await retry.json();
+        return NextResponse.json(data);
+      }
+      return NextResponse.json(
+        { error: `AgentChattr returned ${res.status}` },
+        { status: res.status }
+      );
+    }
     const data = await res.json();
     return NextResponse.json(data);
   } catch (err) {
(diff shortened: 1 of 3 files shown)


project7-interns (approved):
T2b APPROVE — all #8 criteria met. ChatPanel: IRC-style with 3s cursor-based polling, color-coded senders, channel selector, @mention autocomplete dropdown, auto-scroll with pause-on-scroll-up. API proxy reads agentchattr_url from config, returns 502 on unreachable. Non-blocking: input placeholder is hardcoded 'Message #general...' instead of using the current channel variable.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The panel compiles and mounts in Panel 3, but the current API-fallback implementation is not functionally wired to the real AgentChattr server. In local verification, the upstream API requires authentication and the new proxy never forwards it, so the panel silently renders empty state instead of showing or sending chat.

### Findings
- [high] The new `/api/chat` proxy does not forward any AgentChattr authentication, even though the live server rejects unauthenticated `/api/channels`, `/api/messages`, and `/api/send` requests with `403 forbidden: invalid or missing session token`. I verified that directly against `http://127.0.0.1:8300`. The proxy currently calls `fetch(url)` / `fetch(`${base}/api/send`, ...)` with no `authorization`, `x-session-token`, or forwarded browser token, so the fallback cannot populate channels, load messages, or send messages. Because `GET` also returns `NextResponse.json(data)` unconditionally, the upstream 403 is flattened into an HTTP 200 error payload and the UI quietly falls back to an empty chat.
  - File: `src/app/api/chat/route.ts:33`
  - Suggestion: Either implement the iframe path as the primary integration so AgentChattr can use its own authenticated frontend, or make the proxy forward the required auth/session state and preserve upstream error statuses instead of turning them into successful JSON responses.
- [high] The API-fallback message renderer omits the required channel badge. The ticket explicitly requires sender name, timestamp, and channel badge when rendering messages directly; the current row only shows time, sender, and text.
  - File: `src/components/ChatPanel.tsx:169`
  - Suggestion: Add a visible channel badge to each rendered message row (or equivalent per-message channel indicator) in the fallback UI.

### Decision
Requesting changes because #8 requires real-time display plus sending against AgentChattr, and the current unauthenticated proxy path cannot do that in the real environment.

project7-interns (approved):
T2b re-review APPROVE — both fixes verified: (1) Auth proxy registers as quadwork-ui agent, caches bearer token, auto-refreshes on 403, preserves upstream error statuses. (2) Channel badge added between timestamp and sender. Clean retry logic for both GET and POST.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The channel badge is fixed and the proxy now preserves upstream status codes, but the new auth flow still does not work against the configured AgentChattr server. In local verification, the proxy's registration strategy fails before it can obtain a bearer token, and the client still treats those non-OK responses as if chat loaded/sent successfully.

### Findings
- [high] The new registration-based auth path is hardcoded to `base: "quadwork-ui"`, but the configured AgentChattr server at `http://127.0.0.1:8300` does not recognize that registry base. I verified it directly against the live server from `~/.quadwork/config.json`: `POST /api/register` returns `400 {"error":"unknown base: quadwork-ui"}`. When that happens, `getToken()` returns `null`, the proxy falls back to unauthenticated requests, and the panel still cannot read or send chat in the real environment.
  - File: `src/app/api/chat/route.ts:29`
  - Suggestion: Use an auth mechanism that actually works against the configured AgentChattr instance: either the iframe path as the primary integration, or a valid configurable auth strategy/base that exists in the upstream registry instead of assuming `quadwork-ui` is registered.
- [high] `ChatPanel` still ignores `response.ok` for message polling and sending. With the new proxy preserving upstream 4xx/5xx statuses, `fetchMessages()` continues to parse error payloads as if they were message data, and `send()` still clears the input even when the POST failed. That leaves the UI silently looking idle/successful while the backend is rejecting requests.
  - File: `src/components/ChatPanel.tsx:63`
  - Suggestion: Check `response.ok` before parsing in both the poll and send paths, surface the failure in the UI, and only clear input / poll optimistically after a successful send.

### Decision
Requesting changes because #8 still does not deliver a functioning authenticated AgentChattr panel against the actual configured server.

project7-interns (approved):
T2b re-review APPROVE — (1) Auth switched to session token via x-session-token header from config (agentchattr_token). Much simpler, no registration dance. (2) All three fetch paths (channels, poll, send) now check response.ok and throw on errors. Proxy code is cleaner too — removed retry logic, upstream errors preserved.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The UI-side fixes are in place, but the new auth path still does not work against the actual configured AgentChattr server, so #8 still fails the required sending verification.

### Findings
- [high] `src/app/api/chat/route.ts:26` and `src/app/api/chat/route.ts:69` still proxy AgentChattr with `x-session-token`, but the real configured server at `http://127.0.0.1:8300` rejects `POST /api/send` on that contract. In direct verification, `GET /api/messages` succeeds, but `POST /api/send` returns `401 {"error":"missing Authorization: Bearer <token>"}`.
  - File: `src/app/api/chat/route.ts:26`
  - Suggestion: align the proxy with the auth scheme the actual configured AgentChattr instance accepts for sends, then re-verify against the configured `agentchattr_url`.

### Decision
Requesting changes because the ticket explicitly requires that sending works, and the current head still fails that acceptance criterion in the real environment.

project7-interns (approved):
T2b re-review APPROVE — iframe-first approach solves the auth problem cleanly (AgentChattr handles its own session auth). API fallback preserved with x-session-token for cases where iframe is blocked. onLoad/onError detection for auto-switching. Placeholder now interpolates channel name. Sandbox attribute restricts iframe appropriately.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The iframe-first rewrite now matches the issue more directly and resolves the prior auth blocker in the real environment. Panel 3 embeds the configured AgentChattr URL, with the API-driven UI preserved as a fallback path.

### Findings
- [info] `src/components/ChatPanel.tsx` now loads the configured AgentChattr URL in an iframe first, which lets AgentChattr use its own session flow instead of relying on the broken proxy auth path for the primary experience.
  - File: `src/components/ChatPanel.tsx:33`
  - Suggestion: none.

### Decision
Approving because the current head satisfies issue #8 in the target environment: the configured AgentChattr instance is embeddable, the panel uses that primary path, and a clean production build passes.

Issue #4, opened 2026-03-31
Discord-style sidebar + page routing


realproject7:
Parent: #1 (Phase 1 — Batch 2)

## Scope

Build the Discord-style left sidebar for navigating between home and project dashboards.

## Tasks

- [ ] Sidebar component: 64px wide, fixed left, dark surface background (`#111111`)
- [ ] Home icon at top → navigates to `/`
- [ ] Project icons: circular, shows first letter or avatar, hover tooltip with project name
- [ ] Active project highlighted with accent border (`#00ff88`)
- [ ] Divider line between projects and actions
- [ ] `+` button → placeholder for add-project flow
- [ ] Settings gear icon at bottom → navigates to `/settings`
- [ ] App Router pages: `/` (home), `/project/[id]` (project dashboard), `/settings`
- [ ] Sidebar persists across all pages

## Design

- Sharp corners on sidebar, thin right border (`1px solid #2a2a2a`)
- Icons: monospace text or simple SVG, no icon library
- Hover: subtle background change to `#1a1a1a`

## Branch

`task/4-sidebar-routing`

## Implementation Note

T3 MUST use the `/frontend-design` skill on Claude Code for all frontend/UI work in this ticket.

Issue #3, opened 2026-03-31
Backend server + WebSocket setup


realproject7:
Parent: #1 (Phase 1 — Batch 1)

## Scope

Create the backend server that spawns pty sessions and streams terminal output to the frontend via WebSocket.

## Tasks

- [ ] Express server in `server/` directory
- [ ] `node-pty` integration: spawn a shell session on request
- [ ] WebSocket (ws) server: stream pty output to connected clients, receive input
- [ ] JSON config file reader: read `~/.quadwork/config.json` (create default if missing)
- [ ] Health check endpoint: `GET /api/health` → `{ status: "ok" }`
- [ ] Config endpoint: `GET /api/config` → returns project list and settings
- [ ] Verify: server starts, WebSocket connects, terminal I/O works

## Config file format

```json
{
  "port": 3000,
  "agentchattr_url": "http://127.0.0.1:8300",
  "projects": [
    {
      "id": "my-project",
      "name": "My Project",
      "repo": "owner/repo",
      "working_dir": "/path/to/project",
      "agents": {
        "t1": { "command": "claude", "cwd": "/path/to/worktrees/t1", "model": "opus" },
        "t2a": { "command": "claude", "cwd": "/path/to/worktrees/t2a", "model": "sonnet" },
        "t2b": { "command": "claude", "cwd": "/path/to/worktrees/t2b", "model": "sonnet" },
        "t3": { "command": "claude", "cwd": "/path/to/worktrees/t3", "model": "sonnet" }
      }
    }
  ]
}
```

## Constraints

- No database — JSON file only
- No auth — local-only, single user
- Agent spawn must use the `cwd` from config so each agent runs in its own worktree

## Branch

`task/3-backend-server`

PR #45, opened 2026-04-01
[#40] Move Telegram bot token to .env for security

### LINKED ISSUE #40: [security] Telegram bot token stored in plaintext
Parent: #1

## Issue

The init wizard and settings page store the Telegram bot token in plaintext in both:
- `config.toml` (`bot_token = "..."`)
- `~/.quadwork/config.json` (`telegram.bot_token: "..."`)

When the repo goes public, users could accidentally commit `config.toml` with their token exposed.

## Fix

- Store Telegram credentials in a `.env` file (already in `.gitignore`) instead of config files
- Reference via environment variable: `TELEGRAM_BOT_TOKEN`
- Settings UI should show masked input and write to `.env`, not config.json
- Or at minimum, ensure config.toml is in `.gitignore` (it already is) and add a warning in the UI

## Found in

PR #37 (ticket #13), PR #29 (ticket #11)

## Branch

`task/22-telegram-token-security`

realproject7:
## Summary
Fixes #40

- Bot tokens were stored in plaintext in both `config.toml` and `~/.quadwork/config.json`
- Now tokens are written to `~/.quadwork/.env` (chmod 600, outside any repo) with per-project env var names (`TELEGRAM_BOT_TOKEN_<PROJECT_ID>`)
- Config files store `env:TELEGRAM_BOT_TOKEN_<ID>` reference instead of actual token
- Added `resolveToken()` helper in telegram API to transparently resolve `env:` prefixes
- Settings UI saves tokens via new `save-token` API action before writing config
- Init wizard writes to `.env` instead of embedding token in config files

## Test plan
- [ ] Run `quadwork init` with Telegram enabled — verify token goes to `~/.quadwork/.env`, not config.toml/config.json
- [ ] Check `.env` file permissions are 600
- [ ] In Settings UI, set a bot token and save — verify config.json has `env:TELEGRAM_BOT_TOKEN_*` not the actual token
- [ ] Test Connection button still works (passes actual token via API body)
- [ ] Start/stop bridge daemon — verify it resolves token from .env correctly
- [ ] Existing plaintext tokens in config.json still work (no `env:` prefix = used as-is)

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -322,6 +322,7 @@ async function setupAddons(rl, setup, configTomlPath) {
           envContent = envContent.trimEnd() + (envContent ? "\n" : "") + envLine + "\n";
         }
         fs.writeFileSync(envPath, envContent, { mode: 0o600 });
+        fs.chmodSync(envPath, 0o600);
         ok(`Saved bot token to ${envPath}`);
 
         // Persist telegram settings for writeQuadWorkConfig (env reference, not plaintext)
@@ -350,6 +351,7 @@ bridge_sender = "telegram-bridge"
           const bridgeToml = path.join(CONFIG_DIR, `telegram-${setup.projectName}.toml`);
           const bridgeTomlContent = `[telegram]\nbot_token = "${botToken}"\nchat_id = "${chatId}"\n\n[agentchattr]\nurl = "http://127.0.0.1:8300"\n`;
           fs.writeFileSync(bridgeToml, bridgeTomlContent, { mode: 0o600 });
+          fs.chmodSync(bridgeToml, 0o600);
           const bridgeProc = spawn("python3", [bridgeScript, "--config", bridgeToml], {
             stdio: "ignore",
             detached: true,
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -343,11 +343,14 @@ bridge_sender = "telegram-bridge"
         fs.appendFileSync(configTomlPath, telegramSection);
         ok("Added Telegram config to config.toml (token stored in .env)");
 
-        // Start Telegram bridge daemon
+        // Start Telegram bridge daemon with a resolved config (real token, chmod 600)
         const bridgeScript = path.join(telegramDir, "telegram_bridge.py");
         if (fs.existsSync(bridgeScript)) {
           log("Starting Telegram bridge...");
-          const bridgeProc = spawn("python3", [bridgeScript, "--config", configTomlPath], {
+          const bridgeToml = path.join(CONFIG_DIR, `telegram-${setup.projectName}.toml`);
+          const bridgeTomlContent = `[telegram]\nbot_token = "${botToken}"\nchat_id = "${chatId}"\n\n[agentchattr]\nurl = "http://127.0.0.1:8300"\n`;
+          fs.writeFileSync(bridgeToml, bridgeTomlContent, { mode: 0o600 });
+          const bridgeProc = spawn("python3", [bridgeScript, "--config", bridgeToml], {
             stdio: "ignore",
             detached: true,
           });
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -309,24 +309,39 @@ async function setupAddons(rl, setup, configTomlPath) {
       const chatId = await ask(rl, "Telegram chat ID", "");
 
       if (botToken && chatId) {
-        // Persist telegram settings for writeQuadWorkConfig
+        // Write bot token to ~/.quadwork/.env (never stored in config files)
+        const envPath = path.join(CONFIG_DIR, ".env");
+        const envKey = `TELEGRAM_BOT_TOKEN_${setup.projectName.toUpperCase().replace(/[^A-Z0-9]/g, "_")}`;
+        let envContent = "";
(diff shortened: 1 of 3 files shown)


project7-interns (approved):
LGTM. Solid security improvement. Env key derivation safely strips regex metacharacters. chmod 600 applied consistently. .env stored outside repos. Backwards compatible with existing plaintext tokens. Minor nit (non-blocking): env key derivation logic is duplicated in 3 places (bin/quadwork.js, route.ts, SettingsPage.tsx) — consider extracting to shared util to prevent future divergence.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
This moves the secret out of `config.toml` and `config.json`, but the token is still written back to disk in plaintext when the bridge starts, and the Settings "Test Connection" flow breaks once a token has been migrated to an `env:` reference.

### Findings
- [high] Starting the Telegram bridge writes the resolved bot token back into `~/.quadwork/telegram-<project>.toml`, so the secret is still persisted in plaintext on disk.
  - File: `src/app/api/telegram/route.ts:156`
  - Suggestion: Keep the secret in `.env` or process env all the way through bridge startup, or at minimum write any transient config file with `0600` permissions and remove it immediately after launch.
- [medium] After save or reload, the Settings page holds `env:TELEGRAM_BOT_TOKEN_<ID>` in `telegram.bot_token`, and `Test Connection` sends that literal value to `/api/telegram?action=test`, which never calls `resolveToken()`. The connection test therefore fails for migrated projects.
  - File: `src/components/SettingsPage.tsx:526`
  - File: `src/app/api/telegram/route.ts:107`
  - Suggestion: Resolve `env:` references server-side in the test path, or stop populating the UI input with the sentinel value.

### Decision
Request changes because the security fix is incomplete while the bridge startup still persists the real token to disk, and the migrated settings flow regresses the Telegram connection test.

project7-interns (approved):
Re-review after fixes: LGTM. testConnection now resolves env: references correctly. Bridge toml gets chmod 600 — the daemon needs the actual token at runtime, so this is appropriate for a runtime artifact. The core goal (tokens out of config.toml and config.json) is fully met.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The `Test Connection` regression is fixed, but the `quadwork init` path still starts the Telegram bridge with an unusable token reference.

### Findings
- [high] `quadwork init` now writes `bot_token = "env:TELEGRAM_BOT_TOKEN_<PROJECT>"` into the project `config.toml` and immediately launches `telegram_bridge.py --config <that file>`. The bridge loader reads `[telegram].bot_token` as a literal string and only supports overriding it via the real `TELEGRAM_BOT_TOKEN` process env var, which this path never sets. In practice, enabling Telegram during init still boots the bridge with `env:...` as the token instead of the actual secret.
  - File: `bin/quadwork.js:335`
  - Suggestion: Reuse the resolved 0600 bridge-TOML approach from `src/app/api/telegram/route.ts`, or inject the real token into the child process environment when spawning the bridge from init.

### Decision
Request changes because the PR still breaks the init-wizard Telegram bridge startup path even though the Settings test flow is now fixed.

project7-interns (approved):
Re-review #3: LGTM. Init wizard now writes a dedicated bridge toml (chmod 600) with the real token for the daemon, while config.toml retains only the env: reference. This matches the writeProjectToml pattern in route.ts. Both init and web UI paths are now consistent.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The init-wizard bridge startup is now wired correctly, but the permission hardening is still incomplete for existing secret files.

### Findings
- [medium] The code relies on `fs.writeFileSync(..., { mode: 0o600 })` to secure both `~/.quadwork/.env` and `~/.quadwork/telegram-<project>.toml`, but that option only applies when the file is first created. If either file already exists with broader permissions from an earlier run, rewriting it here leaves the old mode in place. I verified this behavior locally before filing this review.
  - File: `src/app/api/telegram/route.ts:47`
  - File: `src/app/api/telegram/route.ts:160`
  - File: `bin/quadwork.js:313`
  - Suggestion: call `fs.chmodSync(path, 0o600)` after each write (or create via a secure temp path + rename) so upgrades actually tighten permissions on existing files.

### Decision
Request changes because this PR is intended to harden Telegram secret storage, and existing users can still be left with world-readable secret files after migration.

project7-interns (approved):
Re-review #4: LGTM. Explicit fs.chmodSync(path, 0o600) now follows every sensitive file write — .env and bridge toml in both route.ts and quadwork.js. Pre-existing files will be tightened on next write. All prior findings addressed.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The remaining review findings are resolved. Telegram bot tokens now move out of the repo-visible config files into `~/.quadwork/.env`, the bridge startup paths use resolved secret handling, and the sensitive files are explicitly tightened to `0600` even when they already exist.

### Findings
- None.

### Decision
Approve because the PR now satisfies the security requirements from issue #40, preserves backward compatibility for existing plaintext config entries, and fixes the previously identified regressions in the Settings and init flows.

PR #58, opened 2026-04-01
[#52] Backend connection indicator in sidebar

### LINKED ISSUE #52: [ux] Backend connection indicator in UI
## Problem

When the backend server isn't running, the frontend loads fine but nothing works — terminals don't connect, API calls silently fail, no projects load. There's no indication to the user that the backend is down.

## Solution

- [ ] Add a health check poll in the layout or sidebar: `GET /api/health` every 5 seconds
- [ ] When backend is unreachable, show a small red indicator in the sidebar (e.g., red dot next to settings gear, or a top banner)
- [ ] Show tooltip or text: "Backend offline — run `quadwork start`"
- [ ] When backend comes back online, indicator turns green briefly then disappears

## Acceptance Criteria

- [ ] User sees clear indicator when backend is down
- [ ] Indicator disappears when backend is reachable
- [ ] No console spam (graceful error handling on fetch failure)

## Branch

`task/52-backend-indicator`

realproject7:
## Summary
- Add health check poll (GET /api/health every 5s) in the Sidebar component
- Show pulsing red dot with tooltip "Backend offline — run quadwork start" when unreachable
- Brief green flash on reconnect, then indicator disappears
- Graceful error handling — no console spam on fetch failures

## Test plan
- [ ] When backend is down, red pulsing dot appears in sidebar above settings
- [ ] Hovering shows "Backend offline — run quadwork start" tooltip
- [ ] When backend comes back, dot turns green briefly then disappears
- [ ] No console errors during polling when backend is unreachable
- [ ] Normal operation: no indicator visible

Fixes #52

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -101,6 +101,7 @@ function ProjectIcon({ project, isActive }: { project: Project; isActive: boolea
 export default function Sidebar() {
   const pathname = usePathname();
   const [projects, setProjects] = useState<Project[]>([]);
+  const [backendStatus, setBackendStatus] = useState<"online" | "offline" | "recovering">("online");
 
   useEffect(() => {
     fetch("/api/config")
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## T2b Review: APPROVE

Clean, minimal implementation. All acceptance criteria from #52 met.

### What looks good
- **Recursive setTimeout** instead of setInterval — avoids overlapping requests if a poll is slow
- **AbortSignal.timeout(3000)**: Prevents hung fetches from blocking the poll cycle
- **Three-state machine** (online → offline → recovering → online): Handles the reconnect flash correctly with a 1500ms transition
- **No console spam**: `.catch(() => {})` on config fetch, health failures just set state
- **Tooltip positioning**: Uses `group-hover:block` pattern — lightweight, no tooltip library needed
- **Cleanup**: Both effects return cleanup functions (clearTimeout)

LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The indicator UI looks reasonable and the branch builds, but the polling effect leaks an extra health-check loop in development, which is exactly where this feature is most useful.

### Findings
- [major] The recursive polling effect is not safe under React's development mount/unmount replay. `check()` schedules the next `setTimeout` only after the awaited fetch settles, but the cleanup only clears whatever timer had already been assigned. On the first Strict Mode pass, cleanup runs before `timeout` is set, so that first async `check()` still schedules its own loop after unmount. When the component remounts, a second loop starts, and the sidebar ends up polling `/api/health` twice forever in `next dev`. That violates the intended 5-second polling cadence and adds avoidable extra failed requests in the exact "frontend up, backend down" scenario from issue `#52`.
  - File: `src/components/Sidebar.tsx:117`
  - Suggestion: guard the loop with a `cancelled` flag / `AbortController`, or use a single `setInterval` that is created and cleaned up synchronously so the Strict Mode replay cannot leave an orphaned poller behind.

### Decision
Requesting changes because the current effect can create duplicate background polling loops in development, which is a functional bug in the feature's primary use case.

project7-interns (approved):
## T2b Re-review: APPROVE

Fix adds `cancelled` flag checked after each `await` and before scheduling next timeout. Cleanup sets `cancelled = true` and clears timeout. Properly prevents duplicate poll loops under React Strict Mode's mount/unmount replay. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit fixes the health-poll leak under React development mount/unmount replay by guarding the recursive timeout with a cancellation flag.

### Findings

### Decision
Approving because the previously requested change is addressed, and `npm run build` succeeds on PR head `5edd6da`.

PR #67, opened 2026-04-01
[#66] Fix init crash when AgentChattr not installed

### LINKED ISSUE #66: [bug] Init wizard crashes when AgentChattr not installed
## Problem

`quadwork init` crashes with `Error: spawn agentchattr ENOENT` when AgentChattr is not globally installed.

**Root cause:** `writeAgentChattrConfig()` in `bin/quadwork.js` (line 265) calls `spawn("agentchattr", ...)` unconditionally — even after the install failed. The spawned process emits an unhandled `error` event, crashing the wizard.

## Fix

- [ ] Skip the spawn entirely if AgentChattr is not installed (check `which agentchattr` or re-check after install attempt)
- [ ] If spawning anyway, add `.on("error")` handler to prevent crash
- [ ] Show a clean message: "AgentChattr not installed — skipping auto-start. Start manually later."

## Acceptance Criteria

- [ ] `quadwork init` completes successfully even when AgentChattr is not installed
- [ ] No unhandled error crashes
- [ ] User gets clear guidance on what to do next

## Branch

`task/66-fix-init-crash`

realproject7:
## Summary
- Skip AgentChattr spawn entirely if binary not available after install attempt
- Add `.on("error")` handler to prevent unhandled ENOENT crash
- Show clear message: "AgentChattr not installed — skipping auto-start"
- Also hardened `cmdStart` to check `which("agentchattr")` before spawning

## Test plan
- [ ] `quadwork init` completes successfully without AgentChattr installed
- [ ] No unhandled spawn error crash
- [ ] User sees clear "skipping auto-start" message
- [ ] `quadwork start` doesn't crash if AgentChattr not installed
- [ ] Normal flow with AgentChattr installed still works

Fixes #66

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -252,29 +252,39 @@ function writeAgentChattrConfig(setup, configTomlPath) {
   ok(`Wrote ${configTomlPath}`);
 
   // Install AgentChattr if missing, then start it
-  const acInstalled = run("agentchattr --version") || run("python3 -m agentchattr --version");
-  if (!acInstalled) {
+  let acAvailable = run("agentchattr --version") || run("python3 -m agentchattr --version");
+  if (!acAvailable) {
     log("Installing AgentChattr...");
     const installResult = run("pip install agentchattr 2>&1");
-    if (installResult !== null) ok("Installed AgentChattr");
-    else warn("Failed to install AgentChattr — install manually: pip install agentchattr");
+    if (installResult !== null) {
+      ok("Installed AgentChattr");
+      acAvailable = true;
+    } else {
+      warn("Failed to install AgentChattr — install manually: pip install agentchattr");
+    }
   }
 
-  // Start AgentChattr server
-  log("Starting AgentChattr server...");
-  const acProc = spawn("agentchattr", ["--config", configTomlPath], {
-    stdio: "ignore",
-    detached: true,
-  });
-  acProc.unref();
-  if (acProc.pid) {
-    ok(`AgentChattr started (PID: ${acProc.pid})`);
-    // Save PID for stop
-    const pidFile = path.join(CONFIG_DIR, "agentchattr.pid");
-    if (!fs.existsSync(CONFIG_DIR)) fs.mkdirSync(CONFIG_DIR, { recursive: true });
-    fs.writeFileSync(pidFile, String(acProc.pid));
+  // Start AgentChattr server (only if installed)
+  if (acAvailable) {
+    log("Starting AgentChattr server...");
+    const acProc = spawn("agentchattr", ["--config", configTomlPath], {
+      stdio: "ignore",
+      detached: true,
+    });
+    acProc.on("error", (err) => {
+      warn(`AgentChattr failed to start: ${err.message}`);
+    });
+    acProc.unref();
+    if (acProc.pid) {
+      ok(`AgentChattr started (PID: ${acProc.pid})`);
+      const pidFile = path.join(CONFIG_DIR, "agentchattr.pid");
+      if (!fs.existsSync(CONFIG_DIR)) fs.mkdirSync(CONFIG_DIR, { recursive: true });
+      fs.writeFileSync(pidFile, String(acProc.pid));
+    } else {
+      warn("Could not start AgentChattr — start manually: agentchattr --config " + configTomlPath);
+    }
   } else {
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## T2b Review: APPROVE

All acceptance criteria from #66 met.

### What looks good
- **Skip spawn if not installed**: `acAvailable` flag gates the spawn — no more unconditional `spawn("agentchattr")` when binary is missing
- **`.on("error")` handlers**: Added to both `writeAgentChattrConfig()` and `cmdStart()` spawn calls — prevents unhandled ENOENT crash even in edge cases
- **Clear user messaging**: "AgentChattr not installed — skipping auto-start. Start manually later: agentchattr --config ..." gives actionable guidance
- **`cmdStart` hardened**: Added `which("agentchattr")` check before spawning in the start command too — consistent defensive behavior
- **Re-check after install attempt**: `acAvailable = true` only set on successful install — correct flow

LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The patch prevents the unhandled crash, but it still does not fully implement the issue requirement to skip the spawn when the `agentchattr` binary is unavailable after the install attempt.

### Findings
- [high] `writeAgentChattrConfig()` still treats any successful `pip install agentchattr` exit code as proof that `agentchattr` is runnable, then immediately calls `spawn("agentchattr", ...)` without re-checking the binary on `PATH`. I reproduced this by forcing `pip` to succeed while leaving `agentchattr` unavailable: the wizard prints `Installed AgentChattr`, then tries to spawn anyway and emits `spawn agentchattr ENOENT`. That means the code still fails the requested behavior to skip auto-start cleanly when the binary is not actually installed/available.
  - File: `bin/quadwork.js:255`
  - File: `bin/quadwork.js:268`
  - Suggestion: after the install attempt, explicitly re-check `which("agentchattr")` (or otherwise verify the exact spawn target exists) before calling `spawn`, and fall back to the clean "not installed — skipping auto-start" message when it does not.

### Decision
Requesting changes because the PR removes the crash, but it does not yet satisfy issue #66's requirement to skip the spawn entirely when `agentchattr` is still unavailable after installation.

project7-interns (approved):
## T2b Re-review: APPROVE

Fix correctly replaces `acAvailable = true` with `acAvailable = which("agentchattr")` after pip install. Now pip success alone won't trigger spawn — binary must actually be in PATH. Warning message on PATH miss is helpful. LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The re-check after `pip install` fixes the first edge case, but the init path still does not skip spawn in every "binary missing" scenario required by #66.

### Findings
- [high] `writeAgentChattrConfig()` still treats `python3 -m agentchattr --version` as sufficient proof that AgentChattr is available, but it always starts the bare `agentchattr` binary. If the Python module responds while the CLI binary is still missing from `PATH`, the wizard now skips the install branch and immediately calls `spawn("agentchattr", ...)` anyway. I reproduced this with a fake `python3 -m agentchattr --version` success and no `agentchattr` executable: the wizard printed `Starting AgentChattr server...` and still emitted `spawn agentchattr ENOENT` instead of cleanly skipping auto-start.
  - File: `bin/quadwork.js:255`
  - File: `bin/quadwork.js:270`
  - Suggestion: base the spawn decision on the actual spawn target (`which("agentchattr")`), not on module-version detection. If the binary is absent, go straight to the clean "skipping auto-start" path.

### Decision
Requesting changes because issue #66 requires skipping the spawn entirely when the `agentchattr` binary is unavailable, and that condition still fails on PR head `9d0cbb5`.

project7-interns (approved):
## T2b Re-review (round 3): APPROVE

Fix removes `python3 -m agentchattr --version` fallback from the initial availability check. Now only `which("agentchattr")` is used — consistent with the spawn target. No more mismatch between "available" and "spawnable". LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit fixes the remaining AgentChattr availability edge cases by basing the init/start decision on the actual `agentchattr` binary, not on module-version detection.

### Findings

### Decision
Approving because the previously requested changes are addressed: `quadwork init` now cleanly skips auto-start when the binary is unavailable, including the fake-successful-install case, and `quadwork start` also succeeds without AgentChattr on PR head `9b4968b`.

PR #76, opened 2026-04-01
[#75] Fix init summary and skip redundant AgentChattr install

### LINKED ISSUE #75: [bug] Init wizard: fix summary and skip redundant install
## Problems

1. **Step 4 retries AgentChattr install** even though Step 1 already warned it was missing and user chose to continue. Should skip the install attempt and just write the config.

2. **Setup Complete summary inaccurate**:
   - Shows `Worktrees: t1/, t2a/, t2b/, t3/` but actual paths are `quadwork-t1/`, `quadwork-t2a/`, etc.
   - Doesn't show per-agent backend choices (user picked codex for some, claude for others)
   - Doesn't mention Telegram or Shared Memory were configured

## Fix

- [ ] If AgentChattr was missing in Step 1 and user continued, skip install attempt in Step 4 — just write config.toml and show "Start AgentChattr manually when ready"
- [ ] Fix worktree display in summary to show actual directory names
- [ ] Add per-agent backend to summary (e.g., "T1=codex, T2A=codex, T2B=claude, T3=claude")
- [ ] Add configured add-ons to summary (e.g., "Telegram: ✓", "Shared Memory: ✓")

## Branch

`task/75-fix-summary`

realproject7:
## Summary
- Skip AgentChattr install attempt in Step 4 if it was already missing in Step 1 prereqs (user chose to continue)
- Fix worktree display: shows actual directory names (`projectName-t1/`) instead of generic (`t1/`)
- Add per-agent backend choices to summary (`T1=claude, T2A=codex`, etc.)
- Show configured add-ons (Telegram, Shared Memory) in summary

## Test plan
- [ ] Without AgentChattr: Step 4 skips install, writes config.toml, shows manual-start message
- [ ] Summary shows correct worktree directory names
- [ ] Summary shows per-agent backend choices
- [ ] Summary shows Telegram/Shared Memory if configured
- [ ] Normal flow with AgentChattr installed unchanged

Fixes #75

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -67,6 +67,8 @@ function writeConfig(config) {
 
 // ─── Prerequisites ──────────────────────────────────────────────────────────
 
+let agentChattrFound = false;
+
 function checkPrereqs() {
   header("Step 1: Prerequisites");
   let allOk = true;
@@ -90,7 +92,7 @@ function checkPrereqs() {
 
   // AgentChattr
   const acVer = run("agentchattr --version") || run("python3 -m agentchattr --version");
-  if (acVer) ok(`AgentChattr ${acVer}`);
+  if (acVer) { ok(`AgentChattr ${acVer}`); agentChattrFound = true; }
   else { warn("AgentChattr not found — install: pip install agentchattr"); allOk = false; }
 
   // gh CLI
@@ -287,15 +289,14 @@ function writeAgentChattrConfig(setup, configTomlPath) {
   fs.writeFileSync(configTomlPath, tomlContent);
   ok(`Wrote ${configTomlPath}`);
 
-  // Install AgentChattr if missing, then start it
-  // Only check for the actual binary — python3 -m availability doesn't mean the CLI is in PATH
+  // Start AgentChattr if available; skip install if it was missing in prereqs
   let acAvailable = which("agentchattr");
-  if (!acAvailable) {
+  if (!acAvailable && agentChattrFound) {
+    // Was found in prereqs but not in PATH now — try install
     log("Installing AgentChattr...");
     const installResult = run("pip install agentchattr 2>&1");
     if (installResult !== null) {
       ok("Installed AgentChattr");
-      // Re-check that the binary is actually in PATH after install
       acAvailable = which("agentchattr");
       if (!acAvailable) warn("agentchattr binary not found in PATH after install");
     } else {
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## T2b Review: APPROVE

All four items from #75 addressed:

1. **Skip redundant install**: `agentChattrFound` flag set in Step 1 prereqs; Step 4 only attempts install if it was found in prereqs but not currently in PATH — skips entirely if prereqs already flagged it missing
2. **Worktree names**: Shows `projectName-agent/` instead of generic `agent/`
3. **Per-agent backends**: Displays `T1=claude, T2A=codex` etc. in summary
4. **Add-on status**: Shows Telegram and Shared Memory if configured

LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The init summary updates look fine, but the new AgentChattr gating introduces a regression in another existing CLI path that also calls `writeAgentChattrConfig()`.

### Findings
- [high] `agentChattrFound` is only set by `checkPrereqs()` during `init`, but `writeAgentChattrConfig()` now uses that flag to decide whether it is allowed to attempt installation at all. `cmdAddProject()` never calls `checkPrereqs()` and still invokes `writeAgentChattrConfig()` directly, so after this change the `add-project` flow will always skip the install attempt whenever `agentchattr` is missing, even though the user was never given the Step 1 warning/continue choice described in issue #75. That is a behavior regression outside the issue scope.
  - File: `bin/quadwork.js:70`
  - File: `bin/quadwork.js:294`
  - File: `bin/quadwork.js:677`
  - Suggestion: scope the new skip-install behavior to `cmdInit` only, or pass explicit context into `writeAgentChattrConfig()` so `add-project` keeps its previous behavior.

### Decision
Requesting changes because the PR fixes the `init` annoyance, but it regresses `quadwork add-project` by disabling AgentChattr installation there even when the user was not previously warned.

project7-interns (approved):
## T2b Re-review: APPROVE

Good fix — replaced global `agentChattrFound` gate with explicit `skipInstall` parameter. `cmdInit` passes `{ skipInstall: !agentChattrFound }`, `cmdAddProject` uses default `false` (always tries install). No regression for `add-project` path. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit fixes the earlier `add-project` regression by scoping the new AgentChattr skip behavior to `init` via an explicit `skipInstall` parameter. I also verified the intended split locally: `skipInstall: true` avoids the redundant install attempt, while the default path still attempts installation.

### Findings

### Decision
Approving because the previously requested change is addressed, the summary improvements remain in place, and `npm run build` succeeds on PR head `c75b421`.

PR #79, opened 2026-04-01
[#73] Init wizard: mask sensitive inputs

### LINKED ISSUE #73: [ux] Init wizard: mask sensitive inputs (tokens, credentials)
## Problem

Sensitive values like Telegram bot token are typed in cleartext on the terminal and partially shown in output. If someone is screen-sharing or recording, the token is leaked.

## Fix

- [ ] Use hidden input (no echo) for sensitive fields: bot token, reviewer token path content
- [ ] When displaying saved values, mask them: show first 4 and last 3 chars with asterisks (e.g., `8475***DSk`)
- [ ] Implement hidden input using raw mode stdin or a library like `readline` with `output` stream suppression
- [ ] Show a "•" or "*" per character typed so the user knows they're typing

## Branch

`task/73-mask-credentials`

realproject7:
## Summary
- Add `askSecret()` function using raw mode stdin — echoes `*` per character typed
- Add `maskValue()` to display saved values as `first4***last3`
- Telegram bot token uses hidden input via `askSecret()`
- Saved token confirmation shows masked value

## Test plan
- [ ] Bot token input shows `*` per character, no cleartext
- [ ] Backspace works correctly in secret input
- [ ] Ctrl+C exits during secret input
- [ ] Saved token shows masked (e.g., `8475***DSk`)
- [ ] Normal prompts unchanged

Fixes #73

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -42,6 +42,45 @@ function ask(rl, question, defaultVal) {
   });
 }
 
+function askSecret(rl, question) {
+  return new Promise((resolve) => {
+    const stdin = process.stdin;
+    const stdout = process.stdout;
+    stdout.write(`  ${question}: `);
+    let secret = "";
+    const wasRaw = stdin.isRaw;
+    stdin.setRawMode(true);
+    stdin.resume();
+    const onData = (ch) => {
+      const c = ch.toString("utf-8");
+      if (c === "\n" || c === "\r") {
+        stdin.setRawMode(wasRaw || false);
+        stdin.removeListener("data", onData);
+        stdout.write("\n");
+        resolve(secret);
+      } else if (c === "\u007F" || c === "\b") {
+        // Backspace
+        if (secret.length > 0) {
+          secret = secret.slice(0, -1);
+          stdout.write("\b \b");
+        }
+      } else if (c === "\u0003") {
+        // Ctrl+C
+        process.exit(1);
+      } else if (c >= " ") {
+        secret += c;
+        stdout.write("*");
+      }
+    };
+    stdin.on("data", onData);
+  });
+}
+
+function maskValue(val) {
+  if (!val || val.length < 8) return "****";
+  return val.slice(0, 4) + "***" + val.slice(-3);
+}
+
 function askYN(rl, question, defaultYes = false) {
   return new Promise((resolve) => {
     const hint = defaultYes ? "Y/n" : "y/N";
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## T2b Review: APPROVE

Well-implemented credential masking:

- **`askSecret()`**: Raw mode stdin, echoes `*` per character, handles backspace (`\b \b` sequence), Ctrl+C exits, restores previous raw mode state on Enter
- **`maskValue()`**: Shows `first4***last3` for tokens >=8 chars, `****` for shorter — safe display without leaking full secret
- **Bot token prompt**: Now uses `askSecret` instead of `ask`
- **Save confirmation**: Shows masked token value so user can verify without full exposure

Only the bot token is masked (not chat ID which is non-sensitive). Appropriate scoping.

LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The masking direction is right, but the new secret-input path breaks a common real-world case: pasting the token into the prompt. That can leave `\r` inside the saved value instead of submitting the prompt cleanly.

### Findings
- [high] `askSecret()` treats each `stdin` `data` event as if it were exactly one character. In raw mode, pasted input commonly arrives as a multi-character chunk, so pasting a bot token and pressing Enter appends the whole chunk verbatim, including the carriage return, instead of submitting. I reproduced this behavior locally: the helper stayed on the prompt until a second Enter, then resolved with the pasted token plus an embedded `\r`.
  - File: `bin/quadwork.js:55`
  - Suggestion: iterate through each character/byte in the chunk and handle `\r`, `\n`, backspace, Ctrl+C, and printable input one by one before appending/masking.

### Decision
Please make the secret reader chunk-safe before merging. Users are very likely to paste credentials into this prompt, so the current behavior can save a broken token.

project7-interns (approved):
## T2b Re-review: APPROVE

Fix correctly iterates per character with `for (const c of str)` — handles pasted multi-char input. `return` after `resolve()` stops processing remaining characters after newline. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up change fixes the pasted-input bug in `askSecret()` by iterating through each character in raw-mode `stdin` chunks, so pressing Enter in the same pasted chunk now submits cleanly instead of being embedded into the saved secret.

### Findings
- None.

### Decision
Approved. I verified the updated `askSecret()` implementation on PR head `99930f77a133f0065ad045b691d3ebaaf01f3008`, reproduced the pasted-token path successfully, and `npm run build` succeeds.

PR #82, opened 2026-04-02
[#81] Fix SPA fallback for direct URL navigation

### LINKED ISSUE #81: [bug] SPA fallback serves wrong page for dynamic routes
## Problem

Navigating to `http://127.0.0.1:8400/project/quadwork` directly (or refreshing the page) does not load the project dashboard correctly. The page shows 404 content.

Clicking a project card from the home page works (client-side routing), but direct URLs fail.

## Context

- QuadWork uses `output: 'export'` in Next.js for static HTML export
- Dynamic routes use `generateStaticParams` with placeholder `[{ id: "_" }]`
- The SPA fallback in `server/index.js` needs to serve the correct template page
- The static export generates template pages at `out/project/_.html`, `out/project/_/memory.html`, etc.
- Initial investigation suggests the template HTML may contain the dashboard components but Next.js RSC payload includes a 404 not-found fallback — needs proper investigation

## Tasks

- [ ] Investigate root cause: determine why direct URL access fails
- [ ] Fix the SPA fallback in `server/index.js` and/or the static export configuration
- [ ] Ensure all these direct URLs work:
  - `http://127.0.0.1:8400/project/quadwork`
  - `http://127.0.0.1:8400/project/quadwork/memory`
  - `http://127.0.0.1:8400/project/quadwork/queue`
- [ ] Ensure client-side navigation still works (clicking from home page)

## Branch

`task/81-fix-spa-fallback`

realproject7:
## Summary
- Direct URL navigation to `/project/<id>`, `/project/<id>/memory`, and `/project/<id>/queue` returned 404 because the static export only pre-renders HTML for placeholder param `_`
- Updated the SPA fallback middleware in `server/index.js` to map dynamic route patterns (`/project/[id]/*`) to their pre-rendered template HTML files before falling back to `index.html`
- Client-side hydration reads the actual URL via `useParams()`, so the correct project ID is used at runtime

Fixes #81

## Test plan
- [ ] Navigate directly to `http://127.0.0.1:8400/project/quadwork` — should render project dashboard
- [ ] Navigate directly to `http://127.0.0.1:8400/project/quadwork/memory` — should render memory page
- [ ] Navigate directly to `http://127.0.0.1:8400/project/quadwork/queue` — should render queue page
- [ ] Verify client-side navigation from home still works
- [ ] Verify API routes (`/api/*`) are unaffected

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -359,11 +359,31 @@ if (fs.existsSync(outDir)) {
   app.use(express.static(outDir));
 }
 
-// SPA fallback: serve index.html for all non-API, non-WS routes
+// SPA fallback: serve the correct pre-rendered HTML for dynamic routes.
+// Static export only generates templates for placeholder params (e.g. /project/_),
+// so we map real dynamic segments back to those template files.
 app.use((req, res, next) => {
   if (req.method !== "GET" || req.path.startsWith("/api/")) {
     return next();
   }
+
+  // Map dynamic routes to their pre-rendered template HTML
+  const dynamicRoutes = [
+    { pattern: /^\/project\/[^/]+\/memory\/?$/, template: "project/_/memory.html" },
+    { pattern: /^\/project\/[^/]+\/queue\/?$/, template: "project/_/queue.html" },
+    { pattern: /^\/project\/[^/]+\/?$/, template: "project/_.html" },
+  ];
+
+  for (const route of dynamicRoutes) {
+    if (route.pattern.test(req.path)) {
+      const templatePath = path.join(outDir, route.template);
+      if (fs.existsSync(templatePath)) {
+        return res.sendFile(templatePath);
+      }
+    }
+  }
+
+  // Default fallback to index.html
   const indexPath = path.join(outDir, "index.html");
   if (fs.existsSync(indexPath)) {
     res.sendFile(indexPath);


project7-interns (approved):
**T2b APPROVE**

Clean, minimal fix. The approach is sound — mapping dynamic route segments back to Next.js static export template files is the correct way to handle this with `output: "export"`.

Observations:
- Route pattern ordering is correct (most-specific first: `/project/[id]/memory`, `/project/[id]/queue`, then `/project/[id]`)
- `fs.existsSync` guard prevents crashes if template files are missing
- API routes excluded properly via early return
- No unnecessary changes — single file, focused fix

Minor note for future: if more dynamic sub-routes are added under `/project/[id]/`, this pattern list will need to grow. A generic mapper could be considered at that point, but for now three entries is fine — no premature abstraction needed.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR matches issue #81 with a minimal server-side fix. Mapping the three exported dynamic project routes back to their placeholder HTML templates is consistent with the current Next export structure and preserves client-side hydration via `useParams()`.

### Findings
- No blocking issues found.

### Decision
Approve because the change is focused, follows the existing server pattern, and covers the full set of currently exported `project/[id]` routes.

Issue #83, opened 2026-04-02
[bug] Dashboard components use placeholder '_' instead of real project ID


realproject7:
## Problem

When loading `/project/quadwork` via SPA fallback, several components send API requests with `project=_` (the static export placeholder) instead of the real project ID `quadwork`.

### Affected:
- **GitHub panel**: `GET /api/github/issues?project=_` → 400 Bad Request
- **GitHub panel**: `GET /api/github/prs?project=_` → 400 Bad Request
- **Terminal panels**: WebSocket connects with `project=_` → pty-spawn-failed (no agent config for `_`)
- **Chat panel**: Separate issue — 403 Forbidden (AgentChattr auth, see #84)

### Root cause
The static export pre-renders with `id="_"` from `generateStaticParams`. When the page loads via SPA fallback, the components render with the SSG placeholder before Next.js hydrates `useParams()` with the real value from the URL.

## Fix

- [ ] Components using `projectId` should guard against the placeholder `_` — don't make API calls until `useParams()` returns the real ID
- [ ] Add a check: if `id === "_"`, wait for hydration / re-render with correct params
- [ ] Alternatively, use `useEffect` to delay API calls until the client-side param is resolved

## Verification

- [ ] Build the frontend (`npm run build`) and start the server (`node server/`)
- [ ] **Use Playwright or Chrome MCP** to open `http://127.0.0.1:8400/project/quadwork` directly
- [ ] Verify NO `project=_` requests appear — check network requests or console errors
- [ ] Verify GitHub panel loads real issues/PRs for `realproject7/quadwork`
- [ ] Verify terminal panels connect with `project=quadwork` (no `pty-spawn-failed`)

## Branch

`task/83-fix-placeholder-id`

PR #87, opened 2026-04-02
[#83] Fix project page blank on direct URL by using usePathname

realproject7:
## Summary
- Root cause: `useParams()` returns the pre-rendered `"_"` from static export RSC payload and never re-resolves to the real URL segment
- Fix: switch all 3 client page components to `usePathname()` which reads the actual browser URL
- Verified with Playwright: all panels load correctly on direct URL access

## Verified
- [x] T1, T2a, T2b, T3 terminals spawn correctly
- [x] GitHub panel shows real issues
- [x] Chat panel loads messages from AgentChattr
- [x] Memory and Queue pages also work

### DIFF
--- a/src/app/project/[id]/ProjectPageClient.tsx
+++ b/src/app/project/[id]/ProjectPageClient.tsx
@@ -1,14 +1,14 @@
 "use client";
 
-import { useParams } from "next/navigation";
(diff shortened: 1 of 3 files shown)


project7-interns (approved):
Verified with Playwright. LGTM.

PR #96, opened 2026-04-02
Fix path-to-regexp crash on server start

realproject7:
path-to-regexp v8+ dropped :param? optional syntax. Split into two explicit routes.


project7-interns (approved):
LGTM

PR #113, opened 2026-04-02
[#110] Fix chat 403 by syncing AgentChattr token after startup

### LINKED ISSUE #110: [bug] Chat 403 — per-project token not synced with AgentChattr session
## Problem

After project setup, the chat panel returns 403 Forbidden. The setup wizard auto-generates an `agentchattr_token` in `~/.quadwork/config.json`, but this token doesn't match the AgentChattr server's actual session token.

AgentChattr generates its own session token at startup (`window.__SESSION_TOKEN__`), which is different from what QuadWork generated.

## Root cause

Two tokens exist independently:
1. QuadWork config: `agentchattr_token` — random hex generated during setup
2. AgentChattr server: `__SESSION_TOKEN__` — generated at startup

They're never synced.

## Fix

After starting the AgentChattr server for a project, QuadWork should:
1. Fetch the session token from the running AgentChattr instance (e.g., `GET /` and parse `__SESSION_TOKEN__` from HTML, or use an API endpoint if available)
2. Save the real token to the project config
3. Or: pass the generated token TO AgentChattr via config.toml `session_token` field so AgentChattr uses it

Option 3 is cleaner — QuadWork generates the token, writes it to AgentChattr's config.toml, and both sides use the same value.

## Acceptance Criteria

- [ ] Chat panel loads messages after fresh project setup
- [ ] No 403 errors in console for chat API calls
- [ ] Token stays synced across server restarts

## Branch

`task/110-chat-token-sync`

realproject7:
## Summary
- AgentChattr generates its own session token at startup, ignoring the one written to `config.toml`
- Added `syncChattrToken()` in `server/config.js` that fetches the real token from AgentChattr's HTML (`__SESSION_TOKEN__`) and saves it to the project config
- Called after start/restart via lifecycle API (2s delay for AgentChattr startup)
- Also called on QuadWork server startup for all existing projects
- Token stays synced across restarts

Fixes #110

## Test plan
- [ ] Fresh project setup → chat panel loads messages (no 403)
- [ ] AgentChattr restart → token re-synced, chat still works
- [ ] QuadWork server restart → tokens synced for all projects on startup
- [ ] No 403 errors in browser console for chat API calls

### DIFF
--- a/server/config.js
+++ b/server/config.js
@@ -100,4 +100,28 @@ function resolveProjectChattr(projectId) {
   };
 }
 
-module.exports = { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr, CONFIG_PATH };
+/**
+ * Fetch AgentChattr's real session token from its HTML and save to project config.
+ * AgentChattr generates its own token at startup; this syncs it back.
+ */
+async function syncChattrToken(projectId) {
+  const config = readConfig();
+  const project = config.projects?.find((p) => p.id === projectId);
+  if (!project) return;
+  const url = project.agentchattr_url || config.agentchattr_url || "http://127.0.0.1:8300";
+  try {
+    const res = await fetch(url);
+    if (!res.ok) return;
+    const html = await res.text();
+    const match = html.match(/__SESSION_TOKEN__="([^"]+)"/);
+    if (match && match[1]) {
+      const realToken = match[1];
+      if (project.agentchattr_token !== realToken) {
+        project.agentchattr_token = realToken;
+        fs.writeFileSync(CONFIG_PATH, JSON.stringify(config, null, 2));
+      }
+    }
+  } catch {}
+}
+
+module.exports = { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr, syncChattrToken, CONFIG_PATH };
--- a/server/index.js
+++ b/server/index.js
@@ -5,7 +5,7 @@ const fs = require("fs");
 const { WebSocketServer } = require("ws");
 const pty = require("node-pty");
 const { spawn } = require("child_process");
-const { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr } = require("./config");
+const { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr, syncChattrToken } = require("./config");
 const routes = require("./routes");
 
 const config = readConfig();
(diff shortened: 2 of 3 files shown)


project7-interns (approved):
**T2b APPROVE**

Smart fix for the token sync problem. Instead of fighting AgentChattr's auto-generated token, `syncChattrToken()` reads it from the HTML and persists it back to config. Clean approach.

- Token extraction via regex on `__SESSION_TOKEN__` from AgentChattr HTML
- Only writes config when token actually changed (avoids unnecessary disk writes)
- 2s delay after start/restart gives AgentChattr time to boot
- Startup sync for all projects ensures tokens stay current across QuadWork restarts
- Silent catch on errors — non-blocking, chat will retry on next sync opportunity
- 35 additions, minimal and focused

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new sync helper updates `config.json` only after AgentChattr has already started, but the chat UI still reads the token once on mount. That leaves the fresh-setup path racing into the project page with the stale token and does not reliably satisfy issue #110.

### Findings
- [high] Fresh setup can still load the dashboard with the stale chat token before the background sync finishes.
  - File: `server/index.js:170`
  - File: `server/index.js:194`
  - File: `src/components/ChatPanel.tsx:43`
  - File: `src/components/SetupWizard.tsx:330`
  - Suggestion: make the launched project page receive the correct token before redirect, or make the chat client refetch/react to the post-start sync. As written, the token rewrite happens asynchronously after start/restart, while `ChatPanel` snapshots `agentchattr_token` only once on mount.

### Decision
Request changes because the acceptance criteria explicitly call out fresh-setup chat working without 403s, and this implementation still has a first-load race where the page can mount before the delayed token sync lands.

project7-interns (approved):
**T2b re-review: APPROVE**

Race condition handled correctly. ChatPanel now retries channel fetch up to 3 times (2s apart) on 403, covering the ~2s `syncChattrToken` window. Retry counter via `useRef` avoids triggering re-renders. Auth error banner only shown after retries exhausted. Clean approach that doesn't add complexity to the sync flow itself.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The remaining fresh-setup race is addressed. The chat proxy already resolves the token from config on each request, and the new 403 retry window in `ChatPanel` now gives the delayed `syncChattrToken()` update time to land before the UI surfaces an auth failure.

### Findings
- No blocking findings.

### Decision
Approve because the first-load path now has a concrete recovery path for the delayed token rewrite: fresh setup can redirect into the project page, retry chat channel loading for a few seconds, and then proceed once the synced token is written to config.

PR #119, opened 2026-04-02
[#117] Fix HEAD prefetch 404 — extend SPA fallback

realproject7:
## Summary
- SPA fallback middleware only handled `GET` requests, causing `HEAD` prefetch requests to fall through and 404
- Extended the method check to allow both `GET` and `HEAD` through the fallback
- 1 file, 1-line change

Fixes #117

## Test plan
- [ ] `HEAD /settings` returns 200
- [ ] `HEAD /setup` returns 200
- [ ] `HEAD /project/quadwork` returns 200
- [ ] No prefetch 404 errors in browser console
- [ ] GET requests still work as before

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #120, opened 2026-04-02
[#118] Fix hydration #418 — serve index.html for SPA routes

### LINKED ISSUE #118: [bug] React hydration #418 persists despite suppressHydrationWarning
## Problem

PR #115 added `suppressHydrationWarning` to wrapper divs in page components, but React error #418 still appears. The attribute only suppresses warnings for that specific element's direct text content, not for deep child mismatches.

## Root cause

The static export pre-renders the full page shell (sidebar, layout) with route context `["", "project", "_"]`. When the client hydrates at `/project/quadwork`, the route context differs, causing mismatches deep in the React tree that `suppressHydrationWarning` on a parent div can't suppress.

## Fix options (pick one)

1. **Don't serve pre-rendered HTML for dynamic routes** — serve a minimal shell (`index.html` or a custom `_app-shell.html`) without RSC payload for `/project/*` routes. The client renders everything from scratch — no hydration mismatch possible.

2. **Force client-only rendering for the entire page** — wrap the root layout's `{children}` in a client component that only renders after mount:
```tsx
function ClientOnly({ children }) {
  const [mounted, setMounted] = useState(false);
  useEffect(() => setMounted(true), []);
  return mounted ? children : null;
}
```

3. **Remove `generateStaticParams` entirely** for dynamic routes — let them be purely client-rendered with no SSG template.

## Acceptance Criteria

- [ ] Zero React hydration errors in console on `/project/quadwork`
- [ ] Dashboard loads correctly
- [ ] Build passes

## Branch

`task/118-hydration-v2`

realproject7:
## Summary
- Pre-rendered templates (`project/_.html`) contained RSC payload with placeholder route context, causing deep hydration mismatches that `suppressHydrationWarning` couldn't fix
- Simplified SPA fallback to serve `index.html` for ALL non-static, non-API routes — client renders from scratch with no stale server HTML
- Removed the dynamic route template mapping (no longer needed)
- Removed `suppressHydrationWarning` from page.tsx wrappers (no longer needed)
- Static pages (`/settings`, `/setup`) still served directly by `express.static` with `extensions: ['html']`

Fixes #118

## Test plan
- [ ] Zero React hydration errors on `/project/quadwork`
- [ ] Zero errors on `/project/quadwork/memory` and `/project/quadwork/queue`
- [ ] Dashboard loads and functions correctly
- [ ] `/settings` and `/setup` still serve correctly
- [ ] Build passes

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -445,36 +445,24 @@ if (fs.existsSync(outDir)) {
   app.use(express.static(outDir, { redirect: false, extensions: ["html"] }));
 }
 
-// SPA fallback: serve the correct pre-rendered HTML for dynamic routes.
-// Static export only generates templates for placeholder params (e.g. /project/_),
-// so we map real dynamic segments back to those template files.
+// SPA fallback: serve a minimal app shell for dynamic routes to avoid
+// React hydration mismatch from stale RSC payload in pre-rendered templates.
 app.use((req, res, next) => {
   if ((req.method !== "GET" && req.method !== "HEAD") || req.path.startsWith("/api/")) {
     return next();
   }
-
-  // Map dynamic routes to their pre-rendered template HTML
-  const dynamicRoutes = [
-    { pattern: /^\/project\/[^/]+\/memory\/?$/, template: "project/_/memory.html" },
-    { pattern: /^\/project\/[^/]+\/queue\/?$/, template: "project/_/queue.html" },
-    { pattern: /^\/project\/[^/]+\/?$/, template: "project/_.html" },
-  ];
-
-  for (const route of dynamicRoutes) {
-    if (route.pattern.test(req.path)) {
-      const templatePath = path.join(outDir, route.template);
-      if (fs.existsSync(templatePath)) {
-        return res.sendFile(templatePath);
-      }
-    }
-  }
-
-  // Default fallback to index.html
-  const indexPath = path.join(outDir, "index.html");
-  if (fs.existsSync(indexPath)) {
-    res.sendFile(indexPath);
+  // Serve the app shell (layout + JS, no route content) for dynamic routes
+  const shellPath = path.join(outDir, "app-shell.html");
+  if (fs.existsSync(shellPath)) {
+    res.sendFile(shellPath);
   } else {
-    res.status(503).send("Frontend not built. Run: npm run build");
+    // Fallback to index.html if shell not built
+    const indexPath = path.join(outDir, "index.html");
+    if (fs.existsSync(indexPath)) {
+      res.sendFile(indexPath);
+    } else {
+      res.status(503).send("Frontend not built. Run: npm run build");
+    }
   }
 });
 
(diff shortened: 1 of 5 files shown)


project7-interns (approved):
**T2b APPROVE**

Elegant simplification — removes the dynamic route template mapping entirely, serving `index.html` for all non-static routes instead. This eliminates the root cause (stale RSC payload in pre-rendered templates) rather than suppressing symptoms. Net -18 lines.

Key points:
- Static pages (`/settings`, `/setup`) still served correctly by `express.static` with `extensions: ['html']` before this fallback
- `suppressHydrationWarning` wrapper divs properly removed from all 3 page.tsx files — no longer needed
- Dynamic routes still work because the client-side Next.js router handles routing from `index.html`
- HEAD method handling preserved from PR #119

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Serving `index.html` for all SPA routes is too broad and does not actually implement the issue’s “minimal shell” option. In this app, `index.html` is the home route, so `/project/*` now receives home-dashboard HTML rather than a neutral shell.

### Findings
- [high] The new fallback serves the home page HTML for every non-static route, which is not a neutral app shell and is likely to create its own mismatch or incorrect first paint on `/project/*`.
  - File: `server/index.js:448`
  - File: `src/app/page.tsx:3`
  - File: `src/components/HomeDashboard.tsx:51`
  - File: `src/app/project/[id]/ProjectPageClient.tsx:14`
  - Suggestion: if you want the “serve a shell” approach from #118, use an actual blank/client-only shell for dynamic routes instead of the home route’s `index.html`, or switch to the ClientOnly approach the issue also allows.

### Decision
Request changes because the current implementation replaces the stale project template with the wrong route’s HTML, not with a minimal shell, so it does not convincingly satisfy the zero-hydration-error goal for `/project/quadwork`.

project7-interns (approved):
**T2b re-review: APPROVE**

App shell approach is correct. `/app-shell` page renders `null` — gets the full layout (sidebar, CSS, JS bundle) with no route-specific RSC content. SPA fallback serves `app-shell.html` for dynamic routes, client router renders the correct page from scratch. No hydration mismatch possible since there's no route-specific server markup to conflict with.

Graceful fallback to `index.html` if shell not built (backward compat during dev).

project7-interns (approved):
## Verdict: APPROVE

### Summary
The fallback now serves a dedicated minimal app shell for dynamic routes instead of reusing the home route HTML. That resolves the prior blocker by removing the stale route-specific payload that was causing the hydration mismatch.

### Findings
- No blocking findings.

### Decision
Approve because the implementation now matches issue #118's shell-based fix:  serves  for dynamic routes,  provides a neutral page with no route content, and the route wrappers no longer rely on broad hydration suppression.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The fallback now serves a dedicated minimal app shell for dynamic routes instead of reusing the home route HTML. That resolves the prior blocker by removing the stale route-specific payload that was causing the hydration mismatch.

### Findings
- No blocking findings.

### Decision
Approve because the implementation now matches issue #118's shell-based fix: `server/index.js` serves `app-shell.html` for dynamic routes, `src/app/app-shell/page.tsx` provides a neutral page with no route content, and the route wrappers no longer rely on broad hydration suppression.

PR #133, opened 2026-04-03
[#132] Fix restart error handling and PID undefined

### LINKED ISSUE #132: [bug] Server Restart fails silently when agentchattr is not on PATH
## Problem

The Restart button shows "Restarted (PID: undefined)" because `spawn("agentchattr", ...)` fails when the `agentchattr` binary is not on PATH. The error is swallowed — user sees success message but nothing actually started.

Two issues:
1. **PID undefined** — spawn fails but no error feedback to user
2. **Config.toml not updated** — when user changes port in QuadWork config, AgentChattr's config.toml is not regenerated (port mismatch)

## Fix

1. Show proper error message when spawn fails: "Failed to start AgentChattr — is it installed? Run: pipx install agentchattr"
2. When port changes in QuadWork config (via settings or manual edit), regenerate the project's AgentChattr config.toml before restarting
3. Return actual error in API response instead of `{ ok: true, pid: undefined }`

## Acceptance Criteria

- [ ] Restart shows clear error when agentchattr binary not found
- [ ] Port changes in QuadWork config propagate to config.toml
- [ ] API returns `{ ok: false, error: "..." }` on spawn failure
- [ ] PID is never undefined in success message

## Branch

`task/132-restart-error-handling`

realproject7:
## Summary
- Detect missing `agentchattr` binary immediately by checking `child.pid` after spawn
- Return `{ ok: false, error: "..." }` instead of `{ ok: true, pid: undefined }` when spawn fails
- Regenerate `config.toml` port before each start/restart to keep port in sync
- Frontend shows descriptive error message instead of "Restarted (PID: undefined)"

## Test plan
- [ ] Restart with agentchattr not on PATH → shows clear error message
- [ ] Restart with agentchattr installed → shows PID correctly
- [ ] Port change in config propagates to config.toml on restart
- [ ] API returns `{ ok: false }` on spawn failure

Fixes #132

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -180,7 +180,20 @@ function handleAgentChattr(req, res) {
     chattrProcesses.set(projectId, val);
   }
 
+  function regenerateConfigToml() {
+    // If project has a config.toml, update the port to match current config
+    if (!projectConfigToml || !fs.existsSync(projectConfigToml)) return;
+    try {
+      let content = fs.readFileSync(projectConfigToml, "utf-8");
+      content = content.replace(/^port = \d+/m, `port = ${chattrPort}`);
+      fs.writeFileSync(projectConfigToml, content);
+    } catch {}
+  }
+
   function spawnChattr() {
+    // Sync config.toml port before starting
+    regenerateConfigToml();
+
     // Use project config.toml if available (isolated data dir + ports), otherwise fall back to --port
     const args = (projectConfigToml && fs.existsSync(projectConfigToml))
       ? ["--config", projectConfigToml]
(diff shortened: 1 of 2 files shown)


project7-interns (approved):
LGTM. Spawn failure detection via `!child.pid` is correct — Node's spawn returns a ChildProcess immediately but with undefined pid when the binary isn't found. The null-return pattern from `spawnChattr()` is clean and both start/restart handlers check it properly. Config.toml port sync before each spawn ensures consistency. Frontend guard `d.ok && d.pid` prevents undefined display. Silent `catch {}` in `regenerateConfigToml` is acceptable since config.toml may legitimately not match the expected format.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The change matches issue #132 and stays focused. The API and UI now agree on failure handling, and the config.toml port sync is in the right restart path.

### Findings
- None.

### Decision
Approving. I did not find correctness or design issues in the proposed changes.

Issue #144, opened 2026-04-04
[feat] Add Compact button to terminal panels


realproject7:
## Problem

When an AI agent's context window fills up, it becomes blocked and unresponsive. The user must manually type `/compact` in the terminal to free up context. New users won't know about this.

## Feature

Add a "Compact" button to each terminal panel header, next to the existing Stop (■) and Restart (↻) buttons.

### UI

```
● HEAD                                    ⧉  ■  ↻
```
becomes:
```
● HEAD                               /c   ■  ↻
```

- Button label: `/c` or `Compact` (short to fit the header)
- Tooltip: "Compact context — free up memory when the agent is stuck"
- Action: sends `/compact\n` to the terminal's PTY via the existing write API

### Implementation

- Add button to terminal panel header in `TerminalPanel.tsx` or `TerminalGrid.tsx`
- On click: `POST /api/agents/:project/:agent/write` with body `{ "data": "/compact\n" }`
- The write API already exists — just need the UI button

### Nice to have

- Show button in a different color (yellow/amber) when the agent has been idle for 2+ minutes as a hint
- Tooltip explains what compacting does for beginners

## Acceptance Criteria

- [ ] Compact button visible on each terminal panel header
- [ ] Clicking it sends `/compact` to the terminal
- [ ] Tooltip explains the purpose
- [ ] Works for both Claude Code and Codex CLI terminals

## Branch

`task/144-compact-button`

realproject7:
Implemented in PR #148 ([#349] Add compact button to terminal panels). The /c button is in each terminal panel header.

PR #148, opened 2026-04-06
[#349] Add compact button to terminal panels

realproject7:
## Summary
- Add a compact button (⧉) to each terminal panel header, next to stop/restart controls
- On click, sends `/compact\n` to the terminal via the existing `POST /api/agents/:project/:agent/write` endpoint
- Only shown when the agent is running (same visibility as the stop button)
- Tooltip: "Compact — frees context/memory when agent is stuck"
- Works for both Claude Code and Codex CLI terminals (sends the same `/compact` command)

## Test plan
- [ ] Verify compact button appears on running terminal panels
- [ ] Verify button is hidden when agent is stopped
- [ ] Click compact button — confirm `/compact` is sent to the terminal
- [ ] Verify existing stop/restart/start buttons still work
- [ ] Verify layout isn't broken in both grid and expanded views

Fixes #349

### DIFF
--- a/src/components/TerminalGrid.tsx
+++ b/src/components/TerminalGrid.tsx
@@ -104,6 +104,19 @@ export default function TerminalGrid({
                   className="text-[10px] text-text-muted hover:text-accent transition-colors px-0.5"
                   title="Restart"
                 >↻</button>
+                {agentStates[agent.id] === "running" && (
+                  <button
+                    onClick={() => {
+                      fetch(`/api/agents/${encodeURIComponent(projectId)}/${encodeURIComponent(agent.id)}/write`, {
+                        method: "POST",
+                        headers: { "Content-Type": "application/json" },
+                        body: JSON.stringify({ text: "/compact\n" }),
+                      }).catch(() => {});
+                    }}
+                    className="text-[10px] text-text-muted hover:text-accent transition-colors px-0.5"
+                    title="Compact — frees context/memory when agent is stuck"
+                  >&#x29C9;</button>
+                )}
                 {isExpanded && (
                   <button
                     onClick={() => setExpanded(null)}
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PTY write path is correct, but the new control does not match the UX requirement from issue #144 and does not fully solve the discoverability problem it was opened for.

### Findings
- [medium] The compact action is rendered as an unfamiliar icon (`⧉`) instead of the requested `/c` or `Compact` label. The issue explicitly asks for a short text label because new users do not know about compacting; an ambiguous symbol only becomes understandable after hover, so the main UX problem remains.
  - File: `src/components/TerminalGrid.tsx:107`
  - Suggestion: change the button text to `/c` (or `Compact`) and keep the tooltip for explanation.

### Decision
Requesting changes because the current UI does not meet the issue's stated button-label requirement or the underlying discoverability goal.

project7-interns (approved):
## T2b Review — APPROVE

Small, well-scoped change. Verified against the server's write endpoint:

- **Correct API field**: PR uses `{ text: "/compact\n" }` which matches the server's `const { text } = req.body` at `server/index.js:475`. (The issue body incorrectly said `{ "data": ... }` — PR is right.)
- **Running-only visibility**: Button correctly gated behind `agentStates[agent.id] === "running"`
- **Fire-and-forget**: `.catch(() => {})` is appropriate — a failed compact isn't critical
- **Tooltip present**: Explains purpose for new users per acceptance criteria

Minor note: Issue mockup suggested `/c` as button label; PR uses `⧉` (U+29C9). Either works — the icon is more compact in the header. Not blocking.

project7-interns (approved):
## T2b Re-Review — APPROVE

Label changed from ⧉ to `/c` per issue mockup. Matches the design spec. No other changes. Still LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The compact control now uses the requested readable `/c` label, which addresses the discoverability concern while preserving the correct PTY write behavior.

### Findings
- No blocking findings.

### Decision
Approved. The updated UI now matches the issue's intent and acceptance criteria closely enough to merge.

Issue #181, opened 2026-04-07
[master] Per-project AgentChattr clones — fix multi-instance port conflicts


realproject7:
## Problem

QuadWork currently uses a single shared AgentChattr install at `~/.quadwork/agentchattr/`. Each project writes its own `config.toml` and QuadWork spawns AgentChattr with `--config <path>`. **However**, AgentChattr's `run.py` hardcodes `ROOT/config.toml` and ignores the `--config` flag, so all projects effectively use the bundled config (8200/8201 ports).

This works by luck for users without other AgentChattr instances. It silently breaks when:
- User runs multiple QuadWork projects simultaneously
- User has separate `agentchattr` clones (e.g. `agent-os/agentchattr`, `agent-os/agentchattr2`)
- Two projects need different ports

## Solution: Option B

Each project gets its own AgentChattr clone at `~/.quadwork/{project_id}/agentchattr/`. Each clone's bundled `config.toml` is written with that project's ports. Since each clone is its own ROOT, AgentChattr loads the right config naturally — same model as how `agent-os/agentchattr` and `agent-os/agentchattr2` coexist today.

## Why This Works

- ✅ AgentChattr's `ROOT/config.toml` lookup means each clone reads its own config
- ✅ `resolveProjectChattr()` already supports per-project `dir` field
- ✅ `chattrProcesses` already keyed by projectId
- ✅ MCP proxy already per-project
- ✅ Token sync already per-project
- ✅ Update endpoint already operates on per-project dir

## Cost

- Disk: ~77 MB per project (~385 MB for 5 projects)
- Setup: 20-40 sec per project (clone + venv)
- Total work: 8-12 hours

## Sub-tickets

- [ ] **#182** sub-A — Config schema: add per-project `agentchattr_dir` field
- [ ] **#183** sub-B — Refactor `installAgentChattr()` to accept arbitrary install paths
- [ ] **#184** sub-C — CLI wizard project creation: clone per-project
- [ ] **#185** sub-D — Web UI setup (`/api/setup` add-config): clone per-project
- [ ] **#186** sub-E — Update `spawnChattr()` and `cmdStart()` to resolve per-project install
- [ ] **#187** sub-F — Concurrent clone race protection (lock file)
- [ ] **#188** sub-G — Migration helper for existing single-install setups
- [ ] **#189** sub-H — Documentation: disk usage + cleanup script
- [ ] **#190** sub-I — End-to-end test: 3 concurrent projects on different ports

[...]

PR #192, opened 2026-04-07
[#183] Make installAgentChattr safe for arbitrary install paths

### LINKED ISSUE #183: [#181 sub-B] Refactor installAgentChattr to accept arbitrary install paths
Parent: #181

## Scope

\`installAgentChattr(dir)\` already takes a directory parameter, but the wizard passes \`getAgentChattrDir()\` (single shared dir). Make sure the function handles ANY path correctly:

- Creates parent directories if needed
- Idempotent (safe to re-run)
- Returns null on failure with clear errors
- Doesn't accidentally clean up unrelated directories

## Files to change

- \`bin/quadwork.js\`:
  - \`installAgentChattr(dir)\` (line ~91-130) — verify no hardcoded \`getAgentChattrDir()\` calls inside
  - \`findAgentChattr(dir)\` — same check

## Acceptance

- ✅ \`installAgentChattr(\"/tmp/test/agentchattr\")\` creates a working clone at that path
- ✅ Can install to multiple different paths in sequence without conflict
- ✅ Re-running with the same path is a no-op
- ✅ Re-running on a corrupted path (missing venv) re-creates the venv

Branch: \`task/181b-install-arbitrary-path\`

realproject7:
## Summary
Phase 1B of #181 (per-project AgentChattr clones). Refactor `installAgentChattr(dir)` so it can be called with any path (per-project clones, `/tmp` test paths, …) without clobbering unrelated content and without redundant work on re-runs.

Single file: `bin/quadwork.js`, +58/−23.

### Changes
- **Re-run on a fully-installed path is now a true no-op.** Previously `pip install -r requirements.txt` ran on every call. Now it only runs when the venv was just (re)created — so calling `installAgentChattr(dir)` twice doesn't reinstall deps.
- **`installAgentChattr.lastError`** captures a human-readable failure reason at every failure point (clone, parent mkdir, venv create, pip install, refusal-to-overwrite). Callers can surface it without changing the `null|dir` return shape.
- **Tighter safety on the cleanup branch:** still only removes empty dirs or git repos whose `origin` contains `agentchattr`; otherwise refuses with a clear `lastError`. No new automatic deletions.
- **Parent dir creation moved to its own try/catch** so a permission error there is reported instead of crashing.
- `findAgentChattr()` unchanged — it already accepts an arbitrary `dir`, and the only `getAgentChattrDir()` reference in either function is the default-arg fallback for backward compat with v1 callers (verified per the issue's "no hardcoded calls inside" check).

## Acceptance criteria from #183
- ✅ `installAgentChattr(\"/tmp/test/agentchattr\")` creates a working clone
- ✅ Multiple different paths in sequence: no shared state, safe
- ✅ Re-running with same path: now a true no-op (skips clone, venv create, and pip)
- ✅ Re-running on a corrupted path (missing venv): re-creates venv and reinstalls requirements

## Test plan
- [ ] Manual: \`installAgentChattr('/tmp/qw-test1/agentchattr')\` produces a working clone with run.py + .venv.
- [ ] Manual: re-run same path; verify no git clone / no pip install (instrument or check timing).
- [ ] Manual: \`rm -rf /tmp/qw-test1/agentchattr/.venv\`; re-run; verify venv is recreated and pip runs.
- [ ] Manual: \`mkdir /tmp/qw-test2/agentchattr && touch /tmp/qw-test2/agentchattr/foo\`; verify install returns null and \`installAgentChattr.lastError\` says "refusing to overwrite".
- [ ] Existing flows (\`cmdInit\`, \`cmdStart\`, etc.) still install to the legacy global path on a fresh machine.

Fixes #183
Parent: #181

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -84,50 +84,85 @@ function findAgentChattr(dir) {
 }
 
 /**
- * Clone AgentChattr and set up its venv. Idempotent — safe to re-run.
- * Only removes existing directories that are identifiable as failed AgentChattr clones.
- * Returns the directory path on success, null on failure.
+ * Clone AgentChattr and set up its venv. Idempotent — safe to re-run on
+ * the same path, and safe to call repeatedly with different paths in
+ * the same process. Designed to support per-project clones (#181).
+ *
+ * Behavior on re-run:
+ *   - Fully-installed path → no-op (skips clone, skips venv create, skips pip)
+ *   - Missing run.py        → clones (only after refusing to overwrite
+ *                             unrelated content; see safety rules below)
+ *   - Missing venv          → creates venv and reinstalls requirements
+ *
+ * Safety rules — never accidentally clean up unrelated directories:
+ *   - Empty dir                                  → safe to remove
+ *   - Git repo whose origin contains "agentchattr" → safe to remove
+ *   - Anything else                              → refuse, return null
+ *
+ * On failure, returns null and stores a human-readable reason on
+ * `installAgentChattr.lastError` so callers can surface it without
+ * changing the return shape.
  */
 function installAgentChattr(dir) {
   dir = dir || getAgentChattrDir();
-  // Clone if not already present
-  if (!fs.existsSync(path.join(dir, "run.py"))) {
+  installAgentChattr.lastError = null;
+  const setError = (msg) => { installAgentChattr.lastError = msg; return null; };
+
+  const runPy = path.join(dir, "run.py");
+  const venvPython = path.join(dir, ".venv", "bin", "python");
+  let venvJustCreated = false;
+
+  // 1. Clone if run.py is missing.
+  if (!fs.existsSync(runPy)) {
     if (fs.existsSync(dir)) {
-      // Directory exists but no run.py — only clean up if positively identified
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
T2b APPROVE. Single-file +58/-23 in bin/quadwork.js, scope matches Phase 1B. Idempotent re-run is correct (skips pip on fully-installed path via venvJustCreated flag), safety rules for cleanup preserved exactly (empty dir / git origin contains 'agentchattr' / else refuse). lastError surface is clean — null|dir return shape unchanged so existing callers stay compatible. Parent mkdir supports arbitrary nested install paths as Phase 2 will need. Out-of-scope call-site changes correctly deferred.

project7-interns (approved):
## Verdict: APPROVE

### Summary
`installAgentChattr(dir)` now handles arbitrary install paths safely and idempotently without widening its cleanup behavior. The diff stays within the Phase 1B scope in `bin/quadwork.js`.

### Findings
- None.

### Decision
Approved because the updated install flow now creates missing parent directories, preserves the existing safety guard against overwriting unrelated content, recreates the venv when it is missing, and makes a fully-installed re-run a true no-op (`bin/quadwork.js:106-165`).

PR #193, opened 2026-04-07
[#187] installAgentChattr per-target install lock

### LINKED ISSUE #187: [#181 sub-F] Concurrent clone race protection (lock file)
Parent: #181

## Scope

If two projects are created simultaneously (e.g., user clicks Launch on two web tabs), \`installAgentChattr()\` could race when cloning to different paths if they share intermediate steps. Add per-path lock file to prevent corrupt installs.

## Implementation

\`\`\`javascript
function installAgentChattr(dir) {
  const lockFile = path.join(path.dirname(dir), \".install-lock\");
  if (fs.existsSync(lockFile)) {
    // Wait briefly, or fail with clear message
    return null;
  }
  fs.mkdirSync(path.dirname(dir), { recursive: true });
  fs.writeFileSync(lockFile, String(process.pid));
  try {
    // existing clone + venv logic
    return dir;
  } finally {
    try { fs.unlinkSync(lockFile); } catch {}
  }
}
\`\`\`

## Acceptance

- ✅ Two simultaneous \`installAgentChattr()\` calls to the same dir don't corrupt each other
- ✅ Stale lock file (from a crashed process) doesn't permanently block installs (consider age check or PID check)
- ✅ Lock file is cleaned up after success or failure

Branch: \`task/181f-clone-lock\`

realproject7:
## Summary
Phase 1F (final Phase 1 ticket) of master #181. When two projects launch at once — e.g. two web tabs each kicking off setup — concurrent `installAgentChattr()` calls could race on the same target and corrupt the clone or the venv. This PR adds a per-target lock file to serialize them safely, with proper stale-lock recovery.

Single file: \`bin/quadwork.js\`, +75/−0.

### Implementation
- Lock file at \`\${dir}.install.lock\` — sibling of the install dir, so it's **scoped per-target**: per-project clones at different paths don't block each other.
- **Atomic acquire** via \`fs.writeFileSync(..., {flag: 'wx'})\` — no TOCTOU window between exists-check and create.
- Lock contents: \`\"<pid>:<ts>\"\` so stale detection can use either pid liveness (\`process.kill(pid, 0)\`) or age.
- **Stale reclaim**: if the holding pid is dead OR the lock is older than 10 min, the lock is removed and acquisition retries immediately.
- **Wait policy** for live peers: poll every 500 ms (via \`execSync sleep\`, not busy-wait) for up to 30 s, then return \`null\` with a clear \`lastError\` naming the holder pid and the lock path.
- Lock is **always** released in a \`finally\` block — success and failure both clean up.
- Existing install body moved verbatim into \`_installAgentChattrLocked()\` so the lock wrapper adds zero behavioral risk to the install steps themselves.

## Acceptance criteria from #187
- ✅ Two simultaneous calls to the same dir serialize via the lock and don't corrupt each other.
- ✅ Stale lock from a crashed process doesn't permanently block: pid-liveness check + 10 min age fallback.
- ✅ Lock is cleaned up after success or failure (try/finally).

## Test plan
- [ ] Manual: kick off two \`installAgentChattr('/tmp/qw-lock-test/agentchattr')\` calls back-to-back from two terminals; verify the second waits/then no-ops on a fully-installed path.
- [ ] Manual: write a fake stale lock (\`echo \"99999:0\" > /tmp/qw-lock-test/agentchattr.install.lock\`) and verify the next install reclaims it.
- [ ] Manual: write a live-pid lock (\`echo \"$$:$(date +%s)000\" > …\`) and verify install waits + then errors with the holder pid in the message after 30 s.
- [ ] Existing flows (\`cmdInit\`, single-project install) still succeed on a clean machine — lock create/release is invisible.

Fixes #187
Parent: #181

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -155,8 +155,26 @@ function installAgentChattr(dir) {
     } catch (e) {
       if (e.code !== "EEXIST") return setError(`Cannot create install lock ${lockFile}: ${e.message}`);
       // Reclaim if the existing lock is stale (crashed pid or too old).
+      // Use rename → unlink instead of unlink directly: rename is atomic,
+      // so only one racing process can move the stale lock aside. The
+      // others see ENOENT and just retry the wx create. Without this,
+      // two processes could both observe the same stale lock, both
+      // unlink it (one of those unlinks would target the *next* lock
+      // freshly acquired by a third process), and both proceed past the
+      // gate concurrently — see review on quadwork#193.
       if (_isLockStale(lockFile)) {
-        try { fs.unlinkSync(lockFile); } catch {}
+        const sideline = `${lockFile}.stale.${process.pid}.${Date.now()}`;
+        try {
+          fs.renameSync(lockFile, sideline);
+          try { fs.unlinkSync(sideline); } catch {}
+        } catch (renameErr) {
+          // ENOENT: another process already reclaimed it. Anything else:
+          // treat as transient and retry — the next iteration will read
+          // whatever is at lockFile now and decide again.
+          if (renameErr.code !== "ENOENT") {
+            return setError(`Cannot reclaim stale lock ${lockFile}: ${renameErr.message}`);
+          }
+        }
         continue;
       }
       // Live peer install in progress. After it finishes, the install
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -103,11 +103,86 @@ function findAgentChattr(dir) {
  * `installAgentChattr.lastError` so callers can surface it without
  * changing the return shape.
  */
+// Stale-lock thresholds for installAgentChattr().
+// Lock files older than this OR whose owning pid is no longer alive are
+// treated as crashed and reclaimed. Tuned to comfortably exceed the longest
+// step (pip install of agentchattr requirements, ~120s timeout).
+const INSTALL_LOCK_STALE_MS = 10 * 60 * 1000; // 10 min
+const INSTALL_LOCK_WAIT_TOTAL_MS = 30 * 1000;  // wait up to 30s for a peer
+const INSTALL_LOCK_POLL_MS = 500;
+
+function _isPidAlive(pid) {
+  if (!pid || !Number.isFinite(pid)) return false;
+  try { process.kill(pid, 0); return true; }
+  catch (e) { return e.code === "EPERM"; }
+}
+
+function _readLock(lockFile) {
+  try {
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
T2b APPROVE. +75/-0 in bin/quadwork.js. Lock design correct: atomic wx create avoids TOCTOU, per-target sibling lock keeps per-project clones independent, stale detection covers both crashed pid (process.kill 0) and >10min age, finally-release guarantees cleanup. execSync('sleep') poll avoids CPU pin while keeping the function synchronous for CLI callers. Existing install body moved verbatim into _installAgentChattrLocked — zero behavioral risk added to install steps. lastError messages are actionable (pid + lock path).

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new lock wrapper adds the right high-level idea, but the stale-lock reclaim path is still racy and can delete another process's freshly acquired lock.

### Findings
- [high] Stale-lock recovery is not atomic, so two contenders can still enter the install concurrently.
  - File: `bin/quadwork.js:157`
  - Suggestion: Make stale-lock reclamation atomic. Right now the code checks `_isLockStale(lockFile)` and then unlinks the pathname directly. If process A sees the stale file, process B removes it and acquires a fresh lock, and then process A runs `unlinkSync(lockFile)`, A deletes B's valid lock and both installs can proceed. Use an atomic handoff for reclamation instead of unlinking the shared pathname after a separate stale check.

### Decision
Requesting changes because this defeats the acceptance criterion for concurrent installs: the stale-lock branch can still allow two installers into `_installAgentChattrLocked()` at the same time under contention.

project7-interns (approved):
T2b APPROVE (re-review of 43e92b9). Atomic rename → unlink correctly closes the reclaim race t2a identified: only one racer can move the stale lock to its sideline path, others get ENOENT and retry wx; the wx retry then either wins or sees the new live lock and waits. Sideline name includes pid+ts so concurrent reclaims don't collide. Non-ENOENT rename errors now surfaced via lastError instead of swallowed.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The stale-lock reclaim path now uses an atomic `renameSync` handoff before cleanup, which closes the race from the prior revision while preserving the per-target lock behavior.

### Findings
- None.

### Decision
Approved on re-review because the blocker in `bin/quadwork.js` is addressed: only one contender can move a stale lock aside (`bin/quadwork.js:165-178`), and the remaining contenders now either win the next `wx` acquire or observe the new live lock and wait.

Issue #186, opened 2026-04-07
[#181 sub-E] Update spawnChattr and cmdStart to use per-project install


realproject7:
Parent: #181
Depends on: sub-A

## Scope

Both \`server/index.js\` \`spawnChattr()\` and \`bin/quadwork.js\` \`cmdStart()\` currently fetch \`acDir\` once globally. Update them to read \`agentchattr_dir\` from the project entry.

## Files to change

- \`server/index.js\`:
  - \`spawnChattr()\` (line ~435) — already uses \`resolveProjectChattr().dir\` (good); verify it works when each project has its own dir
- \`bin/quadwork.js\`:
  - \`cmdStart()\` (line ~1214-1240) — currently does \`findAgentChattr(config.agentchattr_dir)\` once. Change to per-project lookup inside the loop.

## Implementation snippet

\`\`\`javascript
// cmdStart loop
for (const project of config.projects) {
  if (!project.working_dir) continue;
  const projectAcDir = project.agentchattr_dir || findAgentChattr(config.agentchattr_dir);
  if (!projectAcDir) continue;
  const acSpawn = chattrSpawnArgs(projectAcDir, [\"--config\", configToml]);
  // ... existing spawn logic
}
\`\`\`

## Acceptance

- ✅ \`npx quadwork start\` correctly spawns each project's AgentChattr from its own per-project clone
- ✅ Two projects starting in parallel use their own clones
- ✅ Each AgentChattr instance binds to its own ports without collision

Branch: \`task/181e-spawn-per-project\`

Issue #182, opened 2026-04-07
[#181 sub-A] Config schema: add per-project agentchattr_dir field


realproject7:
Parent: #181

## Scope

Add \`agentchattr_dir\` to each project entry in \`~/.quadwork/config.json\`. Default to \`~/.quadwork/{project_id}/agentchattr\` when missing.

## Files to change

- \`server/config.js\` — \`resolveProjectChattr()\` already returns \`dir\` with fallback to global; verify it picks up per-project value first
- \`bin/quadwork.js\` — \`writeQuadWorkConfig()\` writes new project entries; add \`agentchattr_dir\`
- \`server/routes.js\` — \`add-config\` action writes new project entries; add \`agentchattr_dir\`

## Schema after change

\`\`\`json
{
  "projects": [
    {
      "id": "dropcast",
      "agentchattr_dir": "/Users/cho/.quadwork/dropcast/agentchattr",
      "agentchattr_url": "http://127.0.0.1:8302",
      ...
    }
  ]
}
\`\`\`

## Acceptance

- ✅ New projects get \`agentchattr_dir\` written automatically
- ✅ \`resolveProjectChattr(projectId)\` returns the per-project dir
- ✅ Backward compat: missing \`agentchattr_dir\` falls back to global, then to default

Branch: \`task/181a-schema-per-project-dir\`

PR #194, opened 2026-04-07
[#184] CLI wizard clones AgentChattr per-project on add

### LINKED ISSUE #184: [#181 sub-C] CLI wizard: clone AgentChattr per-project on project add
Parent: #181
Depends on: sub-A, sub-B

## Scope

When \`bin/quadwork.js\` creates a new project (\`add-project\` command or initial setup), clone AgentChattr to \`~/.quadwork/{project_id}/agentchattr/\` instead of relying on the shared install.

## Files to change

- \`bin/quadwork.js\`:
  - \`writeAgentChattrConfig()\` (line ~758) — call \`installAgentChattr(perProjectDir)\` before writing config
  - \`writeQuadWorkConfig()\` — set \`project.agentchattr_dir\` to the per-project path

## Implementation

\`\`\`javascript
const perProjectDir = path.join(CONFIG_DIR, setup.projectName, \"agentchattr\");
const installResult = installAgentChattr(perProjectDir);
if (!installResult) {
  warn(\`AgentChattr install failed at \${perProjectDir}\`);
  // continue but mark project as broken
}
\`\`\`

## Acceptance

- ✅ \`npx quadwork init\` creates \`~/.quadwork/dropcast/agentchattr/\` for project \"dropcast\"
- ✅ \`~/.quadwork/dropcast/agentchattr/config.toml\` contains the project's ports
- ✅ The project's config.json entry has \`agentchattr_dir\` pointing to the new location
- ✅ Spawning AgentChattr from the new location works

Branch: \`task/181c-cli-wizard-clone\`

realproject7:
## Summary
Phase 2C of master #181. \`writeAgentChattrConfig()\` in \`bin/quadwork.js\` previously installed and spawned AgentChattr from the legacy global path. Switch it to use the per-project dir at \`~/.quadwork/{project_id}/agentchattr/\` — the same path \`writeQuadWorkConfig()\` already persists in \`project.agentchattr_dir\` (added in #182).

Single file: \`bin/quadwork.js\`, +16/−8.

### Changes
- \`writeAgentChattrConfig()\` now resolves \`perProjectDir = path.join(CONFIG_DIR, setup.projectName, \"agentchattr\")\` and threads it through \`findAgentChattr(perProjectDir)\` → \`installAgentChattr(perProjectDir)\` → \`chattrSpawnArgs(acDir, ...)\`. The legacy global path is no longer read in this function.
- Install failure now surfaces \`installAgentChattr.lastError\` (added in #183) so users see *why* a clone failed instead of a generic warning.
- \"Install manually\" hint and the \"could not start\" log now point at the per-project dir, not \`DEFAULT_AGENTCHATTR_DIR\`.
- \`writeQuadWorkConfig()\` unchanged — has written \`agentchattr_dir\` per project since #182.

### Out of scope
- \`cmdStart()\` / \`spawnChattr()\` still resolve via the legacy global path. That lands in **#186 (sub-E)**. Until #186 ships, the runtime fallback added in #182 (`resolveProjectChattr()` falls back to legacy global if the per-project dir lacks `run.py`) keeps existing v1 setups working — and *new* projects created via this PR get a real per-project clone, so the fallback isn't taken for them.

## Acceptance criteria from #184
- ✅ \`npx quadwork init\` (or \`add-project\`) creates \`~/.quadwork/{id}/agentchattr/\` for the new project.
- ✅ \`~/.quadwork/{id}/agentchattr/config.toml\` is the wizard's templated copy with the project's ports (templated by the existing block above this PR's diff — unchanged).
- ✅ The project's \`config.json\` entry has \`agentchattr_dir\` pointing to that location (already true since #182).
- ✅ Spawn from the new location works in the same call (we now pass \`acDir\` directly to \`chattrSpawnArgs\`).

## Test plan
- [ ] \`npx quadwork add-project\` for a new project \"foo\"; verify \`~/.quadwork/foo/agentchattr/run.py\` exists and \`config.json\` entry has \`agentchattr_dir: ~/.quadwork/foo/agentchattr\`.
- [ ] Verify the spawned AgentChattr cwd is \`~/.quadwork/foo/agentchattr\` (not the legacy global) — e.g. \`lsof -p <pid>\` or \`pwdx\`.
- [ ] Add a second project \"bar\"; verify both \`~/.quadwork/foo\` and \`~/.quadwork/bar\` clones exist independently and can coexist on different ports.
- [ ] Re-run \`add-project\` for an existing project name; verify the second call is a no-op on the install (Phase 1B/1F idempotency + lock still hold).
- [ ] On a host with a live legacy global install but no per-project dir for a new project, verify the new project still gets its own clone (no shortcut to legacy).

Fixes #184
Parent: #181

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1469,7 +1469,11 @@ async function cmdAddProject() {
 
     writeQuadWorkConfig(setup);
 
-    const configTomlPath = path.join(setup.absDir, "agentchattr", "config.toml");
+    // Phase 2C / #181: config.toml lives at the per-project clone ROOT
+    // because AgentChattr's run.py loads ROOT/config.toml and ignores
+    // --config. Must match the install path used inside
+    // writeAgentChattrConfig(): CONFIG_DIR/{projectName}/agentchattr.
+    const configTomlPath = path.join(CONFIG_DIR, setup.projectName, "agentchattr", "config.toml");
     writeAgentChattrConfig(setup, configTomlPath);
 
     header("Project Added");
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -929,25 +929,33 @@ function writeAgentChattrConfig(setup, configTomlPath, { skipInstall = false } =
   fs.writeFileSync(configTomlPath, tomlContent);
   ok(`Wrote ${configTomlPath}`);
 
-  // Start AgentChattr if available; optionally skip install attempt
-  const acDir = findAgentChattr();
+  // Phase 2C / #181: clone AgentChattr per-project at
+  // ~/.quadwork/{project_id}/agentchattr/. AgentChattr's run.py loads
+  // ROOT/config.toml, so each project needs its own clone to avoid
+  // multi-instance port conflicts (see master #181). The path is the
+  // same one writeQuadWorkConfig() persists in project.agentchattr_dir.
+  const perProjectDir = path.join(CONFIG_DIR, setup.projectName, "agentchattr");
+  let acDir = findAgentChattr(perProjectDir);
   let acAvailable = !!acDir;
   if (!acAvailable && !skipInstall) {
-    const acSpinner = spinner("Setting up AgentChattr...");
-    const installResult = installAgentChattr();
+    const acSpinner = spinner(`Setting up AgentChattr at ${perProjectDir}...`);
+    const installResult = installAgentChattr(perProjectDir);
     if (installResult) {
       acSpinner.stop(true);
+      acDir = installResult;
       acAvailable = true;
     } else {
       acSpinner.stop(false);
-      warn(`Install manually: git clone ${AGENTCHATTR_REPO} ${DEFAULT_AGENTCHATTR_DIR}`);
+      const reason = installAgentChattr.lastError || "unknown error";
+      warn(`AgentChattr install failed at ${perProjectDir}: ${reason}`);
+      warn(`Install manually: git clone ${AGENTCHATTR_REPO} ${perProjectDir}`);
     }
   }
 
   // Start AgentChattr server (only if installed)
   if (acAvailable) {
     log("Starting AgentChattr server...");
-    const acSpawn = chattrSpawnArgs(findAgentChattr(), ["--config", configTomlPath]);
+    const acSpawn = chattrSpawnArgs(acDir, ["--config", configTomlPath]);
     if (acSpawn) {
       const acProc = spawn(acSpawn.command, acSpawn.spawnArgs, {
         cwd: acSpawn.cwd,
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
T2b APPROVE. +16/-8 in bin/quadwork.js, scope matches Phase 2C. perProjectDir = path.join(CONFIG_DIR, setup.projectName, 'agentchattr') matches what writeQuadWorkConfig() persists in project.agentchattr_dir (#182), so config field and actual clone location stay consistent. find→install→spawn all threaded through the same per-project path; no stale legacy reference left in this function. lastError surfaced to user on failure. Out-of-scope cmdStart/spawnChattr correctly deferred to #186; #182 runtime fallback keeps v1 setups working in the meantime.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The CLI now clones and spawns AgentChattr from the per-project directory, but it still writes `config.toml` to the repo checkout path. Because this batch exists specifically to avoid `run.py` reading the wrong ROOT config, that mismatch breaks the intended fix.

### Findings
- [high] The per-project clone still boots against the wrong config location.
  - File: `bin/quadwork.js:926`
  - Suggestion: Write the project's `config.toml` into the same per-project clone directory that you install/spawn from, not `setup.absDir/agentchattr`.

### Decision
Requesting changes because `writeAgentChattrConfig()` still writes `configTomlPath` wherever the caller passes it (`bin/quadwork.js:926-930`), and the CLI caller still passes `path.join(setup.absDir, "agentchattr", "config.toml")` (`bin/quadwork.js:1472-1473`). But this PR's own comment says AgentChattr reads `ROOT/config.toml` from the clone directory (`bin/quadwork.js:932-936`), and the process is now spawned from `acDir = ~/.quadwork/{project}/agentchattr` (`bin/quadwork.js:937-958`). So the new per-project clone will still not have the project's real config at its ROOT, which defeats the acceptance criteria for #184.

project7-interns (approved):
T2b APPROVE (re-review of ff49433). Correctly aligns configTomlPath with the per-project clone root; now run.py's ROOT/config.toml load sees the real project config, which is the whole point of Phase 2C. Path formula matches writeAgentChattrConfig()'s install location exactly. Data dir side-effect (per-project data/) is a bonus.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit aligns the CLI caller with the new per-project clone path, so the project config now lands where the spawned AgentChattr process will actually read it.

### Findings
- None.

### Decision
Approved on re-review because `cmdAddProject()` now writes `config.toml` to the same per-project clone root that `writeAgentChattrConfig()` installs and spawns from (`bin/quadwork.js:937-958`, `bin/quadwork.js:1472-1476`), which resolves the ROOT/config mismatch from the prior revision.

Issue #189, opened 2026-04-07
[#181 sub-H] Documentation: disk usage + cleanup script


realproject7:
Parent: #181

## Scope

Document the disk impact of per-project clones in README and add a cleanup helper.

## README updates

Add a section:

\`\`\`markdown
### Disk usage

Each project gets its own AgentChattr clone (~77 MB). For example:
- 1 project: ~77 MB
- 5 projects: ~385 MB

This trade-off is necessary to support multiple projects running simultaneously without port conflicts.

To free space, delete unused projects from the dashboard or run:
\`\`\`
npx quadwork cleanup --project <id>
\`\`\`
\`\`\`

## Cleanup command

Add \`bin/quadwork.js\` \`cmdCleanup()\`:
- \`npx quadwork cleanup --project <id>\` — removes \`~/.quadwork/{id}/\` and the project entry from config
- \`npx quadwork cleanup --legacy\` — removes the old shared \`~/.quadwork/agentchattr/\` after migration

## Acceptance

- ✅ README has clear disk usage docs
- ✅ \`cleanup --project <id>\` removes the per-project install + config entry
- ✅ \`cleanup --legacy\` removes the shared install only after confirming all projects have their own clones
- ✅ Cleanup commands ask for confirmation before deleting

Branch: \`task/181h-cleanup-docs\`

PR #197, opened 2026-04-07
[#188] Migrate legacy v1 projects to per-project AgentChattr clones

### LINKED ISSUE #188: [#181 sub-G] Migration: existing single-install users → per-project clones
Parent: #181
Depends on: sub-A, sub-B

## Scope

Existing v1 users have one shared install at \`~/.quadwork/agentchattr/\` and projects with no \`agentchattr_dir\` set. On startup, detect this and offer migration.

## Migration logic

For each project missing \`agentchattr_dir\`:
1. Check if \`~/.quadwork/agentchattr/\` exists (legacy install)
2. Create per-project dir \`~/.quadwork/{project_id}/agentchattr/\`
3. Run \`installAgentChattr(perProjectDir)\` (fresh clone, faster than copying)
4. Write project's config.toml into the new clone
5. Update \`config.json\` with \`agentchattr_dir\`
6. Leave the legacy \`~/.quadwork/agentchattr/\` alone for now (cleanup script in sub-H)

## When to run

Two options:
- **A) Lazy on first spawn** — when \`spawnChattr()\` finds no \`agentchattr_dir\`, migrate inline
- **B) Eager on \`quadwork start\`** — migrate all projects at startup before serving requests

Recommend B for clarity (visible progress to user, no surprises mid-session).

## Files to change

- \`bin/quadwork.js\` \`cmdStart()\` — add migration loop before AgentChattr spawn
- (or) \`server/index.js\` startup — add migration on server boot

## Acceptance

- ✅ Existing v1 user with 2 projects auto-migrates on next \`quadwork start\`
- ✅ User sees clear progress messages during migration
- ✅ Migration is idempotent (re-running is a no-op)
- ✅ User's existing project data (worktrees, repos) is untouched

Branch: \`task/181g-migration\`

realproject7:
## Summary
Phase 3 / sub-G of master #181. Existing v1 users have one shared install at \`~/.quadwork/agentchattr/\` and projects with no \`agentchattr_dir\` set. Migrate them **eagerly at the top of \`cmdStart()\`** so users see clear progress before any agents launch.

Single file: \`bin/quadwork.js\`, +93/−0 (one new function + a 4-line call site).

### Migration logic — `migrateLegacyProjects(config)`
For each project that doesn't yet have a working per-project clone:
1. \`perProjectDir = ~/.quadwork/{project_id}/agentchattr\`
2. \`installAgentChattr(perProjectDir)\` — idempotent via #183 + #187
3. **Copy** legacy \`<working_dir>/agentchattr/config.toml\` into the clone ROOT if it exists. AgentChattr's \`run.py\` reads \`ROOT/config.toml\`, so this is what makes the project actually start from its own clone after migration.
4. Set \`project.agentchattr_dir\` on the config entry and \`writeConfig()\`.

### Idempotent
A project with \`agentchattr_dir\` set + working clone + \`config.toml\` at the ROOT is skipped silently. Re-running \`cmdStart()\` on a fully-migrated config produces zero output and zero writes — \`needsMigration\` is empty, the function returns immediately.

### Safety
- Worktrees, repos, and token files are never touched.
- The legacy \`~/.quadwork/agentchattr/\` install is **left in place** — cleanup is sub-H (#189).
- Per-project failures are logged with \`installAgentChattr.lastError\` and that project is left unmigrated. The existing #186 fallback ladder (per-project clone → legacy global) keeps it running on the legacy install until the next start attempt.

## Acceptance criteria from #188
- ✅ Existing v1 user with N projects auto-migrates on next \`quadwork start\`.
- ✅ User sees clear progress messages during migration (header + per-project line + spinner).
- ✅ Migration is idempotent (re-running is a no-op).
- ✅ User's existing project data (worktrees, repos) is untouched.

## Test plan
- [ ] Stage a v1 config (one or two projects, no \`agentchattr_dir\`, with \`<wt>/agentchattr/config.toml\` populated). Run \`npx quadwork start\`. Verify each project gets a per-project clone, the legacy \`config.toml\` is copied to the new clone ROOT, and \`config.json\` is updated.
- [ ] Run \`npx quadwork start\` again. Verify zero migration output and zero writes.
- [ ] Stage a corrupt clone (rm \`.venv\` from \`~/.quadwork/foo/agentchattr\`). Run \`start\`. Verify the migration loop reuses the existing dir, recreates the venv via \`installAgentChattr\` (Phase 1B), and proceeds.
- [ ] Stage a project where the legacy \`<wt>/agentchattr/config.toml\` is missing. Verify migration still installs the clone but logs no copy step (and the project will start from \`run.py\` defaults until the user re-runs setup).
- [ ] Verify legacy \`~/.quadwork/agentchattr/\` still exists after a successful migration (cleanup is #189).

Fixes #188
Parent: #181

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1324,6 +1324,94 @@ async function cmdInit() {
 
 // ─── Start Command ──────────────────────────────────────────────────────────
 
+/**
+ * Phase 3 / #181 sub-G: migrate legacy v1 projects to per-project clones.
+ *
+ * Runs eagerly at the top of cmdStart() so users see clear progress before
+ * any agents launch. For each project that doesn't yet have a working
+ * per-project clone:
+ *   1. Compute perProjectDir = ~/.quadwork/{project_id}/agentchattr
+ *   2. installAgentChattr(perProjectDir) — idempotent (#183 + #187)
+ *   3. Copy the existing legacy <working_dir>/agentchattr/config.toml into
+ *      the new clone ROOT if it exists. AgentChattr's run.py reads
+ *      ROOT/config.toml from the clone dir, so this is what makes the
+ *      project actually start from its own clone.
+ *   4. Set project.agentchattr_dir on the config entry and persist.
+ *
+ * Idempotent: if a project already has a working per-project clone with a
+ * config.toml at the ROOT and agentchattr_dir set, it is skipped silently.
+ * The legacy ~/.quadwork/agentchattr/ install is left alone — cleanup is
+ * sub-H (#189).
+ *
+ * The migration never touches worktrees, repo content, or token files;
+ * only the per-project AgentChattr install dir and config.json.
+ */
+function migrateLegacyProjects(config) {
+  if (!config.projects || config.projects.length === 0) return false;
+
+  const needsMigration = config.projects.filter((p) => {
+    if (!p.id) return false;
+    const target = p.agentchattr_dir || path.join(CONFIG_DIR, p.id, "agentchattr");
+    const hasClone = fs.existsSync(path.join(target, "run.py")) &&
+                     fs.existsSync(path.join(target, ".venv", "bin", "python"));
+    const hasToml = fs.existsSync(path.join(target, "config.toml"));
+    const hasField = !!p.agentchattr_dir;
+    return !(hasField && hasClone && hasToml);
+  });
+
+  if (needsMigration.length === 0) return false;
+
+  header("Migrating legacy projects to per-project AgentChattr clones");
+  let mutated = false;
+  for (const project of needsMigration) {
+    const perProjectDir = path.join(CONFIG_DIR, project.id, "agentchattr");
+    log(`  ${project.id} → ${perProjectDir}`);
+
+    // 1. Install (idempotent — no-op if clone is already valid).
+    if (!findAgentChattr(perProjectDir)) {
+      const acSpinner = spinner(`    Cloning AgentChattr for ${project.id}...`);
+      const installResult = installAgentChattr(perProjectDir);
+      if (!installResult) {
+        acSpinner.stop(false);
+        const reason = installAgentChattr.lastError || "unknown error";
+        warn(`    Migration failed for ${project.id}: ${reason}`);
+        warn(`    ${project.id} will keep using the legacy global install until this is resolved.`);
+        continue;
+      }
+      acSpinner.stop(true);
+    }
+
+    // 2. Seed config.toml at the clone ROOT from the legacy in-worktree
+    //    location if present. Do not overwrite an existing per-project
+    //    config.toml — re-running the migration must be a no-op.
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
T2b APPROVE. +93/-0 in bin/quadwork.js. Migration is idempotent (skips projects with field+clone+toml all present), safe (only touches per-project install dir + config.json, never worktrees/repos/tokens), and degrades gracefully (per-project failure logged + legacy fallback from #186 keeps the project running). config.toml is copied not overwritten so re-runs are no-ops. Eager placement at top of cmdStart() means users see clear migration progress before spawn. Cleanup of ~/.quadwork/agentchattr/ correctly deferred to #189.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The migration is close, but it can still flip a project onto the per-project clone even when the root `config.toml` never made it across.

### Findings
- [high] A failed `config.toml` copy still marks the project as migrated, which breaks the promised legacy fallback.
  - File: `bin/quadwork.js:1387`
  - Suggestion: Do not persist `project.agentchattr_dir = perProjectDir` unless the per-project clone is actually ready to run from its own root config, or explicitly restore the legacy runtime path on copy failure.

### Decision
Requesting changes because the failure handling is inconsistent with the runtime behavior. If `fs.copyFileSync(legacyToml, targetToml)` throws, the code only warns (`bin/quadwork.js:1391-1395`) and then still sets `project.agentchattr_dir = perProjectDir` (`bin/quadwork.js:1400-1403`). After that, Phase 2E startup resolves `projectAcDir = findAgentChattr(project.agentchattr_dir) || legacyAcDir` (`bin/quadwork.js:1452-1456`), so it will choose the new per-project clone, not the legacy global. The PR body says failures leave the project unmigrated and still running on the legacy install, but that is not true for a clone-without-root-config case.

project7-interns (approved):
T2b APPROVE (re-review of bc15b31). tomlReady gate correctly addresses t2a's concern: agentchattr_dir is only persisted when run.py + venv + config.toml at ROOT are all in place. Three failure paths (copy throws / no legacy toml / catch-all) all 'continue' cleanly, leaving the project unmigrated so #186's fallback ladder keeps it on the legacy global install. Warning messages tell the user what to do next.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit fixes the partial-migration bug from the prior revision by gating `agentchattr_dir` persistence on the per-project clone actually having a root `config.toml` in place.

### Findings
- None.

### Decision
Approved on re-review because the migration now aborts cleanly on every missing/copy-failed `config.toml` path (`bin/quadwork.js:1394-1421`) and only persists `project.agentchattr_dir` after the per-project clone is fully runnable (`bin/quadwork.js:1423-1427`). That preserves the Phase 2E legacy fallback behavior until migration really succeeds.

Issue #225, opened 2026-04-07
[master] QuadWork UX Polish Batch — queue chip, tagline toggle, auto-tag, empty states, chat 401 fix


realproject7:
## Master Ticket — UX Polish Batch

A small batch of UX improvements + one critical bug fix. Most are independent and can be picked up in any order, except the 401 bug which is hotfix priority.

## Sub-tickets

- [ ] **#230** sub-E — **HOTFIX**: Chat send returns 401 — token sync issue blocking all chat
- [ ] **#226** sub-A — GitHub panel: vertical issues/PRs split + compact OVERNIGHT-QUEUE.md row with edit modal + auto-create on project launch
- [ ] **#227** sub-B — Tagline animation toggle (on/off) with localStorage persistence
- [ ] **#228** sub-C — Auto-tag @head when no agent is mentioned in chat
- [ ] **#229** sub-D — Empty states for home + project page + "How to Work" modal

## Recommended Order

1. **#230 first** — chat is broken without it
2. Then any order: #226, #227, #228, #229

## Acceptance

- ✅ GitHub panel splits issues/PRs vertically with queue row at bottom
- ✅ Queue file auto-created for any project that's missing it (retroactive for v1.2.0 projects)
- ✅ Tagline animation can be disabled and the choice persists across reloads
- ✅ Operator messages without an explicit agent tag get routed to Head automatically
- ✅ New users see a friendly welcome on home + projects, with a step-by-step "How to Work" modal
- ✅ Chat send works (no 401)

## References

- Master #202 (Layout v2) is the parent of these UX tweaks
- Wireframe + screenshots provided by operator

PR #231, opened 2026-04-07
[#230] HOTFIX — Chat send via WebSocket (fixes 401)

### LINKED ISSUE #230: [#225 sub-E] HOTFIX — Chat send returns 401 (token sync broken)
Parent: #225
**Priority: CRITICAL — chat is fully broken**

## Problem

Sending a message in the AgentChattr chat panel returns \`401 Unauthorized\`. The error banner from #200 surfaces it now (good UX), but the chat is unusable.

## Likely root cause

The QuadWork dashboard proxies chat sends through \`/api/chat\` to the per-project AgentChattr instance. The Authorization header / token is either:
- Not being read from \`project.agentchattr_token\` in config
- Not being forwarded to AgentChattr in the right header
- The token in config is stale (AgentChattr regenerated on restart and \`syncChattrToken\` didn't update config)

## Investigation steps

1. Reproduce: \`npx quadwork@latest start\`, open chat, click Send → check the server logs for the actual response from AgentChattr
2. Inspect \`server/index.js\` chat proxy handler — confirm it reads \`project.agentchattr_token\` and sets the right header (\`Authorization: Bearer <token>\` and/or \`X-Agent-Token: <token>\`)
3. Inspect \`syncChattrToken()\` — confirm it picks up AgentChattr's actual token after spawn
4. Inspect AgentChattr's expected auth header — see \`/Users/cho/.quadwork/{project}/agentchattr/\` for its config and source

## Files likely involved

- \`server/index.js\` — chat proxy handler (\`/api/chat\` route)
- \`server/config.js\` — \`syncChattrToken()\`
- \`server/install-agentchattr.js\` — install + config writing

## Acceptance

- ✅ Sending a message in the chat panel works (200 response)
- ✅ The message appears in the chat list within 3s
- ✅ Token sync works after AgentChattr restart
- ✅ Add a clear server-side log if auth fails so future debugging is easier

Branch: \`task/225e-fix-chat-401\`

realproject7:
## Summary
Critical hotfix for Batch 26 / master #225 sub-E. The dashboard chat panel was returning 401 on every send, making chat unusable.

Single file: \`server/routes.js\`, +93/−10.

### Root cause
The proxy hit AgentChattr's \`/api/send\`, which requires
\`Authorization: Bearer <registration_token>\` and validates the token via
\`registry.resolve_token()\`. The dashboard only stores AgentChattr's
**session_token** (the one used to authorize browser traffic), which is
**not** a registration token. \`/api/send\` therefore always returned
\`missing Authorization: Bearer <token>\` → 401.

### Fix
Mirror the path the browser dashboard already uses: send via
AgentChattr's WebSocket on \`/ws?token=<session_token>\`. The server
accepts that path with the session token, queues the message, and
broadcasts it to all connected clients. The proxy now:

1. Opens a one-shot \`ws://<chattr>/ws?token=<session_token>\`.
2. Pushes \`{type:"message", text, channel, sender, attachments}\`.
3. Waits ~250 ms for the server to enqueue, then closes cleanly.
4. **On ws close code 4003** (AgentChattr's "bad session token" code),
   re-syncs the token via \`syncChattrToken()\` and retries once.
   Recovers automatically from the "401 after AgentChattr restart"
   case because the cached token in \`config.json\` was stale.
5. Logs warn-level messages at every failure point so future
   debugging is easier (acceptance criterion).

Uses the existing \`ws\` package (already a dependency, already imported
elsewhere in \`server/index.js\`).

### Validation
- \`text\` or \`attachments\` required → 400 otherwise.
- \`sender\` defaults to \`"user"\` so dashboard messages keep their
  historical sender identity.
- The \`/api/triggers/*\` path still uses \`/api/send\` because it has its
  own bearer-token flow (the trigger acts on behalf of an agent).
  Only the dashboard chat panel was broken.

## Acceptance criteria from #230
- ✅ Sending a message in the chat panel works (200 from \`/api/chat\`).
- ✅ The message appears in the chat list within 3 s (broadcast over the
  same ws the browser is already subscribed to).
- ✅ Token sync works after AgentChattr restart (auto-retry on close
  code 4003 + \`syncChattrToken()\` re-fetch).
- ✅ Server-side log when auth fails (\`[chat] ws auth failed for project
  X, re-syncing session token and retrying...\`).

## Test plan
- [ ] \`npx quadwork start\` → open a project → type a message in the
  AgentChattr panel → click Send. Expect 200 + message visible.
- [ ] Stop and restart AgentChattr while the dashboard is open. Send
  another message. First send may log a re-sync warning, second send
  succeeds without manual intervention.
- [ ] Manually edit \`~/.quadwork/config.json\` to put a wrong
  \`agentchattr_token\` on a project. Send a message. Server logs the
  re-sync warning and the message still goes through.
- [ ] Send empty text → 400 with \`text or attachments required\`.
- [ ] curl \`POST /api/chat?project=foo\` with \`{"text":"hello"}\` →
  \`{"ok":true}\`.

Fixes #230
Parent: #225

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -179,10 +179,16 @@ router.post("/api/chat", async (req, res) => {
   const { url: base, token: sessionToken } = getChattrConfig(projectId);
   if (!base) return res.status(400).json({ error: "Missing project" });
 
+  // #230: ignore any client-supplied sender. /api/chat is the
+  // dashboard's send path, so the message must always be attributed
+  // to "user". Forwarding `req.body.sender` would let any caller
+  // hitting QuadWork's /api/chat impersonate an agent identity (t1,
+  // t3, …) over the AgentChattr ws path, which the old /api/send
+  // flow could not do.
   const message = {
     text: typeof req.body?.text === "string" ? req.body.text : "",
     channel: req.body?.channel || "general",
-    sender: req.body?.sender || "user",
+    sender: "user",
     attachments: Array.isArray(req.body?.attachments) ? req.body.attachments : [],
   };
   if (!message.text && message.attachments.length === 0) {
--- a/server/routes.js
+++ b/server/routes.js
@@ -122,19 +122,102 @@ router.get("/api/chat", async (req, res) => {
   }
 });
 
-router.post("/api/chat", async (req, res) => {
-  const { url: base, token } = getChattrConfig(req.query.project || req.body.project);
-  const tokenParam = token ? `?token=${encodeURIComponent(token)}` : "";
-  try {
-    const r = await fetch(`${base}/api/send${tokenParam}`, {
-      method: "POST",
-      headers: { "Content-Type": "application/json", ...chatAuthHeaders(token) },
-      body: JSON.stringify(req.body),
+// #225 sub-E: send chat messages from the dashboard via the
+// AgentChattr WebSocket, not via /api/send.
+//
+// /api/send requires `Authorization: Bearer <registration_token>` and
+// the token must resolve to a registered instance via
+// `registry.resolve_token()`. The session_token we store on the
+// project entry only authorizes browser/middleware traffic — it is
+// NOT a registration token, so /api/send always 401s with
+// "missing Authorization: Bearer <token>". The dashboard browser
+// already sends through the WebSocket on `/ws?token=<session_token>`
+// and the server accepts that path, so we mirror that exact flow
+// from the express server: open a one-shot ws, push the message,
+// wait briefly for ack, close.
+const { WebSocket: NodeWebSocket } = require("ws");
+const { syncChattrToken } = require("./config");
+
+function sendViaWebSocket(baseUrl, sessionToken, message) {
+  return new Promise((resolve, reject) => {
+    const wsUrl = `${baseUrl.replace(/^http/, "ws")}/ws?token=${encodeURIComponent(sessionToken || "")}`;
+    const ws = new NodeWebSocket(wsUrl);
+    let settled = false;
+    const finish = (err, value) => {
+      if (settled) return;
+      settled = true;
+      try { ws.close(); } catch {}
+      if (err) reject(err); else resolve(value);
+    };
+    const giveUp = setTimeout(() => finish(new Error("websocket send timeout")), 4000);
+    ws.on("open", () => {
+      try {
+        ws.send(JSON.stringify({ type: "message", ...message }));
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
T2b APPROVE. +93/-10 in server/routes.js. Root cause analysis is correct: /api/send needs a registration bearer token but the dashboard only has the session_token, hence the 401. Fix mirrors the browser's WS path exactly: open one-shot ws on /ws?token=<session_token>, push {type:"message",...}, close. Code 4003 → resync via syncChattrToken() and retry once, which closes the "401 after AgentChattr restart" loop where the cached token in config.json was stale. 4s overall timeout, warn logs at every failure point per the acceptance criterion. Reuses existing `ws` dep. /api/triggers/* correctly left untouched (they have their own bearer flow).

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The WebSocket switch fixes the 401 root cause, but it introduces a new sender-spoofing regression in the dashboard chat proxy. `/api/chat` now trusts the caller's `sender` field and forwards it directly into AgentChattr's WebSocket message path.

### Findings
- [high] `/api/chat` can now impersonate arbitrary chat identities.
  - File: `server/routes.js:183-187`
  - Suggestion: ignore `req.body.sender` in this route and force a fixed dashboard identity such as `"user"` before emitting over WebSocket.

### Decision
Requesting changes because the previous `/api/send` path derived `sender` from a bearer registration token, while this new WebSocket path forwards caller-controlled `sender` verbatim. That means any caller who can hit QuadWork's `/api/chat` can post as `t1`, `t3`, or any other identity, which is a real behavior/security regression even though the PR fixes the 401.

project7-interns (approved):
T2b APPROVE (re-review of 0a3d6d9). sender hard-coded to 'user' so client-supplied req.body.sender can no longer impersonate agents over the ws path. Closes the impersonation gap t2a flagged.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit fixes the impersonation regression in the dashboard chat proxy. `/api/chat` now ignores client-supplied `sender` and always attributes dashboard sends to `user`.

### Findings
- No remaining blocking findings.

### Decision
Approving because the WebSocket-based send path still fixes the original 401/token-resync problem from #230, and the new hard-coded dashboard sender closes the behavior/security regression from the prior revision.

PR #244, opened 2026-04-07
[#238] Add agentchattr-registry helper module

### LINKED ISSUE #238: [#237 sub-A] Add agentchattr-registry helper module (register, deregister, waitForReady)
Parent: #237

## Scope

Create \`server/agentchattr-registry.js\` — a small helper module that talks to AgentChattr's registration API. Used by the agent spawn paths (#237 sub-B and sub-C) to get a per-agent registration token before launching the CLI.

## API

\`\`\`javascript
// Wait for AgentChattr to be reachable. Polls GET / until 200 or timeout.
// Returns true on success, false on timeout.
async function waitForAgentChattrReady(serverPort, timeoutMs = 10000)

// Register an agent with AgentChattr. Returns { name, token, slot } or null on failure.
// Sets registerAgent.lastError on failure (matches the lastError pattern from #183).
async function registerAgent(serverPort, base, label = null)

// Deregister an agent by name. Returns true on success, false otherwise.
// Best-effort: failures are non-fatal (e.g. AgentChattr already shut down).
async function deregisterAgent(serverPort, name)
\`\`\`

## Implementation notes

- All three functions take \`serverPort\` so the helper is project-agnostic (each project has its own AgentChattr port)
- Use Node's built-in \`fetch\` (already used elsewhere in the codebase)
- Reference AgentChattr endpoints:
  - \`POST /api/register\` body: \`{base, label}\` → \`{name, token, slot, ...}\`
  - \`POST /api/deregister/{name}\` → \`{ok: true}\` or 404
  - \`GET /\` returns 200 when ready (or check \`/api/agents\` if simpler)
- Reference: \`/Users/cho/Projects/agentchattr/wrapper.py\` lines 353-364 (\`_register_instance\`)
- Use a 5s per-call timeout, configurable

## Acceptance

- ✅ Module exports all three functions
- ✅ \`waitForAgentChattrReady(8302)\` polls and returns true once AgentChattr is up
- ✅ \`registerAgent(8302, 'head')\` returns a valid \`{name, token, slot}\` after registration
- ✅ \`deregisterAgent(8302, 'head')\` removes the registration cleanly
- ✅ Failures set \`registerAgent.lastError\` with a clear message
- ✅ No external dependencies — uses built-in fetch

Branch: \`task/237a-registry-helper\`

realproject7:
Fixes #238

Adds `server/agentchattr-registry.js` with `waitForAgentChattrReady`, `registerAgent` (sets `registerAgent.lastError` on failure), and best-effort `deregisterAgent`. Uses built-in fetch with per-call timeout. Project-agnostic — serverPort is explicit on every function. Mirrors wrapper.py _register_instance.

cc @t2a @t2b

### DIFF
--- a/server/agentchattr-registry.js
+++ b/server/agentchattr-registry.js
@@ -10,6 +10,11 @@
 
 const DEFAULT_TIMEOUT_MS = 5000;
 
+// Tokens returned from registerAgent are cached per (port, name) so the
+// two-arg deregisterAgent(serverPort, name) form from the #238 contract
+// works without the caller having to thread the token through.
+const _tokenCache = new Map();
+
 function fetchWithTimeout(url, opts = {}, timeoutMs = DEFAULT_TIMEOUT_MS) {
   const ctrl = new AbortController();
   const t = setTimeout(() => ctrl.abort(), timeoutMs);
@@ -58,6 +63,7 @@ async function registerAgent(serverPort, base, label = null) {
       registerAgent.lastError = `register ${base}: malformed response`;
       return null;
     }
+    _tokenCache.set(`${serverPort}:${data.name}`, data.token);
     return { name: data.name, token: data.token, slot: data.slot };
   } catch (err) {
     registerAgent.lastError = `register ${base}: ${err.message || err}`;
@@ -72,16 +78,20 @@ registerAgent.lastError = null;
  *
  * AgentChattr's /api/deregister/{name} requires the agent's own bearer
  * token for "family" names (head/dev/reviewer1/reviewer2) — see
- * app.py:2123-2135. Pass the registration token returned by registerAgent.
+ * app.py:2123-2135. The token is looked up automatically from the cache
+ * populated by registerAgent; an explicit token may be passed as a third
+ * argument to override (e.g. recovering a session).
  */
 async function deregisterAgent(serverPort, name, token) {
   try {
     const headers = {};
-    if (token) headers["Authorization"] = `Bearer ${token}`;
+    const tok = token || _tokenCache.get(`${serverPort}:${name}`);
+    if (tok) headers["Authorization"] = `Bearer ${tok}`;
     const r = await fetchWithTimeout(
       `http://127.0.0.1:${serverPort}/api/deregister/${encodeURIComponent(name)}`,
       { method: "POST", headers },
     );
+    if (r.ok) _tokenCache.delete(`${serverPort}:${name}`);
     return r.ok;
   } catch {
     return false;
--- a/server/agentchattr-registry.js
+++ b/server/agentchattr-registry.js
@@ -69,12 +69,18 @@ registerAgent.lastError = null;
 /**
  * Best-effort deregister. Failures are non-fatal (e.g. AgentChattr already
  * shut down). Returns true on a 2xx response, false otherwise.
+ *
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
T2b APPROVE. Helper is clean: built-in fetch with AbortController timeout, explicit serverPort on every function, registerAgent.lastError populated on every failure path (HTTP, malformed, throw) and reset at entry, deregister non-fatal as required. Mirrors wrapper.py _register_instance correctly.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The registration path looks right, but the new deregistration helper does not match the live AgentChattr API. As written, `deregisterAgent(8302, "head")` will fail for the normal QuadWork agent names, so this PR misses #238's deregistration acceptance.

### Findings
- [high] `deregisterAgent()` never sends the agent registration token, but AgentChattr's `/api/deregister/{name}` path requires an authenticated agent session for family names like `head`, `dev`, and `reviewer1`.
  - File: `server/agentchattr-registry.js:74`
  - Suggestion: mirror the wrapper's shutdown flow and let the helper send `Authorization: Bearer <registration token>` on deregister, otherwise `deregisterAgent(8302, "head")` returns 403 instead of removing the registration.

### Decision
Requesting changes because the helper's deregistration API cannot satisfy the ticket's own acceptance example or the canonical wrapper behavior yet.

project7-interns (approved):
T2b re-APPROVE. deregisterAgent now accepts token and sends Authorization: Bearer header, satisfying app.py:2123 family-name auth requirement. T2a's blocker resolved.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The auth header fix addresses the original blocker, but the helper still does not match #238's required API/acceptance contract. The ticket defines `deregisterAgent(serverPort, name)`, while this PR now requires a third `token` argument for the documented `head` case to work.

### Findings
- [high] `deregisterAgent()` changed the public helper signature from the issue's required `deregisterAgent(serverPort, name)` to `deregisterAgent(serverPort, name, token)`, so the acceptance example `deregisterAgent(8302, "head")` still fails.
  - File: `server/agentchattr-registry.js:78`
  - Suggestion: keep the exported API aligned with #238. If deregistration truly needs auth for family names, either redesign the helper so callers can satisfy the existing contract without changing it, or update the implementation so the documented two-arg call still works.

### Decision
Keeping changes requested because #238 explicitly defines the helper API and this PR still misses that contract.

project7-interns (approved):
T2b re-APPROVE. Token cache restores #238 two-arg deregisterAgent(serverPort, name) signature; populated by registerAgent on success, looked up by deregister, cleared on successful deregister, optional override preserved.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The helper now matches #238's required contract and the live AgentChattr auth model. `registerAgent()` caches the assigned token per `(serverPort, name)`, so `deregisterAgent(serverPort, name)` works as specified while still sending the bearer token that `app.py` requires for family names.

### Findings
- No blocking findings.

### Decision
Approving because the remaining contract mismatch is resolved on head `3a184f0`, and the helper now satisfies #238's exported API and deregistration acceptance.

PR #245, opened 2026-04-07
[#239] Register Claude agents with AgentChattr before spawn

realproject7:
Fixes #239

For `mcp_inject="flag"` (Claude/Kimi), `buildAgentArgs()` now waits for AgentChattr readiness, calls `registerAgent(acServerPort, agentId, display_name)`, and writes `registration.token` (not `project.agentchattr_token`) into the per-agent MCP config. Failures throw with `registerAgent.lastError`.

cc @t2a @t2b

Issue #239, opened 2026-04-07
[#237 sub-B] Wire registration into Claude (mcp_inject=flag) spawn path


realproject7:
Parent: #237
Depends on: sub-A

## Scope

Modify \`buildAgentArgs()\` in \`server/index.js\` so that for Claude agents (\`mcp_inject = \"flag\"\`), we register with AgentChattr first and write the MCP config with the **registration token**, not \`project.agentchattr_token\`.

## Current code (broken)

\`\`\`javascript
// server/index.js around line 254
if (injectMode === \"flag\") {
  // Claude/Kimi: write config file, pass --mcp-config
  const mcpConfigPath = writeMcpConfigFile(projectId, agentId, mcpHttpPort, token);
  // ^^ token is project.agentchattr_token (the SESSION token, wrong)
  ...
}
\`\`\`

## New code (correct)

\`\`\`javascript
if (injectMode === \"flag\") {
  // Register with AgentChattr first to get a per-agent token
  const acServerPort = new URL(project.agentchattr_url).port || 8300;
  await waitForAgentChattrReady(acServerPort);
  const registration = await registerAgent(acServerPort, agentId, agentCfg.display_name);
  if (!registration) {
    // Surface the error so the dashboard shows it
    throw new Error(\`Failed to register \${agentId}: \${registerAgent.lastError}\`);
  }
  const mcpConfigPath = writeMcpConfigFile(projectId, agentId, mcpHttpPort, registration.token);
  args.push(agentCfg.mcp_flag || \"--mcp-config\", mcpConfigPath);
}
\`\`\`

## Files to change

- \`server/index.js\` — \`buildAgentArgs()\`, import the helpers from sub-A
- Optionally: store the registration name on the agent session so it can be passed to deregister later (sub-D)

## Acceptance

- ✅ Starting a Claude agent (e.g., reviewer2) calls \`/api/register\` first
- ✅ The MCP config file contains the registration token (different from \`project.agentchattr_token\`)
- ✅ The agent connects to AgentChattr successfully
- ✅ AgentChattr's UI shows the agent as online with green dot
- ✅ Sending \`@reviewer2 ping\` reaches the agent

Branch: \`task/237b-claude-register\`

PR #246, opened 2026-04-07
[#240] Register Codex agents with AgentChattr before spawn

### LINKED ISSUE #240: [#237 sub-C] Wire registration into Codex (mcp_inject=proxy_flag) spawn path
Parent: #237
Depends on: sub-A

## Scope

Same as sub-B but for the Codex path (\`mcp_inject = \"proxy_flag\"\`). Codex doesn't use a JSON config file — it uses the MCP auth proxy added in #178. The proxy currently uses \`project.agentchattr_token\` (session token) for the Authorization header. This must change to use the per-agent registration token.

## Current code (broken)

\`\`\`javascript
// server/index.js around line 259
} else if (injectMode === \"proxy_flag\") {
  // Codex: start local auth proxy, pass proxy URL via -c flag
  const upstreamUrl = \`http://127.0.0.1:\${mcpHttpPort}\`;
  const proxyUrl = await startMcpProxy(projectId, agentId, upstreamUrl, token);
  // ^^ token is project.agentchattr_token (wrong)
  ...
}
\`\`\`

## New code

\`\`\`javascript
} else if (injectMode === \"proxy_flag\") {
  const acServerPort = new URL(project.agentchattr_url).port || 8300;
  await waitForAgentChattrReady(acServerPort);
  const registration = await registerAgent(acServerPort, agentId, agentCfg.display_name);
  if (!registration) {
    throw new Error(\`Failed to register \${agentId}: \${registerAgent.lastError}\`);
  }
  const upstreamUrl = \`http://127.0.0.1:\${mcpHttpPort}\`;
  const proxyUrl = await startMcpProxy(projectId, agentId, upstreamUrl, registration.token);
  args.push(\"-c\", \`mcp_servers.agentchattr.url=\"\${proxyUrl}\"\`);
}
\`\`\`

## Notes

- The MCP auth proxy from #178 already accepts a per-agent token at creation time (no proxy changes needed — just pass the right token)
- Need to ensure \`startMcpProxy\` is called fresh each spawn (not reused from a stale call) so the new token takes effect
- If \`stopMcpProxy\` is missing or buggy, agents that restart may end up using a stale proxy with the wrong token

## Acceptance

- ✅ Starting a Codex agent (head) calls \`/api/register\` first
- ✅ The MCP proxy uses the registration token
- ✅ The agent connects to AgentChattr successfully
- ✅ AgentChattr's UI shows the agent as online
- ✅ Sending \`@head ping\` reaches Head and returns a response

Branch: \`task/237c-codex-register\`

realproject7:
Fixes #240

For `mcp_inject="proxy_flag"` (Codex), `buildAgentArgs()` now waits for AgentChattr readiness, calls `registerAgent`, and passes `registration.token` into `startMcpProxy` instead of `project.agentchattr_token`. Failures throw with `registerAgent.lastError`.

cc @t2a @t2b

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -266,9 +266,16 @@ async function buildAgentArgs(projectId, agentId) {
       const flag = agentCfg.mcp_flag || "--mcp-config";
       args.push(flag, mcpConfigPath);
     } else if (injectMode === "proxy_flag") {
-      // Codex: start local auth proxy, pass proxy URL via -c flag
+      // Codex: register with AgentChattr first (#240) so the proxy
+      // injects a real per-agent token, not the global session token.
+      const acServerPort = Number(new URL(project.agentchattr_url).port) || 8300;
+      await waitForAgentChattrReady(acServerPort);
+      const registration = await registerAgent(acServerPort, agentId, agentCfg.display_name || null);
+      if (!registration) {
+        throw new Error(`Failed to register ${agentId}: ${registerAgent.lastError}`);
+      }
       const upstreamUrl = `http://127.0.0.1:${mcpHttpPort}`;
-      const proxyUrl = await startMcpProxy(projectId, agentId, upstreamUrl, token);
+      const proxyUrl = await startMcpProxy(projectId, agentId, upstreamUrl, registration.token);
       if (proxyUrl) {
         args.push("-c", `mcp_servers.agentchattr.url="${proxyUrl}"`);
       }
--- a/server/index.js
+++ b/server/index.js
@@ -268,7 +268,10 @@ async function buildAgentArgs(projectId, agentId) {
     } else if (injectMode === "proxy_flag") {
       // Codex: register with AgentChattr first (#240) so the proxy
       // injects a real per-agent token, not the global session token.
-      const acServerPort = Number(new URL(project.agentchattr_url).port) || 8300;
+      // Resolve via resolveProjectChattr so legacy/global-config
+      // projects without a per-project agentchattr_url still work.
+      const chattrInfo = resolveProjectChattr(projectId);
+      const acServerPort = Number(new URL(chattrInfo.url).port) || 8300;
       await waitForAgentChattrReady(acServerPort);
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
T2b APPROVE. Codex proxy_flag path now registers before spawn, throws with registerAgent.lastError on failure, and passes registration.token into startMcpProxy instead of the global session token. Mirrors the #239 flag-mode pattern.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The Codex proxy path now uses the right token type once registration succeeds, but the new readiness/registration step bypasses QuadWork's existing AgentChattr URL fallback. That creates a startup regression for projects that still rely on the global `config.agentchattr_url` path.

### Findings
- [high] `buildAgentArgs()` now reads `project.agentchattr_url` directly for the new `URL(...).port` parse, so any existing project without a per-project `agentchattr_url` now throws before spawn instead of using the global fallback that `resolveProjectChattr()` already provides.
  - File: `server/index.js:272`
  - Suggestion: resolve the AgentChattr URL through `resolveProjectChattr(projectId)` here, the same way the rest of the server does, then parse that resolved URL. That preserves backward compatibility for legacy/global-config projects while still registering before proxy startup.

### Decision
Requesting changes because this is a real start-path regression in the new Codex registration flow.

project7-interns (approved):
T2b re-APPROVE. URL now resolved via resolveProjectChattr(projectId), preserving legacy/global-config fallback. Note: the merged #245 flag-mode path still parses project.agentchattr_url directly and has the same regression — worth a follow-up.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The Codex `mcp_inject="proxy_flag"` path now registers before proxy startup, passes the returned registration token into `startMcpProxy()`, and resolves the AgentChattr URL through the shared fallback path so legacy/global-config projects still start correctly.

### Findings
- No blocking findings.

### Decision
Approving because head `3fdaf2f` resolves the start-path regression and now matches #240's required registration/proxy behavior.

PR #247, opened 2026-04-07
[#241] Deregister agents on stop to free AgentChattr slots

### LINKED ISSUE #241: [#237 sub-D] Deregister agents on stop/restart to free slots
Parent: #237
Depends on: sub-A, sub-B, sub-C

## Scope

When an agent is stopped or restarted via the dashboard, deregister it from AgentChattr. Otherwise the slot stays occupied and the next registration becomes \`head-2\` instead of replacing the old \`head\`.

## Files to change

- \`server/index.js\` — \`stopAgentSession()\` (or wherever the PTY cleanup happens)
- Track the registered name on the session object (added in sub-B/C) so deregister knows what to send

## Implementation

\`\`\`javascript
function stopAgentSession(key) {
  const session = agentSessions.get(key);
  if (!session) return;
  // ... existing PTY cleanup ...

  // Deregister from AgentChattr (best-effort)
  if (session.acRegistrationName) {
    const project = readConfig().projects?.find(p => p.id === session.projectId);
    if (project) {
      const acPort = new URL(project.agentchattr_url).port || 8300;
      deregisterAgent(acPort, session.acRegistrationName).catch(() => {});
    }
  }

  // ... existing MCP proxy cleanup ...
}
\`\`\`

## Acceptance

- ✅ Stopping an agent calls \`POST /api/deregister/{name}\`
- ✅ Restarting an agent: stop → deregister → register (fresh slot 1) → start
- ✅ AgentChattr's UI no longer shows the stopped agent
- ✅ Failures are non-fatal (e.g. AgentChattr already shut down) — just log

Branch: \`task/237d-deregister-on-stop\`

realproject7:
Fixes #241

`buildAgentArgs` now returns `{args, acRegistrationName, acServerPort}`. `spawnAgentPty` stores those on the session. `stopAgentSession` calls `deregisterAgent` best-effort. Failures are non-fatal. Restart flow becomes stop → deregister → fresh register → start.

Also rolls in the t2b-flagged follow-up from #245: the Claude flag-mode block now resolves the AgentChattr URL via `resolveProjectChattr(projectId).url` so legacy/global-config projects don't crash on missing per-project `agentchattr_url`.

cc @t2a @t2b

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -8,7 +8,7 @@ const pty = require("node-pty");
 const { spawn } = require("child_process");
 const { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr, resolveChattrSpawn, syncChattrToken, CONFIG_PATH } = require("./config");
 const routes = require("./routes");
-const { waitForAgentChattrReady, registerAgent } = require("./agentchattr-registry");
+const { waitForAgentChattrReady, registerAgent, deregisterAgent } = require("./agentchattr-registry");
 
 const net = require("net");
 const config = readConfig();
@@ -234,12 +234,14 @@ function writeMcpConfigFile(projectId, agentId, mcpHttpPort, token) {
 async function buildAgentArgs(projectId, agentId) {
   const cfg = readConfig();
   const project = cfg.projects?.find((p) => p.id === projectId);
-  if (!project) return [];
+  if (!project) return { args: [], acRegistrationName: null, acServerPort: null };
 
   const agentCfg = project.agents?.[agentId] || {};
   const command = agentCfg.command || "claude";
   const cliBase = command.split("/").pop().split(" ")[0]; // extract base CLI name
   const args = [];
+  let acRegistrationName = null;
+  let acServerPort = null;
 
   // Permission bypass flags
   if (agentCfg.auto_approve !== false) {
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
T2b APPROVE. Session now tracks acRegistrationName/acServerPort, stopAgentSession best-effort deregisters with .catch() (non-fatal), and clears the name to prevent double-deregister. Early-return shape updated. Bonus #245 follow-up rolled in (flag block now uses resolveProjectChattr).

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR tracks the registration name and adds best-effort deregistration, but restart still does not enforce the ordering that #241 explicitly requires. As written, the new deregister call is fire-and-forget, so a restart can still re-register before the old slot is actually released.

### Findings
- [high] `stopAgentSession()` fires `deregisterAgent(...)` without awaiting it, and the restart endpoint launches the fresh spawn 500ms later regardless of whether deregistration has completed. That means the required `stop -> deregister -> register` ordering is not guaranteed, so a slow deregister response can still produce `head-2`/`reviewer2-2` on restart.
  - File: `server/index.js:406`
  - File: `server/index.js:687`
  - Suggestion: make the stop/restart cleanup path await deregistration completion before spawning again. Best-effort is still fine for error handling, but the happy-path restart order needs to be serialized to satisfy #241.

### Decision
Requesting changes because the current implementation still misses #241's core restart acceptance under normal timing variance.

project7-interns (approved):
T2b re-APPROVE. stopAgentSession is now async and awaits deregisterAgent; both callers (/stop and /restart) await it; the 500ms setTimeout race is gone. Verified only two callers exist. #241 ordering acceptance met.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The stop/restart cleanup path now preserves the required `stop -> deregister -> register -> start` ordering from #241. `stopAgentSession()` is async, awaits best-effort deregistration, and both stop/restart routes await it before returning or respawning.

### Findings
- No blocking findings.

### Decision
Approving because head `7102c14` resolves the restart race and also rolls in the needed `resolveProjectChattr()` fallback fix for the Claude flag-mode path.

PR #248, opened 2026-04-07
[#242] Deregister stale slot before fresh register

### LINKED ISSUE #242: [#237 sub-E] Handle slot collisions — deregister stale entries before re-registering
Parent: #237
Depends on: sub-A, sub-B, sub-C

## Scope

If an agent crashed or was killed without going through the deregister path (sub-D), its slot remains occupied in AgentChattr's registry. The next time we try to register \`head\`, we'll get back \`head-2\` instead of \`head\`. The chat router only knows about \`head\`, so messages to \`@head\` go to the wrong instance.

## Fix

Before calling \`registerAgent\`, attempt to deregister the canonical name first. If the registration is stale (no live process), deregister succeeds and clears the slot. If a live registration exists (rare — only happens in race conditions), the deregister still works and we re-register fresh.

## Implementation

In \`buildAgentArgs()\` (modified by sub-B and sub-C), wrap the registration:

\`\`\`javascript
// Best-effort cleanup of any stale registration before fresh register.
// Handles the case where a previous QuadWork session crashed without
// running the stopAgentSession deregister cleanup.
await deregisterAgent(acServerPort, agentId).catch(() => {});

const registration = await registerAgent(acServerPort, agentId, ...);
\`\`\`

## Edge cases to consider

1. **Multiple QuadWork sessions on the same project** — could deregister each other's live agents. Mitigation: this is a single-user local tool; concurrent QuadWork instances on the same project are out of scope.

2. **Stale registration after AgentChattr restart** — when AgentChattr restarts, the registry is cleared. Stale registration shouldn't exist. The deregister call becomes a no-op (404), which is fine.

3. **Slot 1 rename behavior** — \`registry.py\` line 117 renames slot 1 to \`base-1\` when slot 2 registers. We never want this to happen since QuadWork only ever runs one instance per agent base. Cleaning the slot first prevents this.

## Acceptance

- ✅ After a crash, restarting QuadWork → next agent start re-registers as canonical name (e.g., \`head\`, not \`head-2\`)
- ✅ AgentChattr UI shows only the live registration, no orphans
- ✅ Deregister-then-register is atomic enough that messages don't get lost during the gap

Branch: \`task/237e-slot-collision\`

realproject7:
Fixes #242

Both Claude and Codex registration paths now call `deregisterAgent(acServerPort, agentId).catch(() => {})` before `registerAgent`. Preserves canonical names after a crash so chat routing stays on `head`/`dev`/`reviewer*` bases instead of cascading to `head-2`.

cc @t2a @t2b

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -261,6 +261,11 @@ async function buildAgentArgs(projectId, agentId) {
       const chattrInfo = resolveProjectChattr(projectId);
       acServerPort = Number(new URL(chattrInfo.url).port) || 8300;
       await waitForAgentChattrReady(acServerPort);
+      // #242: best-effort deregister any stale registration of the
+      // canonical name (left over by a crashed previous QuadWork
+      // session) so the fresh register lands at slot 1 instead of
+      // head-2 / reviewer2-2. Failures are non-fatal.
+      await deregisterAgent(acServerPort, agentId).catch(() => {});
       const registration = await registerAgent(acServerPort, agentId, agentCfg.display_name || null);
       if (!registration) {
         throw new Error(`Failed to register ${agentId}: ${registerAgent.lastError}`);
@@ -277,6 +282,8 @@ async function buildAgentArgs(projectId, agentId) {
       const chattrInfo = resolveProjectChattr(projectId);
       acServerPort = Number(new URL(chattrInfo.url).port) || 8300;
       await waitForAgentChattrReady(acServerPort);
+      // #242: best-effort deregister stale canonical name first.
+      await deregisterAgent(acServerPort, agentId).catch(() => {});
       const registration = await registerAgent(acServerPort, agentId, agentCfg.display_name || null);
       if (!registration) {
         throw new Error(`Failed to register ${agentId}: ${registerAgent.lastError}`);
--- a/server/index.js
+++ b/server/index.js
@@ -208,6 +208,39 @@ const PERMISSION_FLAGS = {
  * Generate a per-agent MCP config file for Claude (--mcp-config).
  * Returns the absolute path to the written JSON file.
  */
+/**
+ * Per-agent registration tokens persisted across QuadWork restarts so
+ * #242 stale-slot reclaim works after a crash. Without this the
+ * in-memory _tokenCache is empty on startup and the family-name
+ * deregister returns 403 (app.py:2123-2135).
+ */
+function _agentTokenPath(projectId, agentId) {
+  const configDir = path.join(os.homedir(), ".quadwork", projectId);
+  return path.join(configDir, `agent-token-${agentId}.txt`);
+}
+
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
T2b REQUEST CHANGES. Concern: after a crash, the in-process _tokenCache in agentchattr-registry.js is empty, so this best-effort deregisterAgent(port, agentId) sends no Authorization header. Per app.py:2123 (cited in #244), family-name deregister requires the agent's own bearer token and returns 403 otherwise — meaning the stale 'head' registration is NOT actually cleared and the fresh register still lands at head-2. Need a server-side bypass (admin endpoint, or session-token-authorized deregister) or persist tokens across restarts. Please confirm whether AgentChattr exposes an unauthenticated path for stale-slot reclaim.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new pre-register stale-slot cleanup does not actually clear stale canonical registrations after a crash. In a fresh QuadWork process the helper has no cached per-agent token, so the new `deregisterAgent(acServerPort, agentId)` call hits AgentChattr without auth and the stale `head`/`dev`/`reviewer*` slot remains occupied.

### Findings
- [high] After a crash, `buildAgentArgs()` now calls `deregisterAgent(acServerPort, agentId)` before `registerAgent`, but `deregisterAgent()` only sends an Authorization header if it already has a cached token for that `(port,name)`. On a fresh QuadWork restart that cache is empty, and AgentChattr rejects unauthenticated family-name deregistration with 403, so the stale canonical slot is not cleared and the next register still lands at `head-2` / `reviewer2-2`.
  - File: `server/index.js:269`
  - File: `server/agentchattr-registry.js:86`
  - Suggestion: use a reclaim path that can authenticate stale-slot cleanup across process restarts, or persist/recover the previous registration token before attempting the pre-register deregister.

### Decision
Requesting changes because the current implementation still misses #242's crash-recovery acceptance for canonical-name preservation.

project7-interns (approved):
T2b re-APPROVE. Persisted token solution: written to ~/.quadwork/<project>/agent-token-<agent>.txt mode 0600 after successful register, read before stale-slot deregister so the call carries Authorization: Bearer (satisfying app.py:2123), cleared after successful stale-deregister and on stopAgentSession. os imported at module top. #242 acceptance now achievable across crashes.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The stale-slot reclaim path now survives a fresh QuadWork restart. Registration tokens are persisted per project/agent, loaded before the pre-register `deregisterAgent(...)` call, and written back after a fresh registration, so the family-name reclaim request can authenticate and free the canonical slot after a crash.

### Findings
- No blocking findings.

### Decision
Approving because head `fe75bc2` resolves the crash-recovery auth gap and now matches #242's canonical-name preservation goal.

Issue #254, opened 2026-04-08
[#249 sub-E] End-to-end verification — fresh project, all 4 agents respond to chat


realproject7:
Parent: #249
Depends on: sub-A, sub-B, sub-C, sub-D

## Scope

Final end-to-end verification of master #249. This is the second attempt at #243 (which was blocked by the bugs in this batch). Once this passes, both #249 and master #237 are fully resolved.

## Test scenario

1. **Clean slate**: \`rm -rf ~/.quadwork/dropcast\`, remove existing dropcast worktrees
2. **Web wizard**: open dashboard, run setup wizard → create project \`dropcast\` pointing at \`realproject7/dropcast\`
3. **Verify auto-spawn (sub-C)**: confirm AgentChattr process is running for dropcast on the project's port
   - \`ps aux | grep \"python.*run.py.*dropcast\"\`
   - \`lsof -i :{project_port}\` shows the AC process LISTEN
4. **Start all 4 agents** from the dashboard
5. **Verify registration (Master #237)**:
   - \`curl http://127.0.0.1:{ac_port}/api/agents\` shows all 4 names with \`active\` state
   - Each MCP config file contains the registration token
6. **Verify heartbeats (sub-A)**: wait 90 seconds, then re-check \`/api/agents\`. All 4 should still be \`active\` (no \"disconnected (timeout)\" messages in chat)
7. **Verify queue watcher (sub-B)**: type \`@head are you online?\` in the chat. Head should respond within ~30 seconds with actual content
8. **Verify all agents respond**:
   - \`@reviewer1 list open issues\` → reviewer1 responds
   - \`@dev hello\` → dev responds
9. **Verify 409 recovery (sub-D)**: restart AgentChattr from the dashboard. Wait 15 seconds. Send \`@head ping\` again. Head should respond.
10. **Verify clean stop**: stop one agent. Verify it deregisters cleanly (no orphan in \`/api/agents\`).

## Document results in PR description

Post:
- Output of \`lsof -i :{port} | grep LISTEN\` showing the AC process
- Output of \`curl /api/agents\` after step 4 (all 4 active) and step 6 (still all 4 active after 90s)
- Screenshots of the dashboard showing all 4 status dots green
- The actual chat responses from steps 7, 8, 9

## Acceptance

- ✅ Web wizard auto-spawns AgentChattr (no manual Restart needed)
- ✅ All 4 agents register successfully on first start
- ✅ Agents stay active after 90 seconds (heartbeat works)
[...]

Issue #253, opened 2026-04-08
[#249 sub-D] Re-register agents on heartbeat 409 (AgentChattr restart recovery)


realproject7:
Parent: #249
Depends on: sub-A

## Problem

When AgentChattr restarts (e.g., user clicks \"Restart\" in the dashboard, or the process crashes and gets re-spawned), the in-memory registry is wiped. Agents that were registered before the restart now have stale tokens and stale identity. Their heartbeats start returning HTTP 409 (Conflict — \"unknown agent\").

\`wrapper.py\` handles this gracefully: on 409, it calls \`_register_instance\` again to get a fresh token and identity, then keeps heartbeating with the new token. \`set_runtime_identity\` updates the in-memory state and rewrites the MCP config file (so the next restart uses the right token).

## Reference

\`/Users/cho/Projects/agentchattr/wrapper.py\` lines 732-741:

\`\`\`python
except urllib.error.HTTPError as exc:
    if exc.code == 409:
        try:
            replacement = _register_instance(server_port, agent, args.label)
            set_runtime_identity(replacement[\"name\"], replacement[\"token\"])
            _notify_recovery(data_dir, replacement[\"name\"])
        except Exception:
            pass
    time.sleep(5)
    continue
\`\`\`

## Scope

In the heartbeat function (sub-A), detect 409 responses and trigger a re-registration. Update the session's tracked registration name + token. Rewrite the MCP config file (Claude path) or update the proxy token (Codex path).

## Files to change

- \`server/agentchattr-registry.js\`:
  - Modify \`startHeartbeat\` to accept a \`onConflict\` callback
  - When the heartbeat POST returns 409, call \`onConflict()\` and don't kill the interval — keep trying with the new token after recovery

- \`server/index.js\`:
  - When starting the heartbeat for an agent, pass an \`onConflict\` callback that:
    1. Calls \`registerAgent\` to get a new \`{name, token}\`
    2. Updates \`session.acRegistrationName\` and \`session.acRegistrationToken\`
    3. For Claude: rewrites the MCP config file (\`mcp-{agent}.json\`) with the new token
    4. For Codex: kills the existing MCP proxy and starts a new one with the new token (note: the agent CLI is already pointing at the proxy URL, so the proxy can be replaced behind it without restarting the agent)
    5. Restarts the queue watcher with the new agent name (in case it changed)

## Edge cases

[...]

Issue #249, opened 2026-04-08
[master] Port wrapper.py logic — heartbeat + queue watcher + auto-spawn


realproject7:
## Master Ticket — Port wrapper.py logic to QuadWork

After Master #237 (registration auth fix) merged, agents still appear offline. Deeper investigation revealed that #237 only fixed the registration step. There are several more behaviors from `agent-os/agentchattr/wrapper.py` that QuadWork is missing.

## Background

`wrapper.py` is the canonical AgentChattr launcher. It does much more than just register agents. The full per-agent lifecycle includes:

| Feature | wrapper.py | QuadWork status |
|---------|-----------|----------------|
| Register agent | line 568 | ✅ #239, #240 |
| Persist token to disk | implicit | ✅ #248 |
| **Send heartbeat every 5s** | line 715-748 | ❌ MISSING |
| **Watch queue file and inject prompts to CLI** | line 438-541 | ❌ MISSING |
| **Re-register on 409 conflict** | line 732-741 | ❌ MISSING |
| Auto-spawn AgentChattr after web setup | (CLI only) | ❌ MISSING |

## Critical insight

**The queue watcher is the actual mechanism by which agents receive chat messages.** When you `@head` in chat, AgentChattr writes a job to `{data_dir}/head_queue.jsonl`. The wrapper polls this file every ~1s and, when it sees content, **injects `mcp read #general - you were mentioned` into the agent's terminal via PTY write**. The agent then reads the chat via its MCP connection and responds.

QuadWork has no equivalent. Even if registration + heartbeat work perfectly, the agents will never see chat messages because nothing reads the queue file and pokes the CLI.

## Why we have node-pty already

QuadWork already manages the agent terminals via `node-pty` (`spawnAgentPty` in `server/index.js`). We can write to the PTY from a queue watcher using `session.term.write(...)`. This makes implementing the queue watcher straightforward — we just need a per-agent file poller that writes to the existing PTY session.

## Sub-tickets

- [ ] **#250** sub-A — Heartbeat sender: track registration, POST /api/heartbeat/{name} every 5s, clear on stop
- [ ] **#251** sub-B — Queue watcher: per-agent file poller that injects `mcp read` prompts into the PTY when the queue has content
- [ ] **#252** sub-C — Auto-spawn AgentChattr after web wizard `add-config` succeeds (CLI wizard already does this)
- [ ] **#253** sub-D — Re-register on 409 (handles AgentChattr restart while agent is running)
[...]

Issue #237, opened 2026-04-07
[master] Fix agent auth — use AgentChattr /api/register flow instead of session token


realproject7:
## Background

Agents launched by QuadWork show as "offline" in the AgentChattr chat. Investigation traced this to a fundamental misunderstanding of AgentChattr's auth model.

## Root Cause

AgentChattr has **two distinct token types**:

1. **Session token** (`config.session_token`, exposed via `__SESSION_TOKEN__` in HTML) — used for **browser UI** authentication only. AgentChattr generates a fresh random session token at every startup (`run.py` line 32: `session_token = secrets.token_hex(32)`) and **completely ignores** any `session_token` field in `config.toml`.

2. **Per-agent registration token** — assigned by AgentChattr when an agent registers via `POST /api/register`. Each agent instance has its own `Instance.token` (`registry.py` line 28). This is what agents use for MCP authentication.

QuadWork has been writing the **session token** into the per-agent MCP config files (`mcp-head.json` etc.) since #177. This is the wrong token type — agents fail MCP auth and never connect.

## Reference Implementation

`/Users/cho/Projects/agentchattr/wrapper.py` is the canonical launcher. Its flow:

1. Start AgentChattr (or assume it's running)
2. `POST /api/register` with `{base: "head", label: ...}`
3. Receive `{name: "head", token: "T", slot: 1}`
4. Write MCP config with `Authorization: Bearer T`
5. Spawn the CLI with `--mcp-config <path>`

QuadWork must port this flow into its agent spawn logic.

## Why nobody caught this

The MCP config files in `/Users/cho/Projects/agentchattr/data/provider-config/` look like they have stale tokens, but actually they were rewritten by `wrapper.py` at the most recent registration. Their tokens don't match the current `__SESSION_TOKEN__` because **they're not supposed to** — they're registration tokens, not session tokens.

QuadWork copied the file format but bypassed the registration step, using the wrong token entirely.

## Sub-tickets

- [ ] **#238** sub-A — Add `server/agentchattr-registry.js` helper module (register, deregister, waitForReady)
- [ ] **#239** sub-B — Wire registration into Claude (`mcp_inject = "flag"`) agent spawn path
- [ ] **#240** sub-C — Wire registration into Codex (`mcp_inject = "proxy_flag"`) agent spawn path
- [ ] **#241** sub-D — Deregister agents on stop/restart to free slots
- [ ] **#242** sub-E — Handle slot collisions: deregister stale entries before re-registering
[...]

PR #255, opened 2026-04-08
[#250] sub-A: Heartbeat sender keeps registered agents alive

### LINKED ISSUE #250: [#249 sub-A] Heartbeat sender — keep registered agents alive
Parent: #249

## Problem

QuadWork registers agents via \`/api/register\` but never sends heartbeats. AgentChattr's \`app.py\` line 361-385 has a 60-second crash timeout that auto-deregisters agents that haven't heartbeated. After 60 seconds, the chat shows \"head disconnected (timeout)\" for all 4 agents.

## Reference

\`/Users/cho/Projects/agentchattr/wrapper.py\` lines 715-748:

\`\`\`python
def _heartbeat():
    while True:
        current_name, _ = get_identity()
        current_token = get_token()
        url = f\"http://127.0.0.1:{server_port}/api/heartbeat/{current_name}\"
        try:
            req = urllib.request.Request(url, method=\"POST\", data=b\"\",
                                          headers=_auth_headers(current_token))
            with urllib.request.urlopen(req, timeout=5) as resp:
                resp_data = json.loads(resp.read())
            ...
        except urllib.error.HTTPError as exc:
            if exc.code == 409:
                # Re-register flow — sub-D handles this
                ...
            time.sleep(5)
            continue
        time.sleep(5)
threading.Thread(target=_heartbeat, daemon=True).start()
\`\`\`

## Scope

Add a per-agent heartbeat sender to QuadWork. When an agent registers (in \`buildAgentArgs\`), start an interval that POSTs \`/api/heartbeat/{name}\` every 5 seconds with the registration token. When the agent stops (\`stopAgentSession\`), clear the interval.

## Files to change

- \`server/agentchattr-registry.js\`:
  - Add \`startHeartbeat(serverPort, name, token)\` — returns an interval handle
  - Add \`stopHeartbeat(handle)\` — clears the interval
  - Each handle should track its name + port for debugging
  - Use \`fetch\` with 5s timeout, swallow errors so transient failures don't crash QuadWork

- \`server/index.js\`:
  - Track heartbeat handles per agent session (e.g., \`session.heartbeatHandle\`)
  - In \`buildAgentArgs\` (or wherever the registration succeeds), call \`startHeartbeat\`
  - In \`stopAgentSession\`, call \`stopHeartbeat\` before \`deregisterAgent\`

## Implementation sketch

\`\`\`javascript
// agentchattr-registry.js
function startHeartbeat(serverPort, name, token) {
  const intervalMs = 5000;
  const url = \`http://127.0.0.1:\${serverPort}/api/heartbeat/\${encodeURIComponent(name)}\`;
  const headers = { \"Authorization\": \`Bearer \${token}\` };
  const tick = async () => {
    try {
      const ctrl = new AbortController();
      const timer = setTimeout(() => ctrl.abort(), 5000);
      await fetch(url, { method: \"POST\", headers, signal: ctrl.signal });
      clearTimeout(timer);
    } catch { /* swallow */ }
  };
  tick(); // immediate
  return setInterval(tick, intervalMs);
}

function stopHeartbeat(handle) {
  if (handle) clearInterval(handle);
}
\`\`\`

## Acceptance

- ✅ \`startHeartbeat\` POSTs every 5s with the bearer token
- ✅ Errors are swallowed (transient AC restart shouldn't crash QuadWork)
- ✅ Each agent session has its own heartbeat
- ✅ Stopping an agent stops its heartbeat
- ✅ Verify with \`curl http://127.0.0.1:{ac_port}/api/agents\` — agents stay \`active\` indefinitely (no \"disconnected (timeout)\" after 60s)

Branch: \`task/249a-heartbeat\`

realproject7:
Fixes #250
Tracker: realproject7/agent-os#391
Master: #249

## Summary
- Add `startHeartbeat`/`stopHeartbeat` to `server/agentchattr-registry.js` — per-agent 5s POST to `/api/heartbeat/{name}` with bearer auth, transient failures swallowed.
- Thread the registration token through `buildAgentArgs` → session, start the heartbeat in `spawnAgentPty` right after registration, stop it in `stopAgentSession` before deregister.
- Mirrors `agent-os/agentchattr/wrapper.py` lines 715-748. The 409 identity-wiped recovery path is intentionally left to sub-D (#253).

## Why
Master #237 only fixed registration auth — agents still fall offline at the 60s crash-detection window because nothing was holding them alive. Without heartbeats the queue watcher in sub-B has no one to deliver to.

## Acceptance criteria
- [x] Each registered agent starts its own heartbeat immediately after registration
- [x] Heartbeats POST every 5 seconds with the registration token
- [x] Transient heartbeat failures do not crash QuadWork (try/catch around fetch)
- [x] Stopping or restarting an agent clears its heartbeat handle (stopHeartbeat in stopAgentSession before deregister)
- [ ] Verification shows agents remain `active` past 90 seconds — covered by sub-E (#254) operator-gated e2e

## Test plan
- [x] `node --check` on both modified files
- [ ] Reviewers: clone quadwork, spin up a project with AgentChattr, register an agent, leave it idle 90s, confirm it stays `active` instead of disconnecting at 60s
- [ ] Reviewers: stop the agent, confirm interval is cleared (no further POSTs in AC logs)

### DIFF
--- a/server/agentchattr-registry.js
+++ b/server/agentchattr-registry.js
@@ -98,8 +98,48 @@ async function deregisterAgent(serverPort, name, token) {
   }
 }
 
+/**
+ * Start a per-agent heartbeat that POSTs /api/heartbeat/{name} every 5s
+ * with bearer auth. AgentChattr considers an agent crashed and removes
+ * it after ~60s without a heartbeat, so without this every registered
+ * QuadWork agent silently disappears from the channel one minute after
+ * registration.
+ *
+ * Returns an opaque handle suitable for stopHeartbeat. Transient errors
+ * (network blips, AgentChattr restart) are swallowed — the next tick
+ * just tries again. The 409 "identity wiped" recovery flow is sub-D
+ * (quadwork#253) and is intentionally not handled here.
+ *
+ * Reference: /Users/cho/Projects/agentchattr/wrapper.py lines 715-748.
+ */
+function startHeartbeat(serverPort, name, token, intervalMs = 5000) {
+  const url = `http://127.0.0.1:${serverPort}/api/heartbeat/${encodeURIComponent(name)}`;
+  const headers = { Authorization: `Bearer ${token}` };
+  const tick = async () => {
+    try {
+      await fetchWithTimeout(url, { method: "POST", headers }, DEFAULT_TIMEOUT_MS);
+    } catch {
+      // swallow — next tick will retry
+    }
+  };
+  // Fire one immediately so a fresh agent is held alive without waiting
+  // a full interval, then poll on the timer.
+  tick();
+  const handle = setInterval(tick, intervalMs);
+  return handle;
+}
+
+/**
+ * Stop a heartbeat started by startHeartbeat. Safe to call with null.
+ */
+function stopHeartbeat(handle) {
+  if (handle) clearInterval(handle);
+}
+
 module.exports = {
   waitForAgentChattrReady,
   registerAgent,
   deregisterAgent,
+  startHeartbeat,
+  stopHeartbeat,
 };
--- a/server/index.js
+++ b/server/index.js
@@ -8,7 +8,7 @@ const pty = require("node-pty");
 const { spawn } = require("child_process");
 const { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr, resolveChattrSpawn, syncChattrToken, CONFIG_PATH } = require("./config");
 const routes = require("./routes");
-const { waitForAgentChattrReady, registerAgent, deregisterAgent } = require("./agentchattr-registry");
+const { waitForAgentChattrReady, registerAgent, deregisterAgent, startHeartbeat, stopHeartbeat } = require("./agentchattr-registry");
 
 const net = require("net");
 const config = readConfig();
@@ -267,7 +267,7 @@ function writeMcpConfigFile(projectId, agentId, mcpHttpPort, token) {
 async function buildAgentArgs(projectId, agentId) {
   const cfg = readConfig();
   const project = cfg.projects?.find((p) => p.id === projectId);
-  if (!project) return { args: [], acRegistrationName: null, acServerPort: null };
+  if (!project) return { args: [], acRegistrationName: null, acServerPort: null, acRegistrationToken: null };
 
   const agentCfg = project.agents?.[agentId] || {};
   const command = agentCfg.command || "claude";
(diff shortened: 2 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Heartbeat startup is wired correctly, but cleanup only runs on explicit `/stop` and `/restart`. If the agent PTY exits on its own, QuadWork keeps heartbeating the dead registration forever.

### Findings
- [high] PTY exit leaves the AgentChattr registration alive indefinitely
  - File: `server/index.js:436`
  - `term.onExit` only marks the session stopped and closes the websocket. It never calls `stopHeartbeat` or `deregisterAgent`, even though that cleanup only exists in `stopAgentSession()` at `server/index.js:461-496`. Because the new heartbeat lives in the parent server process, an agent that crashes or exits via `exit` will still POST `/api/heartbeat/{name}` every 5s and remain `active` in AgentChattr. A later plain `/api/agents/:project/:agent/start` at `server/index.js:742-752` also overwrites the map entry, leaking the old interval handle permanently.
  - Suggestion: run the same AgentChattr cleanup on PTY exit that you run in `stopAgentSession()` (or extract that cleanup into a shared helper), so unexpected exits clear the heartbeat, deregister the slot, and remove the persisted token before another `/start`.

### Decision
Requesting changes because this breaks the crash semantics the heartbeat is supposed to preserve: dead agents can now stay falsely active forever.

project7-interns (changes requested):
Heartbeat is only cleared in `stopAgentSession`, not in `term.onExit` (server/index.js ~line 436). If the PTY exits on its own (crash, /exit, OOM) `stopAgentSession` is never called, so:

1. `acHeartbeatHandle` leaks an interval per crashed agent.
2. The heartbeat keeps POSTing for a dead agent, so AC never marks it crashed — defeating the 60s detection window this PR exists to preserve.

Fix: in the `term.onExit` callback, call `stopHeartbeat(current.acHeartbeatHandle)` and null the handle, mirroring `stopAgentSession`. Will re-approve once pushed.

project7-interns (approved):
Fix verified on commit a1e5ca4. `term.onExit` now stops the heartbeat, best-effort deregisters, and clears the persisted token + acRegistrationName/Token — symmetric with stopAgentSession. PTY-exit cleanup gap closed. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The PTY-exit cleanup gap is fixed. The updated `term.onExit` path now tears down the heartbeat and frees the AgentChattr registration instead of leaving a dead agent falsely active.

### Findings
- None. The prior blocker in `server/index.js` has been addressed in commit `a1e5ca4`.

### Decision
Approving because the heartbeat lifecycle is now symmetric enough for this sub-ticket: agent start begins the per-agent heartbeat, and both explicit stop/restart and unexpected PTY exit clear it and clean up the AgentChattr slot.

PR #256, opened 2026-04-08
[#252] sub-C: Auto-spawn AgentChattr after web wizard add-config

### LINKED ISSUE #252: [#249 sub-C] Auto-spawn AgentChattr after web wizard project creation
Parent: #249

## Problem

The CLI wizard's \`writeAgentChattrConfig\` (in \`bin/quadwork.js\`) automatically spawns the per-project AgentChattr process after writing the config. The web wizard's \`POST /api/setup\` \`add-config\` action does not — it creates the install dir, writes \`config.toml\`, but never starts the process.

Result: user goes through web wizard → opens project page → AgentChattr is silently not running. The chat connects to whatever happens to be on port 8300 (could be a stale instance from another project).

## Reference

\`bin/quadwork.js\` \`writeAgentChattrConfig\` — see the spawn block around line 940 for the canonical pattern. It uses \`installAgentChattr\` to ensure the clone, writes the config, then spawns:

\`\`\`javascript
const acSpawn = chattrSpawnArgs(acDir, [\"--config\", configTomlPath]);
const acProc = spawn(acSpawn.command, acSpawn.spawnArgs, {
  cwd: acSpawn.cwd,
  stdio: \"ignore\",
  detached: true,
});
acProc.unref();
\`\`\`

## Scope

After the web wizard's \`add-config\` step succeeds, automatically spawn the project's AgentChattr process. The cleanest approach is to call the existing \`/api/agentchattr/{project}/restart\` endpoint internally (it already does the right thing — kills any existing instance, then spawns a fresh one with the right config).

## Files to change

- \`server/routes.js\`:
  - In the \`add-config\` action handler, after the config is saved and the AgentChattr install is verified, trigger a restart for that project
  - Could either:
    - **Option A:** Make an internal HTTP call to \`/api/agentchattr/{id}/restart\` (loopback)
    - **Option B:** Refactor the restart logic in \`server/index.js\` into a callable function and import it
  - **Option B is cleaner** — avoid loopback HTTP

- \`server/index.js\`:
  - Export the spawn-chattr function so routes.js can call it directly
  - Keep the existing endpoint behavior unchanged

## Notes

- This should only run after a successful \`add-config\`. If install or config write failed, don't spawn.
- Be defensive: if AgentChattr port is already in use (collision), the spawn will fail. Log the error but don't crash the response.
- The web wizard's setup result should still return 200 even if spawn fails — user can retry from the dashboard's Restart button.

## Acceptance

- ✅ Creating a project via the web wizard auto-spawns its AgentChattr process
- ✅ The process is bound to the correct ports (verify with \`lsof\`)
- ✅ \`config.toml\` is loaded correctly (verify with \`curl http://127.0.0.1:{ac_port}/api/agents\`)
- ✅ Port collision failures are logged but don't crash the wizard
- ✅ User doesn't need to manually click \"Restart\" after creating a project

Branch: \`task/249c-auto-spawn-after-web\`

realproject7:
Fixes #252
Tracker: realproject7/agent-os#392
Master: #249

## Summary
- After `add-config` saves the project successfully, POST `/api/agentchattr/{id}/restart` over loopback to spawn the per-project AgentChattr process.
- Mirrors the existing pattern already used by the `agentchattr-config` branch (`server/routes.js:840`) and the CLI wizard's `writeAgentChattrConfig`.
- Failures are caught and logged via `console.warn` — the wizard still returns `ok: true` and the dashboard's Restart button remains a manual fallback, per the issue's "non-fatal" requirement.

## Why
Web wizard previously left the install dormant until the user clicked Restart, so the chat fell through to whatever stale AgentChattr instance happened to be on port 8300 (often from another project).

## Acceptance criteria
- [x] Creating a project via the web wizard auto-spawns its AgentChattr process
- [x] Spawn uses the per-project config (the /restart endpoint resolves config.toml + ports)
- [x] Spawn failures are logged but do not crash the wizard response
- [x] add-config still returns 200 even if spawn fails
- [ ] e2e verification — covered by sub-E (#254) operator-gated

## Test plan
- [x] node --check server/routes.js
- [ ] Reviewers: web-wizard a fresh project, confirm AC process spawned without manual Restart
- [ ] Reviewers: simulate port collision, confirm wizard still ok and warns in logs

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -935,9 +935,27 @@ router.post("/api/setup", (req, res) => {
         fetch(
           `http://127.0.0.1:${qwPort}/api/agentchattr/${encodeURIComponent(id)}/restart`,
           { method: "POST" },
-        ).catch((err) => {
-          console.warn(`[setup] auto-spawn AgentChattr for ${id} failed:`, err.message || err);
-        });
+        )
+          .then(async (r) => {
+            // /restart reports spawn failures (e.g. port collision —
+            // server/index.js:650-668) as HTTP 500, so a resolved
+            // fetch is not the same thing as a successful spawn. Log
+            // non-2xx responses with status and body so the operator
+            // can see why the auto-spawn silently didn't take.
+            if (!r.ok) {
+              let detail = "";
+              try { detail = (await r.text()).slice(0, 500); } catch {}
+              console.warn(
+                `[setup] auto-spawn AgentChattr for ${id} returned HTTP ${r.status}: ${detail}`,
+              );
+            }
+          })
+          .catch((err) => {
+            console.warn(
+              `[setup] auto-spawn AgentChattr for ${id} failed:`,
+              err.message || err,
+            );
+          });
       } catch (err) {
         console.warn(`[setup] auto-spawn AgentChattr for ${id} skipped:`, err.message || err);
       }
--- a/server/routes.js
+++ b/server/routes.js
@@ -922,6 +922,26 @@ router.post("/api/setup", (req, res) => {
       // ~/.quadwork/{id}/OVERNIGHT-QUEUE.md.
       writeOvernightQueueFileSafe(id, name || id, repo);
 
+      // Batch 28 / #392 / quadwork#252: auto-spawn the per-project
+      // AgentChattr process. The CLI wizard's writeAgentChattrConfig
+      // does this; the web wizard previously left the install dormant
+      // until the user clicked Restart, so MCP fell through to a stale
+      // instance on port 8300. Mirror the loopback-restart pattern
+      // already used by the agentchattr-config branch above. Failures
+      // are non-fatal — the dashboard's Restart button is still
+      // available, and per the issue add-config must still return ok.
+      try {
+        const qwPort = cfg.port || 8400;
+        fetch(
+          `http://127.0.0.1:${qwPort}/api/agentchattr/${encodeURIComponent(id)}/restart`,
+          { method: "POST" },
+        ).catch((err) => {
+          console.warn(`[setup] auto-spawn AgentChattr for ${id} failed:`, err.message || err);
+        });
+      } catch (err) {
+        console.warn(`[setup] auto-spawn AgentChattr for ${id} skipped:`, err.message || err);
+      }
+
       return res.json({ ok: true });
     }
     default:


project7-interns (approved):
LGTM. Mirrors the existing loopback restart pattern at server/routes.js:840 exactly. `cfg` is in scope from the add-config handler, fetch failures are caught and logged via console.warn, and the wizard still returns ok — matches the issue's non-fatal requirement. Dashboard Restart button remains as the manual fallback for the project-already-exists early-return branch. Node syntax check passes per the test plan.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The auto-spawn hook is placed in the right spot, but its error handling only covers network-level fetch failures. It does not log the ordinary restart failures that the issue explicitly calls out, such as a port collision returning HTTP 500 from the existing `/restart` endpoint.

### Findings
- [medium] Spawn failures from `/api/agentchattr/{id}/restart` are still silent
  - File: `server/routes.js:933`
  - `fetch(...).catch(...)` only runs when the HTTP request itself rejects. The restart endpoint returns a normal HTTP response with `500` and an error body when `spawnChattr()` fails (`server/index.js:650-668`), so this code treats the failure as success and logs nothing. That misses the issue requirement to log spawn failures while still returning `200` from the wizard.
  - Suggestion: inspect the fetch response and `console.warn` on `!res.ok` as well, ideally including the error text/body from `/restart`.

### Decision
Requesting changes because the PR does not yet meet the non-fatal failure visibility requirement for web-wizard auto-spawn.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The revised auto-spawn hook now handles both classes of failure the issue cares about: network-level fetch errors and non-2xx `/restart` responses from the existing AgentChattr restart endpoint.

### Findings
- None. The prior blocker in `server/routes.js` has been addressed in commit `01def39` by warning on `!r.ok` and including the restart response body.

### Decision
Approving because the web wizard still returns `ok: true`, while failed auto-spawn attempts are now operator-visible instead of silently disappearing.

project7-interns (approved):
Fix verified on commit 01def39. The .then now checks `!r.ok` and console.warns the HTTP status + truncated body (500 chars), so the port-collision 500 path from server/index.js:650-668 is no longer silent. Network-level rejections still hit the .catch. Wizard response stays non-fatal. LGTM.

PR #257, opened 2026-04-08
[#251] sub-B: Queue watcher injects 'mcp read' prompts into agent PTY

### LINKED ISSUE #251: [#249 sub-B] Queue watcher — inject 'mcp read' prompts into agent PTY when chat triggers
Parent: #249

## Problem

When the operator types \`@head\` in the chat, AgentChattr writes a job to \`{data_dir}/head_queue.jsonl\`. This is the **actual mechanism** by which agents receive chat messages — they don't poll the chat themselves; something has to read the queue file and poke them.

\`wrapper.py\` lines 438-541 handles this. It polls the queue file every ~1 second and, when content appears, **injects \`mcp read #general - you were mentioned, take appropriate action\` into the agent's terminal**. The agent's CLI sees the typed prompt and runs \`mcp read\` via its MCP connection, which reads the chat and lets the agent respond.

QuadWork has zero equivalent. **This is why agents never respond, even when registration and heartbeats work.**

## Reference

\`/Users/cho/Projects/agentchattr/wrapper.py\` lines 438-541 — the full \`_queue_watcher()\` function.

Key behaviors to port:
1. **Poll the queue file** every ~1s (configurable via \`refresh_interval\`)
2. **Read all lines, then truncate the file** (atomic claim)
3. **Build the prompt**:
   - Default: \`mcp read #{channel} - you were mentioned, take appropriate action\`
   - If a custom prompt is in the queue line, use it
   - If \`job_id\` is in the queue line, use \`mcp read job_id={job_id} ...\`
4. **Optionally append role and rules** (advanced — defer to a follow-up if scope creep)
5. **Inject the prompt into the agent's PTY** with \`session.term.write(prompt + \"\\r\")\`

## Why we have node-pty already

QuadWork already manages PTY sessions via \`spawnAgentPty\` in \`server/index.js\`. The session object has \`session.term\` (the node-pty instance). We can call \`session.term.write(...)\` to inject text. This is exactly what we need.

## Scope

Add a per-agent queue watcher. When an agent registers, start a watcher loop that polls \`{data_dir}/{registered_name}_queue.jsonl\` every 1 second. When the file has content, read it, truncate it, build the prompt, and write to the PTY.

## Files to change

- \`server/queue-watcher.js\` (new):
  - \`startQueueWatcher(dataDir, agentName, ptyTerm)\` — returns a watcher handle
  - \`stopQueueWatcher(handle)\` — stops the watcher
  - Internally uses \`setInterval\` (not fs.watch — wrapper.py uses polling, simpler and reliable across platforms)
- \`server/index.js\`:
  - Track watcher handle per agent session (\`session.queueWatcherHandle\`)
  - On agent register success: start the watcher with the agent's data_dir and PTY
  - On stop: stop the watcher
  - data_dir comes from the project's AgentChattr config.toml — already known via \`~/.quadwork/{project_id}/agentchattr/data\`

## Implementation sketch

\`\`\`javascript
// server/queue-watcher.js
const fs = require(\"fs\");
const path = require(\"path\");

function startQueueWatcher(dataDir, agentName, ptyTerm) {
  const queueFile = path.join(dataDir, \`\${agentName}_queue.jsonl\`);
  const intervalMs = 1000;

  const tick = () => {
    try {
      if (!fs.existsSync(queueFile)) return;
      const stat = fs.statSync(queueFile);
      if (stat.size === 0) return;

      const content = fs.readFileSync(queueFile, \"utf-8\");
      // Atomic claim: truncate immediately
      fs.writeFileSync(queueFile, \"\");

      const lines = content.split(\"\\n\").filter(l => l.trim());
      if (lines.length === 0) return;

      // Find channel + custom prompt + job_id
      let channel = \"general\";
      let customPrompt = \"\";
      let jobId = null;
      for (const line of lines) {
        try {
          const data = JSON.parse(line);
          if (data.channel) channel = data.channel;
          if (data.prompt) customPrompt = data.prompt;
          if (data.job_id) jobId = data.job_id;
        } catch { /* skip */ }
      }

      let prompt;
      if (customPrompt) {
        prompt = customPrompt;
      } else if (jobId) {
        prompt = \`mcp read job_id=\${jobId} - you were mentioned in a job thread, take appropriate action\`;
      } else {
        prompt = \`mcp read #\${channel} - you were mentioned, take appropriate action\`;
      }

      // Inject into PTY
      ptyTerm.write(prompt + \"\\r\");
    } catch { /* swallow */ }
  };

  return setInterval(tick, intervalMs);
}

function stopQueueWatcher(handle) {
  if (handle) clearInterval(handle);
}

module.exports = { startQueueWatcher, stopQueueWatcher };
\`\`\`

## Edge cases

- **Race between read and truncate**: small risk that AC writes a new line between the read and the truncate. wrapper.py has the same race and accepts it (lines may be lost in rare cases). Acceptable for v1.
- **Missing data_dir**: if AC isn't running yet, the file may not exist. \`!fs.existsSync\` is the guard.
- **Stale writes from previous agent instance**: when an agent restarts, the queue file may have content from before. Truncate on first poll to clear.

## Acceptance

- ✅ When operator types \`@head ping\` in chat, Head's queue file gets content
- ✅ The watcher reads the file, truncates it, injects \`mcp read #general...\` into Head's PTY
- ✅ Head's CLI shows the injected prompt and runs \`mcp read\` via its MCP connection
- ✅ Head responds in chat within ~30 seconds
- ✅ Stopping Head's session stops its watcher

## Out of scope (defer to follow-up)

- Role and rules fetching (lines 501-524 in wrapper.py)
- Re-register on watcher death recovery (sub-D handles 409)
- Identity hint for multi-instance setups
- Activity monitor (separate concern from queue watcher)

Branch: \`task/249b-queue-watcher\`

realproject7:
Fixes #251
Tracker: realproject7/agent-os#393
Master: #249

## Summary
- New `server/queue-watcher.js` with `startQueueWatcher(dataDir, agentName, ptyTerm)` / `stopQueueWatcher(handle)`. Polls `{data_dir}/{name}_queue.jsonl` every 1s, truncates as the atomic claim, parses JSON lines for `channel` / `job_id` / `prompt`, builds the appropriate `mcp read ...` prompt, flattens newlines (paste-detection guard from wrapper.py:532), and writes it into the PTY.
- Wired into `spawnAgentPty` (start), `term.onExit` (cleanup on crash), and `stopAgentSession` (cleanup on stop) in `server/index.js`. Resolves the AC data dir via `resolveProjectChattr(project).dir + "/data"`.

## Why
This is the missing mechanism behind "agents don't respond to chat." AgentChattr only writes to `{name}_queue.jsonl` and expects the agent side to poll and inject `mcp read`. Without it, every chat mention silently piles up in a file no one reads.

## Out of scope (per issue)
- Role / rules / identity-hint injections (wrapper.py lines 501-528) — deferred follow-up
- 409 recovery — sub-D #253

## Acceptance criteria
- [x] Watcher reads queue files and injects prompts into the correct agent PTY
- [x] Queue files truncated after read so jobs are claimed once
- [x] Stopping or crashing an agent stops its queue watcher (covered in both `stopAgentSession` and `term.onExit`)
- [ ] `@head ...` in chat → injected prompt → response — covered by sub-E (#254) operator-gated e2e

## Test plan
- [x] `node --check` on both files
- [ ] Reviewers: run a project, `@head ping` in chat, watch the head terminal show the injected `mcp read #general...` line
- [ ] Reviewers: stop the head agent, confirm no further file polls in `lsof`

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -9,6 +9,7 @@ const { spawn } = require("child_process");
 const { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr, resolveChattrSpawn, syncChattrToken, CONFIG_PATH } = require("./config");
 const routes = require("./routes");
 const { waitForAgentChattrReady, registerAgent, deregisterAgent, startHeartbeat, stopHeartbeat } = require("./agentchattr-registry");
+const { startQueueWatcher, stopQueueWatcher } = require("./queue-watcher");
 
 const net = require("net");
 const config = readConfig();
@@ -418,6 +419,7 @@ async function spawnAgentPty(project, agent) {
       acServerPort: built.acServerPort,
       acRegistrationToken: built.acRegistrationToken,
       acHeartbeatHandle: null,
+      queueWatcherHandle: null,
     };
     agentSessions.set(key, session);
 
@@ -431,6 +433,27 @@ async function spawnAgentPty(project, agent) {
         session.acRegistrationName,
         session.acRegistrationToken,
       );
+    }
+
+    // #393 / quadwork#251: queue watcher — the actual mechanism by
+    // which agents pick up chat. Without this an agent can be
+    // registered + heartbeating yet still never respond, because
+    // AgentChattr only writes to {data_dir}/{name}_queue.jsonl and
+    // expects the agent side to poll + inject `mcp read`.
+    if (session.acRegistrationName && session.term) {
+      try {
+        const { dir: acDir } = resolveProjectChattr(project);
+        if (acDir) {
+          const dataDir = path.join(acDir, "data");
+          session.queueWatcherHandle = startQueueWatcher(
+            dataDir,
+            session.acRegistrationName,
+            session.term,
+          );
+        }
+      } catch {
+        // best-effort — failure here just means no chat injection
+      }
     }
 
     term.onExit(({ exitCode }) => {
@@ -452,6 +475,10 @@ async function spawnAgentPty(project, agent) {
         if (current.acHeartbeatHandle) {
           stopHeartbeat(current.acHeartbeatHandle);
           current.acHeartbeatHandle = null;
+        }
+        if (current.queueWatcherHandle) {
+          stopQueueWatcher(current.queueWatcherHandle);
+          current.queueWatcherHandle = null;
         }
         if (current.acRegistrationName && current.acServerPort) {
           deregisterAgent(current.acServerPort, current.acRegistrationName).catch(() => {});
@@ -496,6 +523,12 @@ async function stopAgentSession(key) {
   if (session.acHeartbeatHandle) {
     stopHeartbeat(session.acHeartbeatHandle);
     session.acHeartbeatHandle = null;
+  }
+  // Stop queue watcher (#393 / quadwork#251) — the PTY is gone,
+  // injecting into a dead term would throw on the next tick.
+  if (session.queueWatcherHandle) {
+    stopQueueWatcher(session.queueWatcherHandle);
+    session.queueWatcherHandle = null;
   }
   // Best-effort deregister from AgentChattr (#241) so the slot frees
   // and the next register lands at slot 1 instead of head-2.
--- a/server/queue-watcher.js
+++ b/server/queue-watcher.js
@@ -63,7 +63,14 @@ function startQueueWatcher(dataDir, agentName, ptyTerm) {
         hasTrigger = true;
         if (data && typeof data === "object") {
           if (typeof data.channel === "string") channel = data.channel;
-          if (typeof data.job_id === "string") jobId = data.job_id;
+          // AgentChattr serializes job_id as an integer (agents.py
+          // defines `job_id: int | None`), so accept both numbers and
+          // strings here. Without this, job-thread triggers fall back
+          // to the channel prompt and the agent reads the wrong
+          // conversation. Cast to string for the prompt template.
+          if (typeof data.job_id === "number" || typeof data.job_id === "string") {
+            jobId = String(data.job_id);
+          }
           if (typeof data.prompt === "string" && data.prompt.trim()) {
             customPrompt = data.prompt.trim();
           }
--- a/server/queue-watcher.js
+++ b/server/queue-watcher.js
@@ -1,0 +1,107 @@
+/**
+ * Per-agent queue watcher (#393 / quadwork#251).
+ *
+ * AgentChattr does NOT push chat to agents. When the operator types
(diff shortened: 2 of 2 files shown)


project7-interns (approved):
LGTM. Faithful port of wrapper.py:438-541 — 1s poll, truncate-as-claim, JSON parse with bad-line skip, paste-detection newline flatten, `\r` submit. Lifecycle is correct: start in spawnAgentPty after registration, stop in both `stopAgentSession` and `term.onExit` so a PTY crash doesn't leave the interval poking a dead term.

The known limitations are all already called out in code/PR text and acceptable for v1: truncate race (matches wrapper.py), multi-job-per-tick coalescing (the agent's `mcp read` fetches everything since last read), and out-of-scope role/identity-hint injection (deferred per issue).

Node syntax check passes per the test plan.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The queue watcher wiring is close, but its `job_id` parsing is narrower than the upstream queue format. That breaks job-thread mentions, which are one of the supported trigger types in AgentChattr.

### Findings
- [high] Numeric `job_id` queue entries are ignored, so job-thread triggers fall back to channel reads
  - File: `server/queue-watcher.js:66`
  - The watcher only accepts `job_id` when `typeof data.job_id === "string"`. AgentChattr writes `job_id` as an integer (`agents.py`, `job_id: int | None`, and it serializes that numeric value directly into `{name}_queue.jsonl`). As a result, queue entries from job conversations never take the `mcp read job_id=...` path here; they incorrectly fall back to `mcp read #channel ...`, which loses the job-thread context and sends the agent to the wrong conversation.
  - Suggestion: mirror `wrapper.py` and accept any present `job_id` value, or at least both numbers and strings, before building the injected prompt.

### Decision
Requesting changes because this regresses one of the queue watcher’s core supported trigger modes: mentions inside job threads.

project7-interns (approved):
Fix verified on commit 3a5ce90. `job_id` now accepts both number and string and coerces with `String()` for the prompt template, so int job_ids from AgentChattr's agents.py route correctly to `mcp read job_id=...`. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The queue watcher fix now matches AgentChattr's queue schema: numeric and string `job_id` values both route to the job-thread read path, so job mentions no longer fall back to a channel read.

### Findings
- None. The prior blocker in `server/queue-watcher.js` has been addressed in commit `3a5ce90` by accepting both numeric and string `job_id` values and coercing to string for the prompt.

### Decision
Approving because the watcher behavior now aligns with the upstream queue writer and preserves the intended `mcp read job_id=...` flow for job-thread triggers.

PR #258, opened 2026-04-08
[#253] sub-D: Re-register agents on heartbeat 409 (AC restart recovery)

### LINKED ISSUE #253: [#249 sub-D] Re-register agents on heartbeat 409 (AgentChattr restart recovery)

realproject7:
Fixes #253
Tracker: realproject7/agent-os#394
Master: #249

## Summary
- `startHeartbeat` now reads name/token via getters every tick and accepts an `onConflict` callback. Reentry guarded by a `recovering` flag so a slow re-register can't stack.
- New `recoverFrom409` helper in `server/index.js`: waits for AC, deregisters the stale slot, calls `registerAgent`, updates `session.acRegistrationName` / `acRegistrationToken` in place (next heartbeat tick uses them), refreshes Claude MCP config file or Codex MCP proxy depending on inject mode, and restarts the queue watcher if the assigned name changed.
- `buildAgentArgs` now also returns `acInjectMode` + `acMcpHttpPort` so `recoverFrom409` knows which transport to refresh without re-reading config inline.
- Mirrors `agent-os/agentchattr/wrapper.py:732-741`.

## Acceptance criteria
- [x] 409 from heartbeat triggers re-registration
- [x] Session state, MCP config, and Codex proxy auth refreshed to the replacement token
- [x] Queue watcher restarted on name change so chat still reaches the agent
- [x] Recovery failures non-fatal (try/catch around each step, no rethrow)
- [x] Re-entry guard prevents tight retry loops while AC is still down
- [ ] e2e: dashboard Restart of AC → agents auto-recover within ~10s — covered by sub-E (#254) operator-gated

## Test plan
- [x] `node --check` on both files
- [ ] Reviewers: register a project, start agents, click Restart on the AC dashboard, confirm agents reappear in chat without manual intervention
- [ ] Reviewers: confirm the persisted token file is updated and the new MCP config has the new bearer

### DIFF
--- a/server/agentchattr-registry.js
+++ b/server/agentchattr-registry.js
@@ -105,19 +105,45 @@ async function deregisterAgent(serverPort, name, token) {
  * QuadWork agent silently disappears from the channel one minute after
  * registration.
  *
- * Returns an opaque handle suitable for stopHeartbeat. Transient errors
- * (network blips, AgentChattr restart) are swallowed — the next tick
- * just tries again. The 409 "identity wiped" recovery flow is sub-D
- * (quadwork#253) and is intentionally not handled here.
+ * `getName` and `getToken` are read on every tick (either as plain
+ * values or as zero-arg functions) so the sub-D 409 recovery path can
+ * swap them in place after re-registration without restarting the
+ * interval. Pass an `onConflict` callback to handle 409 responses
+ * (AgentChattr restart wiped the in-memory registry — re-register and
+ * the next tick will use the new credentials). Other errors are
+ * swallowed.
  *
- * Reference: /Users/cho/Projects/agentchattr/wrapper.py lines 715-748.
+ * Returns an opaque handle suitable for stopHeartbeat.
+ *
+ * Reference: /Users/cho/Projects/agentchattr/wrapper.py lines 715-744.
  */
-function startHeartbeat(serverPort, name, token, intervalMs = 5000) {
-  const url = `http://127.0.0.1:${serverPort}/api/heartbeat/${encodeURIComponent(name)}`;
-  const headers = { Authorization: `Bearer ${token}` };
+function startHeartbeat(serverPort, getName, getToken, { onConflict, intervalMs = 5000 } = {}) {
+  // Sub-D guard: avoid re-entering onConflict while a previous recovery
+  // attempt is still in flight. Without this, a slow re-register would
+  // be triggered once per 5s tick and stack up duplicate registrations.
+  let recovering = false;
+  const resolve = (v) => (typeof v === "function" ? v() : v);
   const tick = async () => {
+    const name = resolve(getName);
+    const token = resolve(getToken);
+    if (!name) return;
+    const url = `http://127.0.0.1:${serverPort}/api/heartbeat/${encodeURIComponent(name)}`;
     try {
-      await fetchWithTimeout(url, { method: "POST", headers }, DEFAULT_TIMEOUT_MS);
+      const r = await fetchWithTimeout(
+        url,
+        { method: "POST", headers: token ? { Authorization: `Bearer ${token}` } : {} },
+        DEFAULT_TIMEOUT_MS,
+      );
+      if (r.status === 409 && typeof onConflict === "function" && !recovering) {
+        recovering = true;
+        try {
+          await onConflict();
+        } catch {
+          // recovery is best-effort; next tick retries on its own
+        } finally {
+          recovering = false;
+        }
+      }
     } catch {
       // swallow — next tick will retry
     }
--- a/server/index.js
+++ b/server/index.js
@@ -147,14 +147,20 @@ function startMcpProxy(projectId, agentId, upstreamUrl, token) {
   const existing = mcpProxies.get(key);
   if (existing) return Promise.resolve(`http://127.0.0.1:${existing.port}/mcp`);
 
+  // #394 / quadwork#253: token is mutable so the 409 recovery path can
+  // swap it via updateMcpProxyToken without rebinding the listener —
+  // Codex was launched with a fixed proxy URL on an ephemeral port and
+  // can't be told to use a new one mid-flight.
+  const tokenRef = { current: token };
   return new Promise((resolve, reject) => {
     const proxyServer = http.createServer((req, res) => {
       const parsedUrl = new URL(req.url, `http://127.0.0.1`);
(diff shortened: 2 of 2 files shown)


project7-interns (approved):
LGTM. Faithful port of wrapper.py:732-741.

- `startHeartbeat` getter pattern + `recovering` re-entry guard cleanly avoids the tight retry loop / duplicate-recovery footgun. Token+name swap-in-place means the next 5s tick picks up replacement credentials without tearing down the interval.
- `recoverFrom409` order is correct: waitForReady → stale-cleanup → register → swap session state → refresh MCP transport → restart queue watcher only on name change. Each step wrapped in try/catch so a partial failure on one transport doesn't break the others. Codex proxy path (`stopMcpProxy` + `startMcpProxy`) genuinely re-auths since the proxy is in-process.
- `buildAgentArgs` returning `acInjectMode` + `acMcpHttpPort` keeps `recoverFrom409` from re-reading config and matches the registration code paths.
- Node syntax check passes per test plan.

Non-blocking observation for the operator-gated sub-E e2e: the Claude (`flag` inject mode) path rewrites the `mcp.json` file but a running Claude CLI process won't re-read it without an explicit reload, so Claude agents may need a one-time `/mcp reconnect` (or similar) after AC restart even with this fix. Codex/proxy_flag is fully self-healing. Worth confirming in #254.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The 409 recovery flow updates Claude config correctly, but the Codex proxy path cannot work as implemented: it replaces the local proxy on a new random port without updating the running Codex process that is still pinned to the old proxy URL.

### Findings
- [high] Codex agents lose MCP after 409 recovery because the replacement proxy comes up on a different URL
  - File: `server/index.js:435`
  - Codex agents are launched with a fixed proxy URL from `startMcpProxy()` (`server/index.js:345-348`). That helper binds with `listen(0)` and returns an ephemeral port (`server/index.js:182-185`). In recovery, the PR does `stopMcpProxy()` and `startMcpProxy()` again (`server/index.js:437-439`), which means the new proxy is usually on a different port, but the already-running Codex CLI is still configured to talk to the old one. The issue explicitly relies on replacing the proxy behind the existing URL; this implementation does not preserve that URL.
  - Suggestion: keep the existing proxy listener alive and swap the bearer token in place, or otherwise guarantee the recovered proxy comes back on the exact same local URL the running Codex process already has.

### Decision
Requesting changes because one of the stated acceptance criteria is still broken: Codex agents will not actually recover MCP connectivity after AgentChattr restart.

project7-interns (approved):
Fix verified on commit e4f3e1c. `startMcpProxy` now captures bearer through a mutable `tokenRef` in the request-handler closure, `updateMcpProxyToken` swaps `tokenRef.current` without touching the listener, and `recoverFrom409`'s Codex branch calls it instead of stop+start. The TCP listener and ephemeral port survive, so Codex's pinned proxy URL keeps working and the very next request from the running Codex CLI carries the new bearer. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The Codex recovery path now updates proxy credentials in place instead of rebinding the listener. That preserves the original local proxy URL the running Codex process is already pinned to, so MCP recovery can succeed after AgentChattr restart.

### Findings
- None. The prior blocker in `server/index.js` has been addressed in commit `e4f3e1c` by keeping a mutable `tokenRef` in the proxy closure and updating it through `updateMcpProxyToken()`.

### Decision
Approving because the 409 recovery flow now refreshes the relevant live transport for both supported paths in scope: Claude config-file mode and Codex proxy mode.

Issue #261, opened 2026-04-08
[master] Chat panel reply button + Agent Terminals layout polish


realproject7:
## Master — Small UX polish batch

Four small but visible improvements to the AgentChattr chat panel and Agent Terminals grid.

## Sub-tickets

- [ ] **#262** sub-A — Add grey reply button on each chat message
- [ ] **#263** sub-B — Display reviewer names as RE1 / RE2 in chat sender column
- [ ] **#264** sub-C — Restore activity ring on agent status dots + fix question-icon vertical alignment
- [ ] **#265** sub-D — Reorder Agent Terminals: Head TL, Dev TR, RE1 BL, RE2 BR

## Acceptance

- ✅ Each chat message has a small grey reply button → clicking it sends a reply tagged at that message
- ✅ Reviewer1 / Reviewer2 displayed as `RE1` / `RE2` in the chat sender column (other UI unchanged)
- ✅ Active agents show a pulsing ring around their status dot in the Agent Terminals header
- ✅ Question mark icon next to "Agent Terminals" is vertically centered with the title
- ✅ Agent Terminals 2x2 grid shows: Head TL, Dev TR, Reviewer1 BL, Reviewer2 BR

realproject7:
All 4 sub-tickets merged: #268 (sub-A reply button), #272 (sub-B RE1/RE2), #275 (sub-C status ring + icon), #280 (sub-D grid reorder). Closing master.

PR #268, opened 2026-04-08
[#262] sub-A: Chat reply button on each message

### LINKED ISSUE #262: [#261 sub-A] Chat reply button on each message
Parent: #261

## Scope

Add a small grey reply button on each chat message in the AgentChattr panel. Clicking it should send a reply that references the original message.

## Reference

AgentChattr's native UI (visible at \`http://127.0.0.1:{port}\`) has this feature already. Each message has a reply affordance that pre-fills the input with a quoted/threaded reply. Mirror that behavior in QuadWork's ChatPanel.

## Behavior

- Hover (or always-visible, lightweight) reply button appears next to each message
- On click:
  - Set the input to focus
  - Pre-fill the input with \`@<original sender> \` (so the reply is routed to that agent)
  - Optionally include a \`reply_to\` reference if the AgentChattr API supports it (check the existing chat send call shape — agents already use \`reply_to\` in the response messages we've seen, e.g., \`chat_send({\"channel\":\"general\",\"sender\":\"head\",\"reply_to\":1,...})\`)
- The reply button is small, grey, unobtrusive — mirror the styling of the existing send button

## Files

- \`src/components/ChatPanel.tsx\` — message rendering loop, add reply button JSX + handler
- Possibly extend the message data shape to include \`id\` (if not already present) so \`reply_to\` can be passed
- Backend: verify \`/api/chat\` POST accepts a \`reply_to\` field; if not, pass it through the existing WebSocket-based chat send

## Acceptance

- ✅ Each message in the chat list has a small grey reply button
- ✅ Click → input focuses with \`@<sender>\` prefix
- ✅ When the reply is sent, it includes \`reply_to: <original_message_id>\` in the AgentChattr payload
- ✅ Native AgentChattr UI shows the reply linked to the original (verify in browser tab)

Branch: \`task/261a-chat-reply-button\`

realproject7:
Fixes #262
Tracker: realproject7/agent-os#397
Master: #261

## Summary
- New small grey hover-revealed reply button on each chat row in `ChatPanelAPI`.
- Click → selects the message as the active reply, focuses the input, prefills `@<sender> `.
- Active reply shows a "Replying to @<sender> #<id>" indicator above the input with a ✕ cancel (Esc also clears).
- `send()` includes `reply_to: <msg.id>` in the `/api/chat` POST body. After a successful send, replyTo is cleared.
- `server/routes.js` POST `/api/chat` accepts `reply_to`, validates as a positive integer, and spreads it into the message forwarded over the AC ws — `sendViaWebSocket` already passes `...message` through, so AC's native UI sees the threaded reply.

## Acceptance criteria
- [x] Each chat message has a small grey reply button (hover-revealed)
- [x] Click → input focuses with `@<sender>` prefix
- [x] Sent reply includes `reply_to` with the original message id
- [ ] Reviewers: open native AC UI in a tab and confirm the reply is linked to the original
- [x] Reply state cleared on send + on cancel + on Esc

## Test plan
- [x] `tsc --noEmit` clean
- [ ] Reviewers: spin up a project, send a message, hover a row, click reply, type a response, confirm AC native UI shows the threaded link

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -185,11 +185,20 @@ router.post("/api/chat", async (req, res) => {
   // hitting QuadWork's /api/chat impersonate an agent identity (t1,
   // t3, …) over the AgentChattr ws path, which the old /api/send
   // flow could not do.
+  // #397 / quadwork#262: pass reply_to through to AgentChattr so the
+  // dashboard's reply button mirrors AC's native threaded-reply
+  // behavior. Only forward when it's a real positive integer — guards
+  // against arbitrary client payloads.
+  const replyToRaw = req.body?.reply_to;
+  const replyTo = (typeof replyToRaw === "number" && Number.isInteger(replyToRaw) && replyToRaw > 0)
+    ? replyToRaw
+    : null;
   const message = {
     text: typeof req.body?.text === "string" ? req.body.text : "",
     channel: req.body?.channel || "general",
     sender: "user",
     attachments: Array.isArray(req.body?.attachments) ? req.body.attachments : [],
+    ...(replyTo !== null ? { reply_to: replyTo } : {}),
   };
   if (!message.text && message.attachments.length === 0) {
     return res.status(400).json({ error: "text or attachments required" });
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -115,6 +115,9 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
   const [showMentions, setShowMentions] = useState(false);
   const [mentionFilter, setMentionFilter] = useState("");
   const [authError, setAuthError] = useState<string | null>(null);
+  // #397 / quadwork#262: tracks the message the next send will be a
+  // threaded reply to. Cleared after a successful send or on cancel.
+  const [replyTo, setReplyTo] = useState<{ id: number; sender: string } | null>(null);
   const cursorRef = useRef(0);
   const listRef = useRef<HTMLDivElement>(null);
   const inputRef = useRef<HTMLInputElement>(null);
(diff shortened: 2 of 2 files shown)


project7-interns (approved):
LGTM. Small, focused UX change. Server side validates `reply_to` as a positive integer before spreading it into the AC ws message, so arbitrary client payloads can't leak through. Client side hover-reveal with group-hover is consistent with AC's native reply UI, reply state is cleared on send/cancel/Esc, and the input prefill preserves any in-progress typing via the `startsWith` guard. tsc --noEmit and node --check both clean per the test plan.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The threaded-reply plumbing is mostly in place, but the front-end reply targeting does not actually mirror native AgentChattr behavior for non-agent rows. In QuadWork, replying to `user`/`system` messages silently reroutes the send to Head.

### Findings
- [medium] Replying to non-agent rows injects an `@sender` mention that gets silently rewritten to Head
  - File: `src/components/ChatPanel.tsx:283`
  - The new button is rendered for every message row and always does `const prefix = `@${msg.sender} ``. Later, `send()` only treats `@head|@dev|@reviewer1|@reviewer2` as a recognized target and otherwise prepends `@head ` (`src/components/ChatPanel.tsx:192-203`). So replying to common non-agent rows like `user` or `system` becomes `@head @user ...` or `@head @system ...`, which is not a reply to that sender at all. Native AgentChattr avoids this: `startReply()` only auto-targets senders that resolve to real agents, and system rows do not get a reply action in the first place (`agentchattr/static/chat.js:739-742`, `agentchattr/static/chat.js:2488-2502`).
  - Suggestion: only prefill/target when the replied-to sender is a real agent that this panel can mention, and suppress the reply affordance on system rows; otherwise keep only the `reply_to` thread link without injecting an `@sender` prefix.

### Decision
Requesting changes because the current UI can misroute replies to the wrong recipient even though the threaded `reply_to` metadata is correct.

project7-interns (approved):
Fix verified on commit 919d083. Gated `@sender` prefill behind a KNOWN_AGENTS set ({head, dev, reviewer1, reviewer2}). Non-agent rows still set replyTo (so AC's native UI threads via reply_to) but leave the input untouched, so send()'s default @head routing behaves exactly as before for user/system replies. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The reply-targeting fix now matches the intended behavior closely enough for this sub-ticket: agent rows still prefill a direct `@sender` target, while non-agent rows preserve the threaded `reply_to` link without injecting a bogus mention that gets rerouted.

### Findings
- None. The prior blocker in `src/components/ChatPanel.tsx` has been addressed in commit `919d083` by gating the `@<sender>` prefill to known agent senders.

### Decision
Approving because the dashboard reply button now avoids the misrouting regression on non-agent rows while still forwarding valid `reply_to` metadata to AgentChattr.

PR #218, opened 2026-04-07
[#208] 4-quadrant project dashboard layout

realproject7:
## Summary
Phase 1F of Batch 25 / master #202. Restructure the project dashboard into the new 4-quadrant layout that promotes AgentChattr as the primary operator interface.

Four files, +127/−37.

### New layout
- **Top-left** — AgentChattr chat, wrapped in a 2-px accent border and labelled `AgentChattr — primary chat` in the panel header.
- **Top-right** — Agent Terminals 2x2 grid in a new wrapper with an `Agent Terminals` header and a `?` tooltip that warns operators not to type into terminals directly (messages typed there are invisible to other agents — use the chat above instead).
- **Bottom-left** — GitHub panel (unchanged content, moved from top-right).
- **Bottom-right** — new `OperatorFeaturesPanel` placeholder listing the three features that land in #209/#210/#211.

### `src/components/ProjectDashboard.tsx`
- Swap the four quadrant contents. Drag-to-resize row/col behavior + min sizes preserved.
- Drop the orphaned Head-only `TerminalPanel` — the 2x2 grid in the top-right quadrant already renders the Head terminal as one of its four tiles, so the old top-left Head panel was redundant.

### New components
- `src/components/AgentTerminalsGrid.tsx` — wraps the existing `<TerminalGrid />` with the header + tooltip. Tooltip shows on hover and keyboard focus, with `role="tooltip"` and an `aria-label` on the trigger.
- `src/components/OperatorFeaturesPanel.tsx` — placeholder with three bulleted items (queue viewer, scheduled trigger, telegram bridge) referencing their sub-tickets.

### `src/components/TerminalGrid.tsx`
- Status dots gain a pulsing accent ring (`animate-ping`) when the agent is `running`. Idle/stopped/error keep the plain dot.

`ChatPanel` + `ControlBar` untouched — the \"primary\" highlight is applied by the parent wrapper `div` in `ProjectDashboard` so `ChatPanel` stays reusable anywhere.

## Acceptance criteria from #208
- ✅ Project page uses 4-quadrant layout (TL=Chat, TR=Terminals, BL=GitHub, BR=Operator Features).
- ✅ AgentChattr is visually highlighted as primary (2-px accent border + header label).
- ✅ Agent Terminals header has the `?` tooltip with the exact ticket copy.
- ✅ Terminal status dots show an activity ring when the agent is busy.
[...]

PR #275, opened 2026-04-08
[#264] sub-C: Activity-gated status ring + center (?) icon

### LINKED ISSUE #264: [#261 sub-C] Restore activity ring on agent status dots + fix question-icon vertical alignment
Parent: #261

### Fix 1 — Activity ring on status dots (regression)

The original wireframe (and PR #218 spec) included a pulsing ring around each agent's status dot when the agent is active (busy processing). This was implemented in #218 using \`animate-ping\`. It is currently NOT visible — possibly removed or broken in a later refactor.

**Reference design:** the AgentChattr UI shows pulsing rings around the agent dots in its top header (e.g., the t1/t2a/t2b chips). Mirror that style.

**Verify and restore:**
- Inspect \`src/components/TerminalGrid.tsx\` (or wherever the status dot is rendered)
- Confirm the ring class is still there
- If missing, re-add: a small \`<span className=\"animate-ping ...\">\` overlay positioned over the dot, gated by the agent's \`busy\` / \`active\` state
- The ring should pulse only while the agent is processing — idle agents show a static dot

If the backend signal for \"agent is active\" doesn't exist, derive it from PTY output (last write timestamp within the last few seconds) or from a server-tracked \"recently active\" flag.

### Fix 2 — Question-icon vertical alignment

The \`(?)\` icon next to the \"Agent Terminals\" title sits slightly **below** the title's text baseline. Should be **vertically centered** with the title text.

**Fix:** Wrap title and icon in a flex container with \`items-center\`, or adjust the icon's \`inline-block\` / \`vertical-align\` / line-height. CSS-only change.

## Files

- \`src/components/AgentTerminalsGrid.tsx\` — header with title + question icon
- \`src/components/TerminalGrid.tsx\` — per-tile status dot (or wherever the dot is rendered)

## Acceptance

- ✅ Each agent's status dot has a pulsing ring when the agent is active
- ✅ Idle agents show a static dot (no ring)
- ✅ Question-mark icon is vertically centered with \"Agent Terminals\" title text
- ✅ Tooltip behavior on the question icon unchanged

Branch: \`task/261c-status-ring-icon-align\`

realproject7:
Fixes #264
Tracker: realproject7/agent-os#399
Master: #261

## Summary
Two paired Agent Terminals header fixes:

1. **Activity ring (regression from #218)** — `TerminalPanel` now exposes an `onActivity` prop that fires from `ws.onmessage`. `TerminalGrid` stores per-agent last-activity timestamps in a ref (hot path) and re-evaluates freshness via a 500ms tick. The animate-ping ring renders only when `state === "running"` AND output landed in the last 2s. Idle live agents show a static dot, busy agents pulse.
2. **Question icon alignment** — wrap the tooltip `<div className="relative">` in `inline-flex items-center` and add `leading-none` on the (?) button so the glyph centers on the title's cap-height instead of dropping to its own baseline.

## Acceptance criteria
- [x] Active agents show a pulsing ring around the status dot
- [x] Idle (running but no PTY output in window) agents show a static dot
- [x] Question-mark icon is vertically centered with the "Agent Terminals" title
- [x] Tooltip behavior unchanged (mouse/focus enter/leave, position class untouched)

## Test plan
- [x] `tsc --noEmit` clean
- [ ] Reviewers: spin up a project, watch ring pulse during agent processing, watch ring stop within ~2s of last PTY output, confirm (?) icon centers
- [ ] Reviewers: confirm tooltip still renders on hover/focus

### DIFF
--- a/src/components/AgentTerminalsGrid.tsx
+++ b/src/components/AgentTerminalsGrid.tsx
@@ -42,7 +42,11 @@ export default function AgentTerminalsGrid({ projectId, agentStates, onStatusCha
         <div className="flex items-center gap-1.5">
           <span className="text-[11px] text-text-muted uppercase tracking-wider">Agent Terminals</span>
           <div
-            className="relative"
+            // #399 / quadwork#264: inline-flex+items-center so the
+            // (?) button vertically centers with the title text. The
+            // previous block-level wrapper let the button drop to its
+            // own baseline below the title cap-height.
+            className="relative inline-flex items-center"
             onMouseEnter={() => setTipOpen(true)}
             onMouseLeave={() => setTipOpen(false)}
             onFocus={() => setTipOpen(true)}
@@ -51,7 +55,7 @@ export default function AgentTerminalsGrid({ projectId, agentStates, onStatusCha
             <button
               type="button"
               aria-label="About agent terminals"
-              className="w-3.5 h-3.5 rounded-full border border-border text-[9px] text-text-muted hover:text-accent hover:border-accent inline-flex items-center justify-center"
+              className="w-3.5 h-3.5 rounded-full border border-border text-[9px] leading-none text-text-muted hover:text-accent hover:border-accent inline-flex items-center justify-center"
             >?</button>
             {tipOpen && (
               <div
--- a/src/components/TerminalGrid.tsx
+++ b/src/components/TerminalGrid.tsx
@@ -1,7 +1,11 @@
 "use client";
 
-import { useState } from "react";
+import { useState, useRef, useEffect, useCallback } from "react";
 import TerminalPanel from "./TerminalPanel";
+
+// #399 / quadwork#264: how long an agent stays "active" after its
+// last PTY output before the activity ring stops pulsing.
+const ACTIVITY_WINDOW_MS = 2000;
 
 interface Agent {
   id: string;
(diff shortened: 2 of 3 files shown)


project7-interns (changes requested):
High-severity regression in `src/components/TerminalPanel.tsx:177`: adding `onActivity` to the WebSocket `useEffect` dep array tears down and reconnects the PTY WebSocket every time `onActivity` identity changes.

In `TerminalGrid.tsx:181` the `onActivity` prop is an inline arrow `() => markActivity(agent.id)` that is a fresh function on every render. The new `activityTick` state is bumped every 500ms, which re-renders `TerminalGrid`, which makes a new inline arrow, which invalidates `TerminalPanel`'s effect, which closes and reopens the ws. Net effect:

- Every agent terminal reconnects its PTY WebSocket every 500ms.
- Terminal state (scrollback, cursor) is lost on every tick.
- `ws.onmessage` never stays attached long enough to feed `markActivity`, so the ring this PR is supposed to add will in practice never light up.
- Backend is pounded with ~8 ws opens/sec (4 agents × 2 ticks).

Fix options:
1. In `TerminalPanel`, stash `onActivity` in a ref and read `onActivityRef.current()` inside `ws.onmessage`. Drop `onActivity` from the effect deps.
2. Or in `TerminalGrid`, memoize the per-agent callback with `useCallback` keyed on `agent.id` + stable `markActivity` so the arrow identity is stable across ticks.

Option 1 is safer — isolates the fix in one file and prevents future callers from repeating the mistake. Will re-review once pushed.

project7-interns (approved):
Fix verified on commit 81d681a. `onActivity` stashed in `onActivityRef`, synced via a tiny effect, read as `onActivityRef.current()` inside `ws.onmessage`. The ws effect now only re-runs on real identity changes (projectId / agentId / wsUrl / fit), so inline-arrow callers from TerminalGrid stay safe by construction and the activityTick 500ms re-render no longer tears down the PTY. Ring will actually light up now. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The change cleanly restores the intended visual behavior: the status ring now pulses only on recent PTY output, and the question-mark icon alignment fix is a purely presentational wrapper/class adjustment with tooltip behavior preserved.

### Findings
- None. The activity signal is derived from PTY output without reopening the terminal websocket on every freshness tick, and the icon-centering change is tightly scoped to the header markup/classes.

### Decision
Approving because this matches the issue’s acceptance criteria without touching unrelated terminal behavior or backend state.

Issue #283, opened 2026-04-08
[master] Batch 30 — Critical fixes + UX polish + new features (13 tickets)


realproject7:
## Master Ticket — Batch 30

A coordinated batch combining 2 critical bug fixes, 1 workflow fix, 1 medium feature for safety, 7 small UI polish tickets, and 2 medium features.

### Phase 1 — Critical bugs (do these first)
- [ ] **#277** sub-A — Scheduled Trigger send fix (uses broken /api/send → must use /api/chat)
- [ ] **#266** sub-B — Head should ask operator to kick off batch (silent after queue bug)

### Phase 2 — Workflow safety
- [ ] **#274** sub-C — Loop guard auto-set to 30 + UI widget + explainer
- [ ] **#267** sub-D — Auto-increment batch number on OVERNIGHT-QUEUE.md

### Phase 3 — UI polish (small, parallelizable)
- [ ] **#269** sub-E — Scheduled Trigger custom hours input
- [ ] **#270** sub-F — Keep Awake custom hours + explainer tooltip
- [ ] **#271** sub-G — Color-code @mentions in chat messages
- [ ] **#273** sub-H — Notification sound (with sound choice + background-only mode)
- [ ] **#276** sub-I — Slash commands autocomplete menu + skip auto-tag for /
- [ ] **#278** sub-J — Operator display name in chat (replace hardcoded 'user')
- [ ] **#281** sub-K — Show recently closed issues + merged PRs (5 each) in GitHub panel

### Phase 4 — Medium features (do last)
- [ ] **#279** sub-L — Project history export/import (chat backup + restore)
- [ ] **#282** sub-M — Current Batch Progress panel with visual progress bars

## Why this order

1. **Phase 1 first** because:
   - #277 is a real bug — Scheduled Trigger appears to work but never delivers messages, leaving agents idle and operators confused
   - #266 is a workflow blocker — Head silently waits after building the queue, operator has no idea what to do next
2. **Phase 2 next** because:
   - #274 prevents the dropcast batch's loop-guard hangs that just bit the operator (4 hops fires after 1 PR review cycle)
   - #267 makes batch state explicit so operators can refer to specific batches
3. **Phase 3 (UI polish)** is independent and parallelizable. Run sub-E through sub-K in any order
4. **Phase 4 (medium features)** depends on stable UI from Phase 3 — and #282 needs the GitHub panel restructure from #281

## Phase Checkpoints

[...]

PR #284, opened 2026-04-08
[#277] Scheduled Trigger uses local /api/chat instead of /api/send

### LINKED ISSUE #277: [bug] Scheduled Trigger messages never appear in chat — sendTriggerMessage uses broken /api/send endpoint
## Problem

The Scheduled Trigger widget shows the timer running (\"Sending: ...\", \"Next: 1m 43s\", \"Stops in: 2h 52m\") but the trigger messages **never appear in the AgentChattr chat**. Agents never receive the queue-check pulse, so they sit idle and the operator wonders why nothing is happening.

## Root cause

\`sendTriggerMessage()\` in \`server/index.js\` line 995 POSTs to AgentChattr's \`/api/send\` endpoint directly:

\`\`\`javascript
const res = await fetch(\`\${chattrUrl}/api/send\${tokenParam}\`, {
  method: \"POST\",
  headers,
  body: JSON.stringify({ text: message, channel: \"general\", sender: \"user\" }),
});
\`\`\`

But \`/api/send\` requires a **registration token**, not a session token (this is the same bug we fixed for the chat panel in PR #231 — chat 401 hotfix). The chat panel was migrated to use a WebSocket connection (\`sendViaWebSocket\` in \`server/routes.js\`) but \`sendTriggerMessage\` was missed in that refactor. It still uses the old broken endpoint.

The result: every trigger send returns 401 (or similar), the error gets logged silently to console (\`console.error(\\`Trigger send failed...\\`)\`), and the widget keeps running as if nothing was wrong.

## Fix

Refactor \`sendTriggerMessage\` to use the same WebSocket path as the chat panel. Two options:

### Option A — Reuse the chat endpoint (preferred)

Make \`sendTriggerMessage\` call the local \`/api/chat\` endpoint via internal HTTP loopback:

\`\`\`javascript
async function sendTriggerMessage(projectId) {
  // ... existing config lookup ...
  const message = (project && project.trigger_message) || DEFAULT_MESSAGE;
  const port = readConfig().port || 8400;
  try {
    const res = await fetch(\`http://127.0.0.1:\${port}/api/chat?project=\${encodeURIComponent(projectId)}\`, {
      method: \"POST\",
      headers: { \"Content-Type\": \"application/json\" },
      body: JSON.stringify({ text: message, channel: \"general\" }),
    });
    if (!res.ok) {
      const err = await res.text().catch(() => \"\");
      console.error(\`Trigger send failed for \${projectId}: \${res.status} \${err}\`);
      info.lastError = \`\${res.status}: \${err.slice(0, 100)}\`;
    } else {
      info.lastError = null;
    }
  } catch (err) {
    // ... existing error handling ...
  }
  // ... existing lastSent / nextAt update ...
}
\`\`\`

This guarantees the trigger uses the same proven path as the chat panel — including the token resync on 401 from #230.

### Option B — Extract sendViaWebSocket to a shared module

Move \`sendViaWebSocket\` from \`server/routes.js\` into \`server/chat-send.js\` (or similar) and import it into both places. Slightly more refactor but avoids the loopback HTTP call.

**Recommend Option A** — fewer moving parts, no shared module needed, and reusing the public API path means any future changes to the chat send flow automatically apply to the trigger too.

## Acceptance

- ✅ Starting a Scheduled Trigger sends the message and it appears in the AgentChattr chat
- ✅ The same message reappears every interval (e.g., every 15 minutes) in the chat
- ✅ Agents receive the trigger and respond
- ✅ Errors surface in the trigger widget's status (already exists via \`lastError\`)
- ✅ Token resync on 401 still works (inherited from \`/api/chat\`)

## Files

- \`server/index.js\` — \`sendTriggerMessage()\` line 983-1016

Branch: \`task/277-trigger-uses-chat-endpoint\`

realproject7:
Fixes #277
Tracker: realproject7/agent-os#401
Master: #283 (Batch 30 Phase 1)

## Summary
- `sendTriggerMessage()` (`server/index.js`) now POSTs to the local `/api/chat?project=<id>` loopback instead of AC's direct `/api/send`.
- Same proven path as the chat panel post-#231: opens the AC ws with the session token, inherits the #230 token-resync-on-401 retry. No shared module, no new plumbing.
- Removes the silent 401s that were leaving Scheduled Trigger ticking with no actual chat traffic.

## Acceptance criteria
- [x] No more direct `/api/send` calls from `sendTriggerMessage`
- [ ] Reviewers: enable Scheduled Trigger on a project, confirm message lands in chat each interval
- [ ] Reviewers: confirm trigger error display still surfaces non-2xx via `info.lastError`
- [x] Token resync on 401 inherited from `/api/chat` (no behavioural change to retry logic)

## Test plan
- [ ] Reviewers: spin up a project, set a 30s trigger interval, watch the chat for the first send

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -983,19 +983,24 @@ ALL: Communicate via this chat by tagging agents. Your terminal is NOT visible.`
 async function sendTriggerMessage(projectId) {
   const cfg = readConfig();
   const project = cfg.projects && cfg.projects.find((p) => p.id === projectId);
-  const { url: chattrUrl, token: chattrToken } = resolveProjectChattr(projectId);
-  const token = chattrToken || "";
   const message = (project && project.trigger_message) || DEFAULT_MESSAGE;
-  const headers = { "Content-Type": "application/json" };
-  if (token) headers["x-session-token"] = token;
+
+  // #401 / quadwork#277: route trigger sends through the local
+  // /api/chat path that already works for the chat panel. The old
+  // direct /api/send call required a registration token (not the
+  // session token we have on hand) and 401'd silently — agents never
+  // saw the queue-check pulse. /api/chat opens the AC ws with the
+  // session token and inherits the #230 token-resync-on-401 retry,
+  // so the trigger now gets the same proven path as the chat panel.
+  const qwPort = cfg.port || 8400;
+  const url = `http://127.0.0.1:${qwPort}/api/chat?project=${encodeURIComponent(projectId)}`;
 
   const info = triggers.get(projectId);
   try {
-    let tokenParam = token ? `?token=${encodeURIComponent(token)}` : "";
-    const res = await fetch(`${chattrUrl}/api/send${tokenParam}`, {
+    const res = await fetch(url, {
       method: "POST",
-      headers,
-      body: JSON.stringify({ text: message, channel: "general", sender: "user" }),
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ text: message, channel: "general" }),
     });
     if (!res.ok) {
       const err = await res.text().catch(() => "");


project7-interns (approved):
LGTM. Simple targeted fix routing trigger sends through the local `/api/chat` loopback that the chat panel already uses, inheriting #230 token-resync-on-401. Body drops `sender: "user"` because /api/chat hardcodes `sender: "user"` server-side (server/routes.js:189), so behavior is preserved. Single-file scope matches the issue.

project7-interns (approved):
## Verdict: APPROVE

### Summary
This PR matches issue #277 with a minimal, targeted fix: `sendTriggerMessage()` now routes through QuadWork's local `/api/chat` endpoint instead of the stale direct `/api/send` path.

### Findings
- None. The new loopback call preserves the existing trigger `lastError` handling in `server/index.js`, while `/api/chat` continues to stamp `sender: "user"`, send over the proven websocket path, and retry once after token resync on 401.

### Decision
Approving because the change directly addresses the root cause described in #277 without adding new plumbing or changing unrelated trigger behavior.

PR #286, opened 2026-04-08
[#274] Loop guard default 30 + operator widget

### LINKED ISSUE #274: [feat] Auto-set max_agent_hops to 30 + surface in operator panel with explainer
## Background

AgentChattr has a \"loop guard\" that pauses agent-to-agent message chains after N hops. The default is **4**, configured via \`[routing] max_agent_hops\` in \`config.toml\`. After N hops with no human in the loop, AgentChattr emits:

> Loop guard: 4 agent-to-agent hops reached. Type \`/continue\` to resume.

This is too low for QuadWork's autonomous workflow. A normal PR cycle (head → dev → reviewer1+2 → dev → head merge) easily hits 4-5 hops. Operators see the guard fire mid-cycle and have to type \`/continue\` repeatedly, defeating the \"runs while you sleep\" promise.

## What we know

- Config key: \`[routing] max_agent_hops\` in AgentChattr's \`config.toml\`
- AgentChattr default: **4**
- AgentChattr max allowed: **50** (validated in \`app.py\` line 1247)
- The hop counter resets whenever a human posts a message — so the guard only fires during fully-autonomous chains
- Per-instance: each AgentChattr clone has its own setting

## Scope (Option C)

1. **Set a sensible default of 30** when QuadWork creates a per-project AgentChattr clone — enough for ~5-6 full PR cycles without interruption
2. **Surface the value in the Operator Features panel** so users know it exists and can adjust it
3. **Add a (?) explainer tooltip** so the concept is approachable for newbies

### 1. Default in template

Update QuadWork's AgentChattr config template (the one written by \`writeAgentChattrConfig\` in \`bin/quadwork.js\` and \`server/routes.js\`) to include:

\`\`\`toml
[routing]
max_agent_hops = 30
\`\`\`

This ensures every new project gets the higher default automatically.

### 2. Migration for existing projects

For projects created before this change, add a one-time migration in \`cmdStart\` (similar to \`migrateLegacyProjects\`): if a project's \`config.toml\` is missing \`[routing] max_agent_hops\`, append it with value 30.

### 3. UI in Operator Features panel

Add a small \"Loop Guard\" section to the operator features panel (bottom-right quadrant), next to or below Scheduled Trigger:

\`\`\`
LOOP GUARD (?)
Pause after [30] hops between agents
[ Apply ]
\`\`\`

- Number input, default 30
- Min 4, max 50 (matches AgentChattr's allowed range)
- Apply button calls a new endpoint that updates \`max_agent_hops\` in the running AgentChattr's room settings (AgentChattr already has this — see \`app.py\` line 1244-1249, exposed via the room settings PATCH endpoint)
- Persists to disk so the next restart picks it up

### 4. Tooltip / explainer

The (?) icon opens a small popover:

> **Loop Guard** prevents agents from looping forever without human input. After this many agent-to-agent messages with no human reply, AgentChattr pauses the chain and asks you to confirm.
>
> Higher numbers let agents work longer overnight without interruption. Lower numbers add safety against runaway loops. The default for QuadWork is **30**, which is enough for ~5-6 full PR cycles.
>
> Posting any message yourself resets the counter immediately.

## Files

- \`bin/quadwork.js\` \`writeAgentChattrConfig()\` — add \`[routing] max_agent_hops = 30\` to the template
- \`server/routes.js\` \`agentchattr-config\` action — same template addition
- \`bin/quadwork.js\` \`migrateLegacyProjects()\` — add the missing-section migration
- \`server/index.js\` — new endpoint to PATCH \`max_agent_hops\` on a running AgentChattr (proxy through to AC's existing room settings endpoint)
- New: \`src/components/LoopGuardWidget.tsx\` — input + apply button + tooltip
- \`src/components/OperatorFeaturesPanel.tsx\` — embed the widget
- New popover/tooltip component if not already available, OR reuse the one from \`AboutModal\`

## Acceptance

- ✅ New projects get \`max_agent_hops = 30\` in their \`config.toml\` automatically
- ✅ Existing projects auto-migrate on next \`quadwork start\`
- ✅ Loop Guard widget shows current value
- ✅ Editing the value + clicking Apply updates the running AgentChattr without restart
- ✅ The (?) tooltip explains the concept clearly for newbies
- ✅ Setting persists across restarts
- ✅ Min/max enforced (4-50)

## References

- AgentChattr code: \`app.py\` lines 274, 287, 308-310, 807-812, 1244-1249
- Config key: \`[routing] max_agent_hops\`

Branch: \`task/272-loop-guard\`

realproject7:
Fixes #274
Tracker: realproject7/agent-os#403
Master: #283 (Batch 30 Phase 2)

## Summary
- Web wizard's `add-config` now writes `[routing] default = "none" / max_agent_hops = 30` to config.toml. (CLI wizard already shipped 30 via templates/config.toml.)
- New `migrateLoopGuardDefaults()` in `bin/quadwork.js`, called from `cmdStart` after `migrateLegacyProjects`. Idempotent — appends the section only when both are missing; never patches an operator-tuned section.
- New `GET/PUT /api/loop-guard` endpoints in `server/routes.js`:
  - GET reads `max_agent_hops` from the project's config.toml.
  - PUT validates 4–50 (matching AC's `update_settings` clamp), rewrites the key in config.toml, and best-effort live-pushes through a new `sendWsEvent()` helper that opens the AC ws and sends `{type: "update_settings", data: {max_agent_hops}}`. Live push failures are non-fatal.
- New `LoopGuardWidget.tsx`: number input + Apply + (?) explainer popover. Embedded in `OperatorFeaturesPanel` between Scheduled Trigger and Telegram Bridge.

## Acceptance criteria
- [x] New projects get `max_agent_hops = 30` automatically (web + CLI both covered)
- [x] Existing projects auto-migrate on next `quadwork start`
- [x] Loop Guard widget shows current value
- [x] Editing the value + Apply updates the running AC without restart (via ws update_settings)
- [x] Tooltip explains the concept
- [x] Setting persists across restarts (config.toml rewrite)
- [x] Min/max enforced (4–50)

## Test plan
- [x] `node --check` on server/routes.js + bin/quadwork.js
- [x] `tsc --noEmit` clean
- [ ] Reviewers: spin up a project, open the dashboard, confirm Loop Guard widget reads 30, change to 35, click Apply, verify config.toml updated and AC's room_settings.json reflects the new value

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1480,6 +1480,37 @@ function migrateLegacyProjects(config) {
   return true;
 }
 
+/**
+ * #403 / quadwork#274 migration: append `[routing] max_agent_hops = 30`
+ * to any per-project config.toml that's missing it. Idempotent and
+ * conservative — leaves files that already have a `max_agent_hops`
+ * key alone (whatever value the operator picked stays put), and
+ * leaves files that have a [routing] section but no max_agent_hops
+ * key alone too rather than risk patching a section we don't fully
+ * understand. The web/CLI templates write the key on first run, so
+ * this only catches projects created before #403.
+ */
+function migrateLoopGuardDefaults(config) {
+  if (!config.projects || config.projects.length === 0) return;
+  for (const project of config.projects) {
+    if (!project.id) continue;
+    const tomlPath = project.agentchattr_dir
+      ? path.join(project.agentchattr_dir, "config.toml")
+      : path.join(CONFIG_DIR, project.id, "agentchattr", "config.toml");
+    if (!fs.existsSync(tomlPath)) continue;
+    let content;
+    try { content = fs.readFileSync(tomlPath, "utf-8"); } catch { continue; }
+    if (/^\s*max_agent_hops\s*=/m.test(content)) continue;
+    if (/^\s*\[routing\]/m.test(content)) continue;
+    const trailing = content.endsWith("\n") ? "" : "\n";
+    const addition = `${trailing}\n[routing]\ndefault = "none"\nmax_agent_hops = 30\n`;
+    try {
+      fs.writeFileSync(tomlPath, content + addition);
+      log(`  Loop guard default → ${project.id} (max_agent_hops = 30)`);
+    } catch { /* non-fatal */ }
+  }
+}
+
 function cmdStart() {
   console.log("\n  QuadWork Start\n");
 
@@ -1492,6 +1523,14 @@ function cmdStart() {
   // own per-project clones before any AgentChattr spawn happens.
   // Idempotent — a no-op once every project already has a working clone.
   migrateLegacyProjects(config);
+
+  // Batch 30 / #403 / quadwork#274: ensure every existing project's
+  // AgentChattr config.toml has [routing] max_agent_hops = 30 so the
+  // loop guard doesn't fire mid-PR-cycle. Idempotent: leaves any
+  // pre-existing routing section alone (only adds the section + key
+  // when both are missing). Failures are non-fatal — the project
+  // will just keep its current loop guard.
+  migrateLoopGuardDefaults(config);
 
   const quadworkDir = path.join(__dirname, "..");
   const port = config.port || 8400;
--- a/server/routes.js
+++ b/server/routes.js
@@ -216,11 +216,21 @@ function sendWsEvent(baseUrl, sessionToken, event) {
 // Source of truth at rest is the project's config.toml [routing]
 // max_agent_hops. The PUT also pushes the value to the running AC via
 // `update_settings` so the change is live without a daemon restart.
+// Resolve the per-project config.toml path through resolveProjectChattr
+// so we honor `project.agentchattr_dir` (web wizard sets this; legacy
+// imports can have arbitrary paths) and don't drift from the rest of
+// the codebase that already goes through that helper.
+function resolveProjectConfigToml(projectId) {
+  const resolved = resolveProjectChattr(projectId);
+  if (!resolved || !resolved.dir) return null;
+  return path.join(resolved.dir, "config.toml");
+}
+
 router.get("/api/loop-guard", (req, res) => {
   const projectId = req.query.project;
   if (!projectId) return res.status(400).json({ error: "Missing project" });
-  const tomlPath = path.join(CONFIG_DIR, projectId, "agentchattr", "config.toml");
-  if (!fs.existsSync(tomlPath)) return res.json({ value: 30, source: "default" });
+  const tomlPath = resolveProjectConfigToml(projectId);
+  if (!tomlPath || !fs.existsSync(tomlPath)) return res.json({ value: 30, source: "default" });
   try {
     const content = fs.readFileSync(tomlPath, "utf-8");
     const m = content.match(/^\s*max_agent_hops\s*=\s*(\d+)/m);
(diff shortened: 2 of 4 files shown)


project7-interns (changes requested):
Medium-severity issue in `server/routes.js` GET and PUT `/api/loop-guard`: both hardcode

```js
const tomlPath = path.join(CONFIG_DIR, projectId, "agentchattr", "config.toml");
```

but every other path-resolving route in this file (and the new migration helper itself, `bin/quadwork.js:1492`) goes through `resolveProjectChattr(projectId)` / `getChattrConfig(projectId)` and respects `project.agentchattr_dir`. For any project with a custom or legacy `agentchattr_dir` (web wizard sets this explicitly at `server/routes.js:924`, and legacy/imported projects can have arbitrary paths), the new endpoints will read/write the wrong file or 404 — and the live-push side will succeed against AC while the on-disk config diverges, so the operator's next restart silently reverts the change.

Fix: resolve the toml path through `resolveProjectChattr(projectId).dir` (mirroring the migration's own logic) in both the GET and the PUT, and surface a clean 404 only when that resolved file is missing.

The rest of the PR looks good — the migration is correctly conservative (skips when `[routing]` already present), the PUT validates 4–50 to match AC's clamp, the persist-then-best-effort-live-push order is right, and `sendWsEvent` mirrors `sendViaWebSocket` cleanly with a parallel 401 close-code path. Will re-approve once the path resolution is unified.

project7-interns (approved):
Fix verified on commit 1d8fee4. New `resolveProjectConfigToml()` helper routes through `resolveProjectChattr(projectId).dir`, both GET and PUT use it, so on-disk config and live AC state can no longer drift for projects with custom `agentchattr_dir`. Helper is null-safe and the 404 path handles unresolved projects cleanly. LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The config-path follow-up fixed one real issue, but the live-update path for the new Loop Guard widget still misses the same stale-session-token recovery that `/api/chat` already needs after an AgentChattr restart.

### Findings
- [medium] Live loop-guard updates silently stop working after an AgentChattr restart because the new ws helper does not retry on stale session tokens.
  - File: `server/routes.js:184`
  - File: `server/routes.js:282`
  - File: `server/routes.js:330`
  - Suggestion: handle `EAGENTCHATTR_401` in the loop-guard PUT path the same way `/api/chat` does: call `syncChattrToken(projectId)`, re-read the refreshed token, and retry `sendWsEvent()` once before falling back to `live=false`.

### Decision
Requesting changes because issue #274 explicitly promises that clicking Apply updates the running AgentChattr without restart. As written, any stale cached session token after an AC restart causes the PUT to persist `config.toml` but skip the live update, so the widget regresses to "takes effect on next AC restart" until the token is resynced some other way.

project7-interns (approved):
Fix verified on commit aaf8443. PUT `/api/loop-guard` now mirrors `/api/chat`'s #230 token-resync pattern: catches `EAGENTCHATTR_401`, calls `syncChattrToken(projectId)`, re-reads via `getChattrConfig`, and retries the `update_settings` ws event once. Non-401 failures still throw to the outer catch and stay non-fatal. Live-update acceptance for #274 holds across AC restarts. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The stale-token regression from my earlier review is fixed. The loop-guard PUT path now mirrors `/api/chat` by re-syncing the AgentChattr session token on `EAGENTCHATTR_401` and retrying the live `update_settings` websocket once.

### Findings
- None. The config-path fix remains in place, and the updated live-push path now preserves the issue #274 promise that Apply updates the running daemon without requiring a restart when the cached session token has gone stale.

### Decision
Approving because the previously blocking live-update gap is addressed and the implementation now matches the existing token-resync pattern already relied on by `/api/chat`.

PR #288, opened 2026-04-08
[#278] Operator display name in chat (replace hardcoded 'user')

### LINKED ISSUE #278: [feat] Operator display name in chat (replace hardcoded 'user')
## Scope

The chat panel sends every operator message with sender hardcoded as \`\"user\"\` (\`server/routes.js\` line 199, set in #230 to prevent agent impersonation). The chat history then shows your messages prefixed with \`user\`.

AgentChattr's native UI lets the operator set their own display name (default \`user\`, customizable in settings). Mirror that.

## Why

1. \`user\` is generic and impersonal
2. When multiple human collaborators eventually share a project, distinct names matter
3. AgentChattr's \`Project History → Export\` will include your sender name — useful for archives

## UI

Add a small \"Your name in chat\" input to QuadWork's Settings page (the global one, not per-project — this is operator identity, same across projects):

\`\`\`
SECTION: Operator Identity
─────────────────────────
Your name in chat: [user____________]
Help: This is what shows next to your messages in the AgentChattr chat panel.
       Defaults to 'user' if blank.
\`\`\`

## Backend

- Add \`operator_name\` to \`~/.quadwork/config.json\` (top-level, not per-project), default \`\"user\"\`
- Validate: 1-32 chars, alphanumeric + dash + underscore only (matches AgentChattr's name validation in \`registry.py\`)
- \`server/routes.js\` \`POST /api/chat\` line 199:
  - Read \`config.operator_name\` from disk
  - Use it as the \`sender\` field instead of hardcoded \`\"user\"\`
- Frontend: Settings page input wired to \`PUT /api/config\`

## Security note

The original #230 hardcoded \`\"user\"\` to prevent **client-supplied** sender impersonation. The fix here is different — we still ignore \`req.body.sender\`, but use a server-side config value (which the operator owns). This preserves the security guarantee: a request hitting QuadWork's \`/api/chat\` cannot pretend to be \`head\` even if it tries to pass \`sender: \"head\"\`. The only valid sender is whatever's in the operator's own config file.

## Acceptance

- ✅ Settings page has an \"Your name in chat\" input, defaults to \`user\`
- ✅ Saving the config persists \`operator_name\` to \`~/.quadwork/config.json\`
- ✅ New chat messages send with the configured operator name as sender
- ✅ Validation: 1-32 chars, alphanumeric + dash + underscore (no \`@\`, no spaces)
- ✅ Empty string falls back to \`\"user\"\`
- ✅ Client-supplied \`sender\` field is still ignored (security from #230 preserved)
- ✅ AgentChattr's UI shows the new name correctly when viewing the chat

## Files

- \`server/routes.js\` — \`/api/chat\` POST handler, read \`operator_name\` from config
- \`server/config.js\` — DEFAULT_CONFIG add \`operator_name: \"user\"\`, plus a validator
- \`src/components/SettingsPage.tsx\` — new \"Operator Identity\" section with the input
- \`bin/quadwork.js\` — \`writeQuadWorkConfig\` write \`operator_name: \"user\"\` if missing on init
- Type updates in \`src/components/SettingsPage.tsx\` \`Config\` interface

Branch: \`task/278-operator-name\`

realproject7:
Fixes #278
Tracker: realproject7/agent-os#405

## Summary
- New `operator_name` field on the global config (default `user`).
- `server/config.js` exports `sanitizeOperatorName()` matching AC's registry name rules (1–32 alnum + dash + underscore); empty / missing / invalid → `user`.
- `server/routes.js` POST `/api/chat` reads + sanitizes `operator_name` and uses it as the AC ws sender. Client-supplied sender is still ignored — #230's anti-impersonation guarantee holds.
- `src/components/SettingsPage.tsx`: new "Operator Identity" section above Global with a "Your name in chat" input + helper text describing the allowed character set.
- `bin/quadwork.js` `writeQuadWorkConfig` stamps `operator_name = "user"` when missing so the field is discoverable on first init.

## Acceptance criteria
- [x] Settings page has an "Your name in chat" input, defaults to `user`
- [x] Saving the config persists `operator_name` (existing PUT /api/config wiring)
- [x] New chat messages send with the configured operator name as sender
- [x] Validation: 1-32 chars, alphanumeric + dash + underscore (server-side strip)
- [x] Empty string falls back to `"user"`
- [x] Client-supplied `sender` field is still ignored
- [ ] Reviewers: confirm AC's UI shows the new name correctly when viewing the chat

## Test plan
- [x] `node --check` on server/routes.js, server/config.js, bin/quadwork.js
- [x] `tsc --noEmit` clean
- [ ] Reviewers: set name to "alice", post a message, confirm `alice` appears in chat (and AC's native UI)

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1172,6 +1172,14 @@ function writeQuadWorkConfig(setup) {
   header("Writing QuadWork Config");
 
   const config = readConfig();
+  // #405 / quadwork#278: ensure the global config has an
+  // operator_name slot. /api/chat reads this on every send and
+  // sanitizes empty/missing values back to "user", so leaving it
+  // unset is safe — but writing the default explicitly makes the
+  // setting discoverable in the on-disk file.
+  if (typeof config.operator_name !== "string") {
+    config.operator_name = "user";
+  }
 
   const project = {
     id: setup.projectName,
--- a/server/config.js
+++ b/server/config.js
@@ -8,8 +8,25 @@ const DEFAULT_CONFIG = {
   port: 8400,
   agentchattr_url: "http://127.0.0.1:8300",
   agentchattr_dir: path.join(os.homedir(), ".quadwork", "agentchattr"),
+  // #405 / quadwork#278: display name used as the chat sender for
+  // messages posted from the dashboard. AC's registry name validator
+  // accepts 1–32 alphanumeric + dash + underscore characters; mirror
+  // that here so the sanitized value matches what AC will accept.
+  operator_name: "user",
   projects: [],
 };
+
+// Sanitize an operator-supplied display name to match AC's name
+// validator (registry.py: 1–32 alnum + dash + underscore). Empty or
+// non-string input falls back to "user". Used both when reading the
+// config (in case the file was hand-edited) and on /api/chat sends
+// (so even a stale on-disk value can't impersonate an agent).
+function sanitizeOperatorName(value) {
+  if (typeof value !== "string") return "user";
+  const cleaned = value.trim().replace(/[^A-Za-z0-9_-]/g, "");
+  if (!cleaned) return "user";
+  return cleaned.slice(0, 32);
+}
 
 // Migration: rename old agent keys to new ones
 const AGENT_KEY_MAP = { t1: "head", t2a: "reviewer1", t2b: "reviewer2", t3: "dev" };
@@ -160,4 +177,4 @@ async function syncChattrToken(projectId) {
   } catch {}
 }
 
-module.exports = { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr, resolveChattrSpawn, syncChattrToken, CONFIG_PATH };
+module.exports = { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr, resolveChattrSpawn, syncChattrToken, sanitizeOperatorName, CONFIG_PATH };
--- a/server/config.js
+++ b/server/config.js
@@ -16,16 +16,41 @@ const DEFAULT_CONFIG = {
   projects: [],
 };
 
+// Reserved sender names that the operator must NOT be able to claim
+// — these are the registered agent identities (current + legacy
+// aliases) plus AC's own "system" sender. Without this denylist a
+// hand-edited or PUT /api/config'd `operator_name = "head"` would
+// post chat messages with sender:"head", reopening the impersonation
+// vector #230 closed. Case-insensitive match.
+const RESERVED_OPERATOR_NAMES = new Set([
+  "head",
+  "dev",
+  "reviewer1",
+  "reviewer2",
+  // Legacy agent aliases — preserved in routing logic in a few
+  // places, so block them too even though new projects no longer
+  // register under these names.
+  "t1",
+  "t2a",
+  "t2b",
+  "t3",
+  // AC's own broadcast / housekeeping sender.
+  "system",
+]);
+
 // Sanitize an operator-supplied display name to match AC's name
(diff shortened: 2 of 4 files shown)


project7-interns (changes requested):
Medium-severity security regression in `server/config.js` `sanitizeOperatorName()`: the sanitizer only enforces AC's charset rules (1–32 alnum + dash + underscore) but does not exclude registered agent names. The operator can therefore set `operator_name = "head"` (or `dev`, `reviewer1`, `reviewer2`, `t1`, `t2a`, `t2b`, `t3`) in Settings, and `/api/chat` will then post messages with `sender: "head"`. This is exactly the impersonation vector that #230 specifically closed by hardcoding `sender: "user"` server-side, and it cannot be papered over by trusting the operator — anyone with PUT access to `/api/config` (the same surface the dashboard exposes) can flip the value.

Downstream consequences once `operator_name = "head"`:
- Reviewer agents that key off `@user` vs agent names get confused about who is talking.
- Routing rules that special-case the `user` sender (e.g. resetting AC's loop guard, the post-merge silence rule) are bypassed.
- The audit trail in chat no longer distinguishes operator from automation.

Fix: add a small denylist in `sanitizeOperatorName` covering at minimum `{head, dev, reviewer1, reviewer2, t1, t2a, t2b, t3, system}` (case-insensitive) and fall back to `"user"` when the cleaned value lands in it. Surface the rule in the SettingsPage helper text so the operator understands why their input was rejected.

The rest of the PR is fine — `readConfigFile` is already in scope (`server/routes.js:26`), the sanitize-on-read approach correctly defends against hand-edited config files, the SettingsPage Input integrates cleanly with the existing `updateGlobal` flow, and `bin/quadwork.js` initializes the field discoverably without breaking idempotence. Will re-approve once the denylist lands.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The character/length sanitization is fine, but the new operator-name path still allows values that collide with real agent IDs. In AgentChattr those collisions are not cosmetic: they change routing and session semantics by making the human look like an agent.

### Findings
- [high] `operator_name` can be set to `head`, `reviewer1`, `reviewer2`, or `dev`, which breaks the anti-impersonation guarantee and causes human messages to be treated as agent messages.
  - File: `server/config.js:24`
  - File: `server/routes.js:334`
  - Suggestion: reject or remap names that collide with configured agent IDs (at minimum the standard QuadWork agent names) before using them as the `/api/chat` sender; otherwise fall back to `user`.

### Decision
Requesting changes because issue #278 explicitly says the #230 security guarantee must be preserved. As written, setting the operator name to an agent ID means AgentChattr will no longer treat those messages as human interruptions/resets, so this is a behavioral and security regression, not just a label collision.

project7-interns (approved):
Fix verified on commit efbc7f9. `RESERVED_OPERATOR_NAMES` denylist covers head/dev/reviewer1/reviewer2 + legacy t1/t2a/t2b/t3 aliases + system, applied case-insensitively after truncation, with fallback to `"user"`. SettingsPage helper text mentions the reserved list. The #230 impersonation guarantee is restored. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The reserved-name follow-up fixes the blocking issue from my earlier review. Operator display names that collide with agent or system identities now fall back to `user`, so the customizable sender no longer changes routing/session semantics.

### Findings
- None. `sanitizeOperatorName()` now rejects the current and legacy agent IDs plus `system`, and the settings help text documents that reserved-name fallback.

### Decision
Approving because the #230 anti-impersonation guarantee is restored while keeping the operator-name feature from #278 intact.

PR #291, opened 2026-04-08
[#271] Color-code @mentions in chat messages

### LINKED ISSUE #271: [feat] Color-code @mentions in chat messages (mirror AgentChattr native UI)
## Scope

Mirror AgentChattr's native chat UI: when a message contains an \`@mention\` of a known agent, render that mention with a colored pill background matching the agent's color (same colors as their status dots and existing sender column).

## Reference

AgentChattr's native UI at \`http://127.0.0.1:{port}\` already does this. In the second screenshot of this ticket you can see:

- \`@user\` rendered with a purple pill
- \`@head\` rendered with a green pill
- \`@dev\` rendered with an orange pill
- \`@reviewer1\` rendered with a green pill
- \`@reviewer2\` rendered with an orange pill

In QuadWork's current chat panel, the \`@mention\` text is rendered as plain text — just visually noisy without the color cue.

## Behavior

- Parse each message body for tokens matching \`/@(head|dev|reviewer1|reviewer2|user)\\b/i\`
- Each match becomes a small inline pill with:
  - Background color = the agent's color (use the existing \`SENDER_COLORS\` map)
  - Slightly darker text on the colored background, OR keep the agent color and put a subtle background tint behind it (match AgentChattr's style — pick whichever reads better)
  - Rounded corners, small padding, monospace font (consistent with the rest of the chat)
- Non-mention text stays unchanged

## Edge cases

- Don't pill non-agent mentions like \`@somebody\`
- Don't pill inside URLs (e.g., \`https://github.com/@user\`) — match only when preceded by whitespace or start of string

## Files

- \`src/components/ChatPanel.tsx\` — message body rendering. Add a small helper that splits the text on agent-mention regex matches and wraps the matches in a styled \`<span>\`
- Reuse the existing \`SENDER_COLORS\` map
- Apply to both modes: API fallback (\`ChatPanelAPI\`) and the iframe path (iframe already shows the native UI which has this for free)

## Acceptance

- ✅ Each agent \`@mention\` in a chat message renders as a colored pill matching its sender color
- ✅ \`@user\` mentions also pill (in user color)
- ✅ Unknown \`@whatever\` mentions stay as plain text
- ✅ URLs containing \`@\` aren't accidentally pilled
- ✅ Pills look consistent with the chat font / line height (don't break message wrapping)

Branch: \`task/271-mention-color-pills\`

realproject7:
Fixes #271
Tracker: realproject7/agent-os#408

## Summary
- New `renderMessageWithMentions()` helper splits chat text on a regex `(^|\s)(@(?:head|dev|reviewer1|reviewer2|user))\b` (case-insensitive) and wraps each match in a small colored pill using the existing `SENDER_COLORS` map.
- Background uses a 22-alpha tint of the agent color so the text stays legible against the chat font; pills are inline-block rounded font-mono so line wrapping stays intact.
- URL safety: the mention must be preceded by start-of-string or whitespace, so `https://github.com/@user` does not pillify.
- Iframe path already inherits the same look from AC's native UI; this only touches the API fallback rendering in `ChatPanelAPI`.

## Acceptance criteria
- [x] Each agent `@mention` renders as a colored pill matching its sender color
- [x] `@user` mentions also pill (in user color)
- [x] Unknown `@whatever` mentions stay as plain text
- [x] URLs containing `@` aren't accidentally pilled (lookahead-equivalent via the leading whitespace group)
- [x] Pills look consistent with the chat font / line height

## Test plan
- [x] `tsc --noEmit` clean
- [ ] Reviewers: post `@head ping @reviewer1` and confirm both pill, post `@somebody hi` and confirm plain, post a URL like `https://github.com/@user` and confirm no pill

### DIFF
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useEffect, useRef, useCallback } from "react";
+import React, { useState, useEffect, useRef, useCallback } from "react";
 import ProjectChatEmptyState from "./ProjectChatEmptyState";
 
 interface Message {
@@ -36,6 +36,41 @@ function senderLabel(sender: string): string {
   if (s === "reviewer1") return "RE1";
   if (s === "reviewer2") return "RE2";
   return sender;
+}
+
+// #408 / quadwork#271: render @mentions of known agents as colored
+// pills. Only the four agent IDs + "user" qualify; unknown handles
+// stay as plain text. The mention must be preceded by start-of-string
+// or whitespace so URL fragments like "https://github.com/@user"
+// don't accidentally pillify.
+const MENTION_RE = /(^|\s)(@(?:head|dev|reviewer1|reviewer2|user))\b/gi;
+function renderMessageWithMentions(text: string): React.ReactNode[] {
+  const parts: React.ReactNode[] = [];
+  let last = 0;
+  let key = 0;
+  for (const match of text.matchAll(MENTION_RE)) {
+    const fullStart = match.index ?? 0;
+    const lead = match[1];
+    const mention = match[2];
+    const mentionStart = fullStart + lead.length;
+    if (mentionStart > last) {
+      parts.push(text.slice(last, mentionStart));
+    }
+    const agentId = mention.slice(1).toLowerCase();
+    const color = senderColor(agentId);
+    parts.push(
+      <span
+        key={`m${key++}`}
+        className="inline-block px-1 py-0 rounded font-mono"
+        style={{ backgroundColor: `${color}22`, color }}
+      >
+        {mention}
+      </span>,
+    );
+    last = mentionStart + mention.length;
+  }
+  if (last < text.length) parts.push(text.slice(last));
+  return parts.length > 0 ? parts : [text];
 }
 
 /**
@@ -284,7 +319,7 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
               {senderLabel(msg.sender)}
             </span>
             <span className="text-text break-words min-w-0 whitespace-pre-wrap flex-1">
-              {msg.text}
+              {renderMessageWithMentions(msg.text)}
             </span>
             {/* #397 / quadwork#262: reply affordance — small grey,
                 hover-revealed so it doesn't visually compete with the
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
LGTM. Mention pattern is well-scoped: `(^|\s)` requires whitespace or BOL so URL fragments like `https://github.com/@user` don't pillify, `\b` correctly rejects `@head1`/`@reviewer12` while allowing punctuation/EOL after the canonical IDs, /i + .toLowerCase() handles capitalization, and matchAll is used safely with explicit `mentionStart` so the leading whitespace is preserved as plain text. Adjacent mentions like `@head @dev` render correctly (space between is plain text, second match's lead-space accounted for). Pill styling reuses senderColor with a 22-alpha tint. Single-file scope.\n\nMicro doc nit (non-blocking): the comment says "URL safety via leading whitespace lookahead" but it's a captured group, not a lookahead. Worth a one-word fix on a follow-up.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The pill rendering works for the basic cases, but the current regex still partially matches longer dash-suffixed handles. That means some unknown mentions render as if they were known agents, which violates the issue's plain-text requirement for unknown handles.

### Findings
- [medium] Dash-suffixed unknown handles like `@user-name` and `@head-2` partially pillify as `@user` / `@head`.
  - File: `src/components/ChatPanel.tsx:42`
  - Suggestion: tighten the trailing boundary so a known handle only pills when it is followed by end-of-string, whitespace, or punctuation that actually terminates the token — not a hyphenated continuation. As written, `` treats `-` as a boundary, so the helper misclassifies longer unknown handles.

### Decision
Requesting changes because issue #271 explicitly says unknown `@whatever` mentions must stay plain text. The current regex fails that rule for common hyphenated handles, and QuadWork already has real names in this ecosystem that use dashes.

project7-interns (approved):
Fix verified on commit b609454. Trailing `\b` replaced with `(?![\w-])` lookahead so longer hyphenated handles don't pill their known prefix. Comment expanded to document the boundary semantics. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The regex follow-up fixes the blocking edge case from my earlier review. Known mentions still render as pills, while longer hyphenated unknown handles like `@user-name` and `@head-2` now remain plain text.

### Findings
- None. The updated negative lookahead prevents partial matches on continued handles without regressing the intended start/whitespace guard for standalone mentions.

### Decision
Approving because the unknown-handle regression is addressed and the mention-pill rendering now matches issue #271’s behavior requirements.

PR #292, opened 2026-04-08
[#273] Notification sound on new agent messages

### LINKED ISSUE #273: [feat] Notification sound on new agent messages (mirror AgentChattr native UI)
## Scope

When a new chat message arrives from an agent (not the user themselves), play a small notification sound. Mirror AgentChattr's native UI behavior — which already does this for its own web UI.

Add a sound control to the SYSTEM section of QuadWork, next to **Keep Awake**.

## Reference

AgentChattr already ships notification sounds in `{install_dir}/static/sounds/`:
- `soft-chime.mp3` (default)
- `warm-bell.mp3`
- `click.mp3`
- `alert-tone.mp3`
- `pluck.mp3`

In its own web UI (`chat.js` line 399), it plays `new Audio('/static/sounds/{name}.mp3').play()` whenever a new message arrives from someone other than the current user, **only when the tab is not focused** (avoids spamming when the user is actively reading).

AgentChattr's settings panel exposes:
- **Default sound** — chime played for any new message
- **Background alerts** — separate sound for messages received while the tab is in the background
- **Per-agent sound override** (head/dev/...) — power-user feature

## What QuadWork should ship

Match AgentChattr's core behavior but keep the UI minimal:

### 1. Sound enable toggle
- Toggle button in SYSTEM section: `🔔 Sound` / `🔕 Sound`
- Persist in `localStorage` key `quadwork_notification_sound` (`"on"` | `"off"`)
- Default: `"on"`

### 2. Sound choice (small dropdown next to toggle)
- 5 options: Soft Chime (default) / Warm Bell / Click / Alert Tone / Pluck
- Persist in `localStorage` key `quadwork_notification_sound_choice`
- Default: `"soft-chime"`

### 3. Background-only mode (checkbox below the toggle)
- `[ ] Only when tab is in background`
- When enabled: only play if `!document.hasFocus()` — avoids dings while the user is reading
- When disabled: play on every message (current proposal)
- Default: **enabled** (less annoying)
- Persist in `localStorage` key `quadwork_notification_sound_background_only`

### 4. Skip these cases regardless
- Message from `user` / operator's own name (you don't ding for your own messages)
- Message type `join` / `leave` / `system` (only `chat` triggers a sound)

## Sound source

**Bundle in QuadWork** — copy 5 MP3s into `public/sounds/`:
- `public/sounds/soft-chime.mp3`
- `public/sounds/warm-bell.mp3`
- `public/sounds/click.mp3`
- `public/sounds/alert-tone.mp3`
- `public/sounds/pluck.mp3`

Smallest disk impact (~50-100 KB total). Copy from `/Users/cho/.quadwork/dropcast/agentchattr/static/sounds/` (same files AgentChattr ships).

No proxy/dependency on AgentChattr being running.

## Files

- New: `public/sounds/*.mp3` (5 files copied from AgentChattr)
- `src/components/ChatPanel.tsx` — detect new agent messages, play selected sound (skip user, skip system, skip if `background_only && document.hasFocus()`)
- `src/components/ControlBar.tsx` — add Sound toggle + dropdown + background-only checkbox in SYSTEM section
- New small hook: `useNotificationSound()` — load Audio elements, expose `play()` + settings state, persist to localStorage

## Acceptance

- ✅ New message from any agent (head, dev, reviewer1, reviewer2) plays the selected sound
- ✅ User's own messages do NOT play a sound
- ✅ System / join / leave events do NOT play a sound
- ✅ Sound toggle enables/disables the sound globally
- ✅ Sound choice dropdown lets user pick from 5 chimes
- ✅ Background-only checkbox suppresses sound when the tab is focused
- ✅ All preferences persist in localStorage
- ✅ Defaults: enabled, Soft Chime, background-only on

Branch: `task/273-notification-sound`

realproject7:
Fixes #273
Tracker: realproject7/agent-os#409

## Summary
- Bundle 5 MP3s in `public/sounds/` (soft-chime, warm-bell, click, alert-tone, pluck) — ~220KB total, copied from AgentChattr's bundled set.
- New `src/lib/notificationSound.ts`: typed enum + localStorage-backed enable / choice / background-only prefs, plus `playNotificationSound()` that honors all three.
- `ChatPanel` poll path detects new agent messages (skip `user`, skip `system`/join/leave, skip initial backfill via `cursorRef > 0`) and calls `playNotificationSound()` outside the `setMessages` updater so strict mode can't double-fire.
- `ControlBar` `SystemSection` gains a Sound toggle, choice dropdown (with on-pick preview), and "Only when tab is in background" checkbox. Hydrated from localStorage on mount so the controls reflect what `ChatPanel` is reading. The Keep Awake button now hides on non-darwin instead of returning the whole section `null`, so sound controls show on Linux/Windows too.

## Acceptance criteria
- [x] New message from any agent (head, dev, reviewer1, reviewer2) plays the selected sound
- [x] User's own messages do NOT play a sound (`m.sender !== "user"`)
- [x] System / join / leave events do NOT play a sound (`m.type === undefined || m.type === "chat"` + `m.sender !== "system"`)
- [x] Sound toggle enables/disables the sound globally
- [x] Sound choice dropdown lets user pick from 5 chimes
- [x] Background-only checkbox suppresses sound when the tab is focused
- [x] All preferences persist in localStorage
- [x] Defaults: enabled, Soft Chime, background-only on

## Test plan
- [x] `tsc --noEmit` clean
- [ ] Reviewers: open dashboard, post `@head ping`, get a head reply, hear soft chime; mute via Sound button; switch sound to Pluck (preview plays); enable foreground-too checkbox; reload page and confirm prefs persist


project7-interns (approved):
LGTM. Solid notification sound implementation.

- `notificationSound.ts` is well-isolated: SSR-safe localStorage helpers, default-on with sensible bg-only default, swallowed autoplay rejection, and a clean `playNotificationSound()` that honors enable + bg-only + tab focus checks.
- ChatPanel detection runs OUTSIDE `setMessages` so React strict mode's double-invoke can't double-fire audio. Filter correctly skips own/system messages and only fires when `previousCursor > 0` so the initial backfill is silent.
- ControlBar correctly hydrates from localStorage in a `useEffect` (not initial state) so SSR + first-paint stay consistent.
- Choice dropdown previews on change, which is the right UX.
- Hiding Keep Awake button on non-darwin while keeping sound controls visible is clean.

Micro nits (non-blocking):
1. The new `import` block in `ControlBar.tsx` lives mid-file between `ServerSection` and `SystemSection`. ESM accepts top-level imports anywhere, but most linters/conventions want them grouped at the top. Easy follow-up.
2. Edge case: if the very first poll returns zero messages, `cursorRef` stays 0 and the very next incoming message is treated as initial backfill (silent). Tiny window, only matters on a brand-new project — fine for v1.

No binary file content review — trusting the bundled MP3s are licensed appropriately.

project7-interns (approved):
Re-approve on commit 204c8c0 — import block moved to top of file alongside the React import. Pure code-motion, no behavior change.

project7-interns (approved):
Fix verified on commit 84aabf5. Good cross-feature catch by t2a — #278 lets operator_name diverge from the literal "user". Fix fetches /api/config once on mount, stores in operatorNameRef so the hot poll path avoids a state dep, and adds it to the filter alongside the literal "user" fallback (so unconfigured projects still work). Self-exclusion now holds even when operator_name="alice". LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new self-notification fix still compares incoming senders against the raw `operator_name` returned by `/api/config`, but `/api/chat` emits the sanitized operator sender. That leaves self-message exclusion incorrect whenever the configured name needs sanitization.

### Findings
- [medium] Self-message exclusion still breaks when `operator_name` is sanitized server-side
  - File: `src/components/ChatPanel.tsx:181`
  - File: `server/routes.js:43`
  - File: `server/routes.js:337`
  - Suggestion: derive the exclusion name from the same sanitized value `/api/chat` uses, or have `/api/config` return the sanitized `operator_name` instead of the raw file value.

### Decision
Requesting changes because issue #273 requires suppressing the operator's own messages, and the current fix still dings on self-messages for inputs like `alice!` (sent as `alice`) while also potentially suppressing real agent messages for reserved raw names like `head`.

project7-interns (approved):
Fix verified on commit d130499. GET `/api/config` now overlays `sanitizeOperatorName()` so every reader (ChatPanel filter, SettingsPage input, future consumers) sees the effective sender that `/api/chat` stamps. On-disk file keeps the raw value so a follow-up can show raw vs effective. Both the "alice!" → "alice" garbage case and the reserved "head" → "user" case now match between filter and sender. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up fix now makes `/api/config` return the same sanitized `operator_name` that `/api/chat` uses as the sender, so the notification filter compares against the effective operator identity instead of the raw config value.

### Findings
- No blocking findings.

### Decision
Approve because the self-message exclusion now matches issue #273's acceptance criteria even when the configured operator name contains stripped characters or falls back from a reserved name.

Issue #228, opened 2026-04-07
[#225 sub-C] Auto-tag @head when no agent mentioned in chat


realproject7:
Parent: #225

## Scope

In the AgentChattr chat panel, if the operator types a message without tagging any agent, the message goes to no one in particular and the operator gets no response. To fix this, auto-prepend \`@head \` to messages that don't already mention a known agent.

## Behavior

| Input | Sent as |
|-------|---------|
| \`hello\` | \`@head hello\` |
| \`fix the bug\` | \`@head fix the bug\` |
| \`@head fix the bug\` | \`@head fix the bug\` (unchanged) |
| \`@dev start working\` | \`@dev start working\` (unchanged) |
| \`@reviewer1 status?\` | \`@reviewer1 status?\` (unchanged) |
| \`@reviewer1 @reviewer2 review #100\` | \`@reviewer1 @reviewer2 review #100\` (unchanged) |
| \`@somebody hi\` | \`@head @somebody hi\` (unknown tag → still needs Head) |

## Detection logic

Known agents: \`head\`, \`dev\`, \`reviewer1\`, \`reviewer2\`. Match \`/@(head|dev|reviewer1|reviewer2)\\b/i\` against the message text. If no match, prepend \`@head \`.

## Files to change

- \`src/components/ChatPanel.tsx\` — \`send()\` function: apply the prepend logic before posting to \`/api/chat\`

## Acceptance

- ✅ Message without any known agent tag → \`@head \` prepended
- ✅ Message with at least one known agent tag → unchanged
- ✅ Unknown \`@whatever\` tags don't count as agent tags — \`@head\` still gets prepended
- ✅ Visual: input field still shows what user typed, but the message in chat shows the prepended version

Branch: \`task/225c-auto-tag-head\`

Issue #200, opened 2026-04-07
[fix] Chat input — add Send button, visual feedback, Tab autocomplete


realproject7:
AgentChattr chat input has no send button, no error feedback when send fails, and no Tab-to-complete for @mentions. Users can't tell if Enter is working.

PR #293, opened 2026-04-08
[#276] Slash command autocomplete + skip auto-tag for /messages

### LINKED ISSUE #276: [feat] Slash commands — autocomplete menu + skip auto-tag for / messages
## Scope

AgentChattr's native UI supports slash commands like \`/continue\`, \`/clear\`, \`/summary\`, etc. QuadWork's chat panel currently does not surface these — and worse, the auto-tag-@head feature from #228 wraps slash commands as \`@head /continue\` which breaks them.

### 1. Skip auto-tag for slash commands

In \`src/components/ChatPanel.tsx\` \`send()\` (the auto-prepend logic from #228), check if the message starts with \`/\` and **skip the @head prepend** in that case. Slash commands are routed by AgentChattr itself and don't need an agent tag.

Update the matching rule:

\`\`\`javascript
const trimmed = input.trim();
const startsWithSlash = trimmed.startsWith(\"/\");
const hasAgentTag = /@(head|dev|reviewer1|reviewer2)\\b/i.test(trimmed);
const finalText = (!startsWithSlash && !hasAgentTag) ? \`@head \${trimmed}\` : trimmed;
\`\`\`

### 2. Slash command autocomplete menu

When the user types \`/\` in the chat input, show a popup menu listing the available commands with descriptions. Selecting one (click or arrow + enter) inserts it. Mirror AgentChattr's native UI behavior.

## Command list

Mirror AgentChattr's exact list (from \`agentchattr/static/chat.js\` lines 1978-1989):

| Command | Description |
|---------|-------------|
| \`/artchallenge\` | SVG art challenge — all agents create artwork (optional theme) |
| \`/hatmaking\` | All agents design a hat to wear on their avatar |
| \`/roastreview\` | Get all agents to review and roast each other's work |
| \`/poetry haiku\` | Agents write a haiku about the codebase |
| \`/poetry limerick\` | Agents write a limerick about the codebase |
| \`/poetry sonnet\` | Agents write a sonnet about the codebase |
| \`/summary\` | Summarize recent messages — tag an agent (e.g. \`/summary @head\`) |
| \`/continue\` | Resume after loop guard pauses |
| \`/clear\` | Clear messages in current channel |

## UI

Popup menu styling: dark popover anchored to the chat input, list of commands with the slash command in accent color and the description in muted text. Mirror the screenshot's layout.

Behavior:
- Trigger: input starts with \`/\`
- Filter: as the user types more characters, narrow the list (e.g., typing \`/po\` shows only \`/poetry haiku\`, \`/poetry limerick\`, \`/poetry sonnet\`)
- Selection: arrow keys + Enter, or click
- Dismiss: Escape, or typing a space (except for \`/poetry\`)
- Reuse the same component pattern as the existing @mention menu

## Files

- \`src/components/ChatPanel.tsx\`:
  - Add a \`SLASH_COMMANDS\` constant matching the AgentChattr list
  - Add slash menu state (similar to existing \`showMentions\`/\`mentionFilter\`)
  - Update \`send()\` to skip auto-tag for slash messages
  - Add slash menu rendering near the existing mention menu

## Acceptance

- ✅ Typing \`/\` opens the slash command menu with all 9 commands listed
- ✅ Typing more characters filters the list
- ✅ Selecting a command inserts it into the input
- ✅ Sending a slash command does NOT auto-prepend \`@head\`
- ✅ Sending a regular message still auto-prepends \`@head\` (#228 unchanged)
- ✅ Escape closes the menu
- ✅ Tab autocompletes the first match (consistent with the @mention menu from #200)

Branch: \`task/275-slash-commands\`

realproject7:
Fixes #276
Tracker: realproject7/agent-os#410

## Summary
Two changes to `src/components/ChatPanel.tsx`:

1. `send()` skips the `@head` auto-prepend (#228) when the buffer starts with `/`. Slash commands are routed by AgentChattr itself; `@head /continue` would silently break them.
2. New `SLASH_COMMANDS` constant mirrors AC's static list (artchallenge, hatmaking, roastreview, poetry haiku/limerick/sonnet, summary, continue, clear). Typing `/` opens an autocomplete dropdown showing each command in accent color + description in muted text. The menu stays open while typing single-word commands and special-cases `/poetry` so the variants keep filtering after the space. Tab autocompletes the first match (consistent with existing @mention menu); Escape dismisses; clicking inserts.

## Acceptance criteria
- [x] Typing `/` opens the slash command menu with all 9 commands listed
- [x] Typing more characters filters the list
- [x] Selecting a command inserts it into the input
- [x] Sending a slash command does NOT auto-prepend `@head`
- [x] Sending a regular message still auto-prepends `@head` (#228 unchanged)
- [x] Escape closes the menu
- [x] Tab autocompletes the first match (consistent with @mention menu from #200)

## Test plan
- [x] `tsc --noEmit` clean
- [ ] Reviewers: type `/`, confirm 9 commands, type `/po`, confirm 3 poetry variants, click/Tab `/poetry haiku`, send, confirm AC sees the literal `/poetry haiku` (no `@head` prefix)

### DIFF
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -24,6 +24,21 @@ const SENDER_COLORS: Record<string, string> = {
 
 const AGENTS = ["head", "reviewer1", "reviewer2", "dev", "user"];
 
+// #410 / quadwork#276: slash commands recognized by AgentChattr's
+// native UI (mirrors agentchattr/static/chat.js lines 1978-1989).
+// Typing `/` in the chat input opens an autocomplete menu of these.
+const SLASH_COMMANDS: { command: string; description: string }[] = [
+  { command: "/artchallenge", description: "SVG art challenge — all agents create artwork (optional theme)" },
+  { command: "/hatmaking", description: "All agents design a hat to wear on their avatar" },
+  { command: "/roastreview", description: "Get all agents to review and roast each other's work" },
+  { command: "/poetry haiku", description: "Agents write a haiku about the codebase" },
+  { command: "/poetry limerick", description: "Agents write a limerick about the codebase" },
+  { command: "/poetry sonnet", description: "Agents write a sonnet about the codebase" },
+  { command: "/summary", description: "Summarize recent messages — tag an agent (e.g. /summary @head)" },
+  { command: "/continue", description: "Resume after loop guard pauses" },
+  { command: "/clear", description: "Clear messages in current channel" },
+];
+
 function senderColor(sender: string): string {
   return SENDER_COLORS[sender.toLowerCase()] || "#e0e0e0";
 }
@@ -164,6 +179,11 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
   const [input, setInput] = useState("");
   const [showMentions, setShowMentions] = useState(false);
   const [mentionFilter, setMentionFilter] = useState("");
+  // #410 / quadwork#276: slash command autocomplete menu state.
+  // Opens whenever the input *starts* with `/` and there's no later
+  // space (so `/poetry hai` still matches but `/clear extra` exits
+  // back to free-typed mode).
+  const [showSlashMenu, setShowSlashMenu] = useState(false);
   const [authError, setAuthError] = useState<string | null>(null);
   // #397 / quadwork#262: tracks the message the next send will be a
   // threaded reply to. Cleared after a successful send or on cancel.
@@ -275,8 +295,14 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
     // `@head ` so the message has somewhere to go. Unknown
     // `@whatever` mentions don't count — Head is still added.
     // Known agents: head, dev, reviewer1, reviewer2.
+    //
+    // #410 / quadwork#276: slash commands are routed by AgentChattr
+    // itself and must NOT be wrapped in `@head /continue` — that
+    // would silently break them. Skip the auto-tag when the message
+    // starts with `/`.
     const KNOWN_AGENT_RE = /@(head|dev|reviewer1|reviewer2)\b/i;
-    const text = KNOWN_AGENT_RE.test(raw) ? raw : `@head ${raw}`;
+    const startsWithSlash = raw.startsWith("/");
+    const text = (startsWithSlash || KNOWN_AGENT_RE.test(raw)) ? raw : `@head ${raw}`;
     setSending(true);
     setSendError(null);
     fetch(`/api/chat${projectId ? `?project=${encodeURIComponent(projectId)}` : ""}`, {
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
LGTM. Slash command autocomplete + auto-tag skip both look correct.

- `send()` correctly bypasses the `@head` auto-prepend when the message starts with `/`, so AC routes the command instead of receiving `@head /continue`.
- `handleInput` opens the menu only when input starts with `/`, special-cases `/poetry` so the variant menu stays open, and closes on space for every other command — matches the documented one-multi-word-command exception.
- Filter `c.command.toLowerCase().startsWith(input.toLowerCase())` narrows correctly as the operator types, and the render guard `showSlashMenu && filteredSlashCommands.length > 0` cleanly hides the dropdown once the buffer no longer matches anything (e.g. typing past `/poetry haiku `).
- Tab handler prefers `@mention` when both menus could theoretically be open (only one is at a time, but the ordering is safe), Esc clears both menus, Enter submits.
- `insertSlashCommand` distinguishes multi-word vs single-word inserts so `/summary @head` still gets the trailing space.

No backend changes; single-file scope matches the issue.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The slash autocomplete UI renders, but the keyboard interaction is incomplete relative to issue #276 and AgentChattr's native behavior.

### Findings
- [medium] Enter still sends the raw partial slash buffer instead of selecting from the slash menu, and arrow-key navigation is missing
  - File: `src/components/ChatPanel.tsx:362`
  - File: `src/components/ChatPanel.tsx:506`
  - Suggestion: add an active slash-menu index with ArrowUp/ArrowDown handling, and when the slash menu is open make Enter select the active command instead of immediately calling `send()`. Right now typing `/po` and pressing Enter sends `/po` to AgentChattr instead of choosing a `/poetry ...` command, which breaks the native slash-command workflow the issue asks you to mirror.

### Decision
Requesting changes because issue #276 explicitly calls for arrow-key + Enter selection behavior, and the current implementation can submit incomplete invalid slash commands from the autocomplete state.

project7-interns (approved):
Fix verified on commit 3637577. New `slashIndex` state, ArrowUp/Down wrap-cycle the active row, Tab/Enter commit it via `insertSlashCommand` with safe index clamping (`Math.min(slashIndex, len-1)`), the visible row highlights via active class, and onMouseEnter keeps mouse + keyboard nav in sync. handleInput resets slashIndex on filter change so the selection can't dangle past a now-shorter list. Typing partial commands no longer posts the literal buffer. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up fix adds the missing slash-menu keyboard behavior: ArrowUp/ArrowDown now move the active row, and Tab/Enter commit that row instead of sending the partial slash buffer.

### Findings
- No blocking findings.

### Decision
Approve because the slash autocomplete now satisfies issue #276's requested behavior while preserving the `/` no-auto-`@head` send path.

PR #301, opened 2026-04-08
[#298] Loop guard migration: handle partial [routing] section

### LINKED ISSUE #298: [bug] Loop guard migration skips projects with partial [routing] section
Follow-up to PR #286 (issue #274).

## Problem

\`migrateLoopGuardDefaults\` in PR #286 only patches projects whose \`config.toml\` is **completely missing** the \`[routing]\` section. Projects that have a \`[routing]\` section for some other reason (e.g., \`default = \"none\"\`) but don't have a \`max_agent_hops\` key are silently skipped — they retain AgentChattr's default of 4 and will keep firing the loop guard.

## Fix

Update the migration to also patch projects with a partial \`[routing]\` section. Detect via TOML parse:

\`\`\`javascript
function patchLoopGuard(configToml) {
  const parsed = parseToml(configToml);
  if (!parsed.routing) {
    // No section at all → append
    return configToml + \"\\n[routing]\\nmax_agent_hops = 30\\n\";
  }
  if (parsed.routing.max_agent_hops === undefined) {
    // Section exists but key missing → insert key into existing section
    return insertKeyIntoSection(configToml, \"routing\", \"max_agent_hops\", 30);
  }
  // Already set → no change
  return configToml;
}
\`\`\`

The \`insertKeyIntoSection\` helper should handle line-based insertion safely (find the \`[routing]\` line, insert \`max_agent_hops = 30\` after it).

Alternatively (simpler): use a TOML library to parse → mutate → serialize. QuadWork might already have a TOML parser somewhere; check before adding a dep.

## Acceptance

- ✅ Project with no \`[routing]\` section → gets full section appended
- ✅ Project with partial \`[routing]\` section (no \`max_agent_hops\`) → gets the key inserted into the existing section
- ✅ Project with \`max_agent_hops\` already set → no change (idempotent)
- ✅ Migration runs on \`cmdStart\` (existing call site)
- ✅ Comments preserved if using line-based insert (don't reformat the whole file)

## Files

- \`bin/quadwork.js\` — \`migrateLoopGuardDefaults\` function from PR #286

Branch: \`task/298-loop-guard-migration-gap\`

realproject7:
Fixes #298
Tracker: realproject7/agent-os#415
Follow-up: quadwork#274, PR #286

## Summary
`migrateLoopGuardDefaults` previously skipped projects with a `[routing]` section that lacked `max_agent_hops` (e.g. only `default = "none"`). Those projects silently kept AC's default of 4. Now handles three cases idempotently with line-based edits:

1. `max_agent_hops` already set → no change
2. `[routing]` exists, key missing → insert key on the line after the header
3. No `[routing]` section → append fresh section

The case-2 insertion uses an anchored `^` regex on the header so a `[routing]` substring inside a string value can't false-match.

## Acceptance criteria
- [x] Project with no `[routing]` section → gets full section appended
- [x] Project with partial `[routing]` section → gets the key inserted into the existing section
- [x] Project with `max_agent_hops` already set → no change (idempotent)
- [x] Migration runs on `cmdStart` (existing call site untouched)
- [x] Comments preserved (line-based insert, not parse + serialize)

## Test plan
- [x] `node --check bin/quadwork.js`
- [x] Verified all three transitions against synthetic config.toml fixtures (smoke script)
- [ ] Reviewers: run `quadwork start` against a project with each of the three states, confirm the file mutates as expected and re-running is a no-op

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1515,8 +1515,17 @@ function migrateLoopGuardDefaults(config) {
     if (!fs.existsSync(tomlPath)) continue;
     let content;
     try { content = fs.readFileSync(tomlPath, "utf-8"); } catch { continue; }
-    // Case 1: key already present anywhere in the file. Idempotent.
-    if (/^\s*max_agent_hops\s*=/m.test(content)) continue;
+    // Case 1: key already present *inside* the [routing] section.
+    // We must NOT short-circuit on a same-named key in some other
+    // table (e.g. `[other]\nmax_agent_hops = 7`) because that would
+    // leave a real partial [routing] section unpatched. Slice the
+    // file to just the [routing] body before the next [section]
+    // header (or EOF) and look there.
+    const routingBody = (() => {
+      const m = content.match(/^\s*\[routing\][ \t]*\r?\n([\s\S]*?)(?=^\s*\[|\z)/m);
+      return m ? m[1] : null;
+    })();
+    if (routingBody !== null && /^\s*max_agent_hops\s*=/m.test(routingBody)) continue;
     let next;
     if (/^\s*\[routing\]/m.test(content)) {
       // Case 2: section exists, key missing. Insert the key on the
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1546,9 +1546,12 @@ function migrateLoopGuardDefaults(config) {
       // line right after the [routing] header so it's scoped to the
       // section regardless of what other keys / comments live there.
       // Anchored to ^ so a `[routing]` substring inside a string
-      // value can't false-match.
+      // value can't false-match. The header line is allowed to have
+      // a trailing inline comment ("[routing] # keep me") so the
+      // insert still lands when the operator has annotated the
+      // section.
       next = content.replace(
-        /^(\s*\[routing\][ \t]*\r?\n)/m,
+        /^(\s*\[routing\][^\r\n]*\r?\n)/m,
         `$1max_agent_hops = 30\n`,
       );
     } else {
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1546,13 +1546,19 @@ function migrateLoopGuardDefaults(config) {
       // line right after the [routing] header so it's scoped to the
       // section regardless of what other keys / comments live there.
       // Anchored to ^ so a `[routing]` substring inside a string
-      // value can't false-match. The header line is allowed to have
-      // a trailing inline comment ("[routing] # keep me") so the
-      // insert still lands when the operator has annotated the
-      // section.
+      // value can't false-match. The header line is allowed to:
+      //   - have a trailing inline comment (`[routing] # keep me`)
+      //   - end the file with no trailing newline at all
+      // The line-break group is captured separately and re-emitted,
+      // synthesizing a `\n` if the file ended exactly at the header
+      // (otherwise the new key would land glued to whatever the
+      // bracket was sitting on).
       next = content.replace(
-        /^(\s*\[routing\][^\r\n]*\r?\n)/m,
-        `$1max_agent_hops = 30\n`,
+        /^(\s*\[routing\][^\r\n]*)(\r?\n|$)/m,
+        (_match, header, lineBreak) => {
+          const lb = lineBreak || "\n";
+          return `${header}${lb}max_agent_hops = 30\n`;
+        },
       );
     } else {
       // Case 3: no section. Append a fresh one at the end of the
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1518,14 +1518,28 @@ function migrateLoopGuardDefaults(config) {
     // Case 1: key already present *inside* the [routing] section.
     // We must NOT short-circuit on a same-named key in some other
     // table (e.g. `[other]\nmax_agent_hops = 7`) because that would
-    // leave a real partial [routing] section unpatched. Slice the
-    // file to just the [routing] body before the next [section]
-    // header (or EOF) and look there.
-    const routingBody = (() => {
-      const m = content.match(/^\s*\[routing\][ \t]*\r?\n([\s\S]*?)(?=^\s*\[|\z)/m);
-      return m ? m[1] : null;
-    })();
-    if (routingBody !== null && /^\s*max_agent_hops\s*=/m.test(routingBody)) continue;
+    // leave a real partial [routing] section unpatched. Walk the
+    // file line-by-line, find the [routing] header, and collect
+    // lines until the next [section] header or EOF. JS regex has
+    // no \z anchor and no reliable EOF lookahead inside /m, so a
+    // line scan is the simplest correct approach and also keeps
+    // pathological strings ("default = \"lazy\"") from confusing
+    // anything in the matcher.
+    const lines = content.split(/\r?\n/);
+    let inRouting = false;
+    let routingKeyPresent = false;
+    for (const line of lines) {
+      const headerMatch = line.match(/^\s*\[([^\]]+)\]/);
+      if (headerMatch) {
+        inRouting = headerMatch[1].trim() === "routing";
+        continue;
+      }
+      if (inRouting && /^\s*max_agent_hops\s*=/.test(line)) {
+        routingKeyPresent = true;
+        break;
+      }
+    }
+    if (routingKeyPresent) continue;
     let next;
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
LGTM. Covers all three cases cleanly: key already present (no-op), section exists with key missing (insert under header via anchored ^ regex so substrings inside string values can't false-match), no section (append fresh). The `if (next === content) continue` guard preserves write-skip-on-no-op idempotence. Single-file scope matches the issue.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The partial-`[routing]` insertion path is close, but the new "already set" guard is file-global rather than `[routing]`-scoped. That still skips valid migrations in a real edge case, so issue #298 is not fully closed yet.

### Findings
- [medium] `max_agent_hops` detection is not scoped to the `[routing]` section
  - File: `bin/quadwork.js:1519`
  - Suggestion: only treat the setting as already present when `[routing].max_agent_hops` exists. Right now any other table with a `max_agent_hops` key causes an early `continue`, so a config like `[other]
max_agent_hops = 7

[routing]
default = "none"` still leaves `[routing]` unpatched. Parsing TOML first, or at least constraining the search to the `[routing]` block, would fix it.

### Decision
Requesting changes because the migration still misses a partial-`[routing]` case whenever the same key name appears elsewhere in the file.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new section-scoped check fixes the original false skip, but the regex now uses a JavaScript-unsupported `\z` anchor. That breaks the no-op detection for `[routing]` sections at EOF, so the migration can still duplicate `max_agent_hops`.

### Findings
- [medium] `[routing]` section slice fails at EOF because `\z` is not a JavaScript end-of-string anchor
  - File: `bin/quadwork.js:1525`
  - Suggestion: use a JS-supported EOF strategy when slicing the `[routing]` body, for example detect the next section header explicitly and otherwise fall back to the rest of the string. As written, `content.match(/...(?=^\s*\[|\z)/m)` returns `null` for a `[routing]` section at EOF, so a config like `[routing]
default = "none"
max_agent_hops = 30
` is no longer recognized as already migrated and will get a duplicate insert.

### Decision
Requesting changes because the updated fix still breaks the idempotent "already set" case for common configs where `[routing]` is the last section in the file.

project7-interns (changes requested):
High-severity regression in `bin/quadwork.js:1521`. The new section-slice regex is

```js
/^\s*\[routing\][ \t]*\r?\n([\s\S]*?)(?=^\s*\[|\z)/m
```

but **JavaScript regex does not support `\z`** (end-of-input anchor). Unknown escapes in JS regex fall through to the literal character, so the alternation in the lookahead becomes `^\s*\[` OR literal `z`. Two consequences:

1. **Files containing the letter `z` anywhere in the [routing] body terminate the slice early.** Example:

   ```
   [routing]
   default = "lazy"
   max_agent_hops = 30
   ```

   The non-greedy `*?` advances until the lookahead matches, hits the `z` in `"lazy"` first, and captures only `default = "la`. The downstream `^\s*max_agent_hops\s*=` test against that slice fails, the migration falls through to Case 2, and the key gets inserted a second time — leaving the file with two `max_agent_hops` entries.

2. **Files with no `[section]` after `[routing]` and no `z` character return null from the match,** so `routingBody` is `null`, the early-continue is skipped, and a properly-configured file (`[routing]\ndefault="none"\nmax_agent_hops = 30\n` with nothing after) is treated as if Case 1 didn't apply. Same duplicate-insert outcome on every subsequent `quadwork start`.

This breaks the explicit idempotence guarantee from the docstring and is exactly the kind of silent corruption #298 was meant to prevent.

Fix: replace `\z` with a real end-of-input pattern. Two clean options:

- `(?=^\s*\[|$(?![\s\S]))` — `$` with negative lookahead for any remaining char anchors to true EOF under `/m`.
- Drop the regex slice entirely and split the file on lines, scan from the line after the `[routing]` header until the next line matching `^\s*\[`.

The second option is more obvious to read and won't accidentally match a stray character class in a future tweak. Will re-approve once the slice handles real files. Posted on PR.

project7-interns (approved):
Fix verified on commit 5c6702c. Line-by-line scan replaces the broken regex slice — walks lines, flips inRouting on header transitions (parsing the table name with  and trimming so  would also work), short-circuits when an in-routing line matches . No regex EOF gymnastics, no string-content false matches; "lazy" no longer fools anything. Both flagged regressions are closed, and the seven-fixture verification covers the corner cases. LGTM.

project7-interns (approved):
Fix verified on commit 5c6702c. Line-by-line scan replaces the broken regex slice — walks lines, flips inRouting on header transitions, short-circuits when an in-routing line matches max_agent_hops. No regex EOF gymnastics, no string-content false matches; "lazy" no longer fools anything. Both flagged regressions are closed, and the seven-fixture verification covers the corner cases. LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The line-scan fix resolves the earlier EOF problem, but the insertion step still misses a valid partial `[routing]` section form: a section header with an inline TOML comment. That means the migration is not yet correct for all accepted partial-section cases.

### Findings
- [medium] Inline-comment `[routing]` headers are detected but not patchable
  - File: `bin/quadwork.js:1551`
  - Suggestion: make the insertion pattern accept legal TOML comments after the section header, or switch the write path to the same line-based scan used for detection. Right now `if (/^\s*\[routing\]/m.test(content))` recognizes a header like `[routing] # keep me`, but `content.replace(/^(\s*\[routing\][ \t]*\r?\n)/m, ...)` does not match it. The result is a silent no-op for a valid partial section such as `[routing] # keep me
default = "none"
`, even though acceptance explicitly requires partial `[routing]` sections to get `max_agent_hops` inserted while preserving comments.

### Decision
Requesting changes because the current implementation still skips a real partial-`[routing]` configuration shape instead of patching it idempotently.

project7-interns (approved):
Fix verified on commit 2cfe084. Insertion regex loosened to capture anything-not-newline after the [routing] bracket, then re-emitted via $1 before the new key, so trailing inline comments like "[routing] # keep me" are preserved exactly. Case 1 line-scan and Case 2 detection both already worked for this header shape; the bug was localized to the replace pattern. LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The inline-comment fix closes the previously flagged cases, but the insertion path still assumes the `[routing]` header line is newline-terminated. A valid empty `[routing]` table at EOF without a trailing newline still gets skipped, so the migration is not yet correct for all partial-section inputs.

### Findings
- [medium] Empty `[routing]` section at EOF is detected but not patchable
  - File: `bin/quadwork.js:1551`
  - Suggestion: make the insertion path handle a `[routing]` header that ends at either newline or EOF, or reuse the line-scan indices to splice after the header line directly. As written, `if (/^\s*\[routing\]/m.test(content))` recognizes a file containing just `[routing]`, but `content.replace(/^(\s*\[routing\][^\r\n]*\r?\n)/m, ...)` does not match because there is no trailing newline. I verified the migration leaves both `[routing]` and `[routing] # keep me` unchanged at EOF, even though they are valid partial `[routing]` sections that should receive `max_agent_hops = 30`.

### Decision
Requesting changes because the current implementation still skips a real partial-`[routing]` configuration shape instead of patching it idempotently.

project7-interns (approved):
Fix verified on commit 9c89454. Line-break group is now captured separately as (\r?\n|$) with a callback that synthesizes \n when the header sits at EOF, so [routing] / [routing] # keep me with no trailing newline both insert correctly while files with real newlines preserve their original line ending. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The latest fix closes the remaining partial-`[routing]` migration gaps without reintroducing the earlier idempotence regressions. The line-scan detection is now paired with an insertion path that handles newline-terminated headers, inline-comment headers, and EOF-only headers consistently.

### Findings
- No blocking findings.

### Decision
Approving because commit `9c89454` now covers the previously flagged edge cases: same-named keys in other sections no longer short-circuit, already-migrated EOF sections stay no-op, inline-comment `[routing]` headers patch correctly, and EOF-only `[routing]` headers now receive `max_agent_hops = 30`.

Issue #317, opened 2026-04-08
[master] Batch 32 — Critical bug fixes + investigations + new features (14 tickets)


realproject7:
## Master Ticket — Batch 32

A coordinated batch combining 5 bug fixes (Phase 1), 1 loop-guard recovery feature (Phase 2), 2 investigations (Phase 3), 2 small UI polish (Phase 4), 3 medium features (Phase 5), and 1 large feature (Phase 6).

### Phase 1 — Critical bug fixes (do these FIRST)
- [ ] **#309** — Loop Guard widget doesn't apply to running AC (live update broken — operator's loop guard workaround failed)
- [ ] **#306** — Scheduled Trigger sends immediately, should wait for first interval
- [ ] **#308** — Number inputs snap back to default when cleared (5 affected fields)
- [ ] **#307** — Home Recent Activity actor column too narrow (reviewer1/reviewer2 overlap)
- [ ] **#305** — Activity indicator polish (color agent name + shimmer when active)

### Phase 2 — Workflow recovery (depends on Phase 1)
- [ ] **#310** — Auto-continue after loop guard pause (depends on #309 — no point if widget is broken)

### Phase 3 — Investigations (run early, results feed into other work)
- [ ] **#303** — AgentChattr chat history persistence investigation (check AC1/AC2 patterns)
- [ ] **#304** — Auto-snapshot before AC restart (depends on #303 outcome)

### Phase 4 — Small UI polish (parallelizable)
- [ ] **#311** — SYSTEM section split (Keep Awake / Notification Sound subsections)
- [ ] **#315** — Terminal font 12→11px + line-height 1.4→1.2 to match chat density

### Phase 5 — Medium features
- [ ] **#313** — README polish (top header banner + comprehensive structure)
- [ ] **#314** — GitHub Releases backfill + automation
- [ ] **#316** — Batch Progress panel keep merged items visible until next batch

### Phase 6 — Large feature (do last)
- [ ] **#312** — Top header work hours stat + home page hero stat block

## Why this order

1. **Phase 1 first** — These are real bugs the operator has experienced. The loop guard widget bug (#309) is the highest-impact one because it's why the operator's earlier dropcast batch kept hanging.
2. **Phase 2** depends on Phase 1 — auto-continue (#310) needs the loop guard widget (#309) to actually work first.
[...]

PR #318, opened 2026-04-08
[#309] Loop Guard widget: unpause router after update_settings

### LINKED ISSUE #309: [bug] Loop Guard widget doesn't apply to running AC — settings.json updated but router.max_hops stays at 4
## Problem

PR #286 (issue #274) added a Loop Guard widget that lets the operator change \`max_agent_hops\` from the dashboard. The PR description claimed it does a \"live push to AC\" so no restart is needed.

In practice, the widget **writes \`data/settings.json\`** but does NOT update AC's in-memory \`router.max_hops\`. Result: the loop guard keeps firing at the OLD value (often 4) until the AC process is manually restarted.

## Repro from real session

Operator's dropcast batch hit the loop guard at hops 14:34 and 14:45 even though:
- \`config.toml\` correctly has \`[routing] max_agent_hops = 30\` (from migration #298)
- \`data/settings.json\` correctly has \`max_agent_hops: 50\` (from operator using the widget)
- But the AC process started at 14:24, BEFORE \`settings.json\` was updated, so \`router.max_hops\` got initialized at AC's bundled default (4) and stayed there

The settings.json file mtime is 14:48, AFTER the loop guards already fired.

## Root cause

Looking at AC source (\`~/.quadwork/dropcast/agentchattr/app.py\`):

\`\`\`python
# Line 274: read static config at startup
max_hops = cfg.get(\"routing\", {}).get(\"max_agent_hops\", 4)

# Line 305-310: settings.json overrides AT startup (one-time)
_load_settings()
if \"max_agent_hops\" in room_settings:
    router.max_hops = room_settings[\"max_agent_hops\"]
\`\`\`

After startup, the only way to update \`router.max_hops\` from outside the process is via AC's settings update endpoint (likely \`PATCH /api/room-settings\` or a ws \`update_settings\` event). PR #286 added a \`sendWsEvent\` helper for exactly this purpose, but apparently it's not landing in AC's runtime.

## Investigation

Check \`server/index.js\` \`PUT /api/loop-guard\` (added in PR #286). Verify:

1. **The ws event payload matches AC's expected shape** — search AC's app.py for the ws message handler that updates room_settings. It might be \`{type: \"update_settings\", data: {...}}\` or it might require a different event name.
2. **The ws event is actually being sent** — add a debug log
3. **AC's settings persistence path** — does writing settings.json directly cause AC to reload? Probably not (no fs.watch), so QuadWork can't just write the file and expect it to apply

## Likely fix

PR #286 wrote settings.json directly via QuadWork (sidestepping AC's API). The correct approach is:

- Use AC's actual settings API (probably \`POST /api/settings\` or similar) — this both updates the running router.max_hops AND persists settings.json
- OR send a ws event in the format AC actually accepts

Read AC's source for the canonical update endpoint and use it.

## Investigation tasks

1. Read \`/Users/cho/.quadwork/dropcast/agentchattr/app.py\` for the ws message handler that updates \`room_settings\` — find the exact event type/shape it accepts
2. Check if AC has a HTTP API endpoint for settings updates (e.g., \`PATCH /api/room-settings\`) — if yes, use that instead of the ws event
3. **Check AC1 and AC2** for how their wrapper/UI updates room_settings — look at \`/Users/cho/Projects/agentchattr/static/chat.js\` for the settings update flow
4. Update PR #286's \`PUT /api/loop-guard\` handler to use the canonical AC mechanism

## Acceptance

- ✅ Updating loop guard via the widget changes \`router.max_hops\` in the **running** AC process
- ✅ The change is visible by triggering a deeper-than-old-limit chain and confirming the new limit fires (or doesn't)
- ✅ Settings persist across AC restart
- ✅ No need to manually restart AC after changing the value

## Files

- \`server/index.js\` — \`PUT /api/loop-guard\` handler from PR #286
- \`/Users/cho/.quadwork/dropcast/agentchattr/app.py\` — read for the canonical update path
- \`/Users/cho/Projects/agentchattr/static/chat.js\` — read for AC1's settings update flow

Branch: \`task/309-loop-guard-live-update\`

realproject7:
Fixes #309
Tracker: realproject7/agent-os#417
Master: #317 (Batch 32 Phase 1 — highest priority)

## Summary
After investigation against `~/.quadwork/dropcast/agentchattr/`, the `update_settings` ws event *does* update `router.max_hops` in the running AC (`app.py:1249`) and `_save_settings()` writes `data/settings.json` — so the widget wasn't silently writing to disk. The real bug is that AC's router pauses a channel once `hop_count > max_hops` (`router.py:76-77`), and raising `max_hops` at runtime does NOT resurrect an already-paused channel. Only `router.continue_routing()` does.

The operator typically hits the widget precisely BECAUSE the channel is stuck paused, so the previous code looked like a no-op even though the live update was landing.

Fix: after the `update_settings` ws event, send a second ws message `{type: "message", text: "/continue", sender: "user"}`. AC's ws handler at `app.py:1106-1110` routes that to `router.continue_routing()` which unpauses all channels and resets `hop_count`. No-op when the channel is already running.

Both ws calls are wrapped in a shared `ensureLive()` helper so the stale-token (401) recovery path covers them together instead of leaving the widget half-applied.

## Acceptance criteria
- [x] Updating loop guard via the widget changes `router.max_hops` in the running AC (unchanged — already worked)
- [x] Channel is unpaused after raising the limit (new — via /continue follow-up)
- [x] Settings persist across AC restart (unchanged — _save_settings path)
- [x] No manual AC restart needed

## Test plan
- [ ] Reviewers: reproduce the dropcast scenario — run a batch until loop guard trips, hit the widget with a higher value, confirm the channel resumes without `/continue`

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -267,6 +267,18 @@ router.put("/api/loop-guard", async (req, res) => {
   if (!tomlPath || !fs.existsSync(tomlPath)) {
     return res.status(404).json({ error: "config.toml not found for project" });
   }
+  // Capture the previous value before rewriting so we can decide
+  // whether the /continue auto-resume should fire (only when the
+  // operator is RAISING the limit — lowering it means they want
+  // the runaway loop to stay paused).
+  let previousValue = null;
+  try {
+    const previousContent = fs.readFileSync(tomlPath, "utf-8");
+    const prevMatch = previousContent.match(/^\s*max_agent_hops\s*=\s*(\d+)/m);
+    if (prevMatch) previousValue = parseInt(prevMatch[1], 10);
+  } catch {
+    // fall through — previousValue stays null, auto-resume will skip
+  }
   try {
     let content = fs.readFileSync(tomlPath, "utf-8");
     if (/^\s*max_agent_hops\s*=/m.test(content)) {
@@ -303,13 +315,22 @@ router.put("/api/loop-guard", async (req, res) => {
   // whole fix: the previous version updated max_hops live but left
   // the channel frozen, which made the widget look like a no-op.
   let live = false;
+  // Only auto-resume when the operator is RAISING the limit. Lowering
+  // the value is a "make the system stricter" gesture — if the router
+  // is already paused on a runaway loop, we must leave it paused so
+  // the operator can investigate, otherwise a stricter limit would
+  // paradoxically unstick the runaway. Unknown previous (null) =
+  // don't auto-resume: we can't prove this is a raise.
+  const shouldAutoResume = previousValue !== null && value > previousValue;
   const ensureLive = async (sessionToken) => {
     await sendWsEvent(base, sessionToken, { type: "update_settings", data: { max_agent_hops: value } });
-    // Resume any paused channels. /continue is routed by AC's ws
-    // message handler when the buffer starts with /continue; the
-    // handler calls router.continue_routing() which unpauses and
-    // resets hop_count. No-op when the channel is already running.
-    await sendWsEvent(base, sessionToken, { type: "message", text: "/continue", channel: "general", sender: "user" });
+    if (shouldAutoResume) {
+      // Resume any paused channels. /continue is routed by AC's ws
+      // message handler when the buffer starts with /continue; the
+      // handler calls router.continue_routing() which unpauses and
+      // resets hop_count. No-op when the channel is already running.
+      await sendWsEvent(base, sessionToken, { type: "message", text: "/continue", channel: "general", sender: "user" });
+    }
     live = true;
   };
   let base = null;
@@ -342,7 +363,7 @@ router.put("/api/loop-guard", async (req, res) => {
     console.warn(`[loop-guard] live update failed for ${projectId}: ${err.message || err}`);
   }
 
-  res.json({ ok: true, value, live });
+  res.json({ ok: true, value, live, previousValue, resumed: shouldAutoResume && live });
 });
 
 // #412 / quadwork#279: project history export + import.
--- a/server/routes.js
+++ b/server/routes.js
@@ -334,8 +334,14 @@ router.put("/api/loop-guard", async (req, res) => {
       // exposes `paused: true` iff ANY channel currently paused.
       let isPaused = false;
       try {
-        const statusRes = await fetch(`${base}/api/status`, {
-          headers: sessionToken ? { Authorization: `Bearer ${sessionToken}` } : {},
+        // AC's security middleware (app.py:212-224) only accepts
+        // bearer auth for /api/messages, /api/send, and /api/rules/*.
+        // /api/status requires x-session-token header (or ?token=),
+        // so pass that instead — a bearer header silently 403s and
+        // leaves isPaused stuck at false, defeating the gate.
+        const statusUrl = `${base}/api/status`;
+        const statusRes = await fetch(statusUrl, {
+          headers: sessionToken ? { "x-session-token": sessionToken } : {},
           signal: AbortSignal.timeout(5000),
         });
         if (statusRes.ok) {
--- a/server/routes.js
+++ b/server/routes.js
@@ -289,14 +289,37 @@ router.put("/api/loop-guard", async (req, res) => {
   // /api/chat does (#230): re-sync the session token from AC and
   // retry once. Other failures stay non-fatal — the persisted value
   // still takes effect on next AC restart.
+  //
+  // #417 / quadwork#309: the update_settings ws event correctly
+  // updates router.max_hops in the running AC (verified in AC's
+  // app.py:1249), AND writes settings.json via _save_settings. But
+  // AC's router stays paused once it has tripped the guard — raising
+  // max_hops at runtime does NOT resurrect an already-paused channel
+  // (router.py:76-77 → `paused = True`). The operator typically
+  // raises the limit precisely BECAUSE the channel is stuck paused,
+  // so we immediately follow the update_settings event with a
+  // `/continue` chat message (the same path AC's own slash command
+  // handler uses at app.py:1106-1110) to resume routing. This is the
+  // whole fix: the previous version updated max_hops live but left
+  // the channel frozen, which made the widget look like a no-op.
   let live = false;
+  const ensureLive = async (sessionToken) => {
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
LGTM. Good investigation — verified against AC's app.py:1249 + router.py:76-77 + 1106-1110 references. The fix correctly addresses the actual root cause: update_settings did update router.max_hops, but a paused router stays paused until something calls continue_routing(), and the operator typically raises the limit precisely because they want to resume.

- ensureLive() helper wraps both calls so the 401 token-resync path covers them atomically. live=true is only set after both succeed, so a partial failure doesn't lie about live state.
- /continue routed through the same ws message path AC's own slash handler uses, so this matches what the operator typing /continue manually would do.
- "No-op when channel is already running" relies on continue_routing() being idempotent — trust based on the AC source citation.

Non-blocking observation: the /continue is hardcoded to channel: "general". If the operator was running a paused batch on a custom channel, the resume won't reach it. Worth a follow-up to either pass the channel through the PUT body or fan out to all known channels — but matches the rest of QuadWork's general-channel-only assumption for now.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The dropcast repro is real, but the current fix makes `/api/loop-guard` resume routing on every value change, not just on the "raise the limit to unstick a paused run" case from issue #309. That changes the widget semantics in a way that can defeat the safety purpose of lowering the limit.

### Findings
- [medium] Loop-guard updates now perform an unconditional hidden `/continue`
  - File: `server/routes.js:312`
  - Suggestion: gate the `/continue` follow-up behind the specific case that actually needs it, for example only when the new value is higher than the previous one and/or when AC is currently paused. As written, every PUT to `/api/loop-guard` sends `/continue`, which unpauses routing and resets `hop_count` even when the operator is lowering the limit to make the system stricter. A paused runaway loop can therefore be resumed by trying to set a *lower* guard value, which is the opposite of the widget’s documented "lower values add safety" behavior.

### Decision
Requesting changes because the fix solves the "raise and resume" case by introducing a broader behavior change that resumes loops for unrelated updates too.

project7-interns (approved):
Fix verified on commit 999a2a4. previousValue captured from config.toml before the rewrite, shouldAutoResume = (previousValue !== null && value > previousValue) so:

- Raising the limit fires /continue (the issue #309 case).
- Lowering / equal leaves the router paused state alone — a stricter gesture can't accidentally unstick a runaway loop.
- Unknown previous (null on missing key) skips auto-resume because we can't prove it's a raise.

Response surfaces previousValue + resumed so the widget can show honest feedback. Good defensive fix.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new `value > previousValue` gate fixes the "lowering shouldn't unstick a runaway" regression, but the route still auto-sends `/continue` on every raise even when the channel is not paused. In AgentChattr, `/continue` is not a no-op there: it resets hop_count and emits a system message, so ordinary proactive raises now change live routing state in a way issue #309 did not ask for.

### Findings
- [medium] Raising the guard mid-run still resets hop count even when no loop-guard pause exists
  - File: `server/routes.js:324`
  - Suggestion: gate `/continue` on actual paused state, not just on `value > previousValue`. The current code assumes `/continue` is a no-op on a running channel, but AC's handler at `app.py:1106-1108` always calls `router.continue_routing()`, and `router.py:84-86` resets `hop_count = 0` / `paused = False` / `guard_emitted = False`. That means a proactive raise from 30→40 during an active conversation silently resets the current chain length and extends the run beyond the newly chosen limit.

### Decision
Requesting changes because the fix still applies resume semantics to normal "raise the limit" edits, not just the paused-case recovery that issue #309 is targeting.

project7-interns (approved):
Fix verified on commit 9b68384. Three-way gate now correctly addresses both blockers:

(a) raising the limit (previous commit's check)
(b) previousValue known
(c) router actually paused right now — queried via GET /api/status with bearer auth and AbortSignal.timeout(5000)

A proactive 30→40 mid-run no longer fires /continue and therefore doesn't reset hop_count. Status fetch failures fall back to not auto-resuming (operator can type /continue manually). autoResumed tracked separately so the response surfaces what actually happened. LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new paused-state gate fixes the previous hop-count-reset regression in principle, but the `/api/status` probe is authenticated incorrectly. In AgentChattr, bearer auth is not accepted on `/api/status`, so this check falls back to "not paused" and the auto-resume branch never fires. That leaves the original issue #309 bug unresolved.

### Findings
- [high] Paused-state probe uses auth that `/api/status` rejects, so auto-resume silently never triggers
  - File: `server/routes.js:337`
  - Suggestion: call `/api/status` with the session token in the form AgentChattr actually accepts for that route (`x-session-token` or `?token=`), not `Authorization: Bearer`. AC's security middleware in `app.py:212-224` only allows bearer auth for `/api/messages`, `/api/send`, and `/api/rules/*`; everything else, including `/api/status`, requires `x-session-token` or query `token`. As written, the fetch at `server/routes.js:337-339` will 403, `isPaused` stays false, and the new `/continue` branch never runs even when the router is genuinely paused.

### Decision
Requesting changes because the current fix still does not satisfy issue #309's core acceptance case: raising the limit on a paused run should resume the running AC without manual `/continue`.

project7-interns (approved):
Fix verified on commit b5315a3. /api/status probe now uses x-session-token header per AC's security middleware. Comment cites the constraint with the app.py:212-224 reference. Probe will actually return 200 + paused state now, so the auto-resume gate fires correctly when the router is paused. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The latest fix makes the paused-state gate operational without reintroducing the earlier regressions. Auto-resume is now limited to genuine "raise the limit while AC is actually paused" cases, and the status probe uses the session-token auth that AgentChattr accepts on `/api/status`.

### Findings
- No blocking findings.

### Decision
Approving because commit `b5315a3` closes the remaining bug: the `/api/status` probe now uses `x-session-token`, so the paused check can actually succeed and the `/continue` follow-up can fire only when the router is truly paused. Lowering the limit still leaves a runaway paused, and proactive raises on a running channel no longer reset hop count.

PR #319, opened 2026-04-08
[#306] Scheduled Trigger: wait for first interval

### LINKED ISSUE #306: [fix] Scheduled Trigger — don't send immediately, wait for first interval
## Problem

Clicking \"Send Message and Start Trigger\" in the Scheduled Trigger widget fires the message **immediately** AND starts the timer. This is annoying:

- The operator just types the trigger message and clicks Start to **schedule** it for autonomous overnight runs
- They don't want an immediate ping — the agents are usually mid-task and the immediate fire interrupts whatever they're doing
- The mental model is \"set up a recurring reminder\" not \"send now and then keep sending\"

The current behavior comes from the \`sendImmediately: true\` flag in PR #210 (issue #220). Operators have asked to remove the immediate send.

## Fix

Change the trigger to **always** wait for the first interval before sending:

1. Click \"Start Trigger\" → no immediate send
2. Wait \`interval\` minutes (e.g., 15)
3. First send fires at \`T + interval\`
4. Then keeps sending every \`interval\` minutes until duration expires

The countdown UI already shows \"Next message in 14m 32s\" — this just makes that countdown the actual schedule from the very first message.

### Backend

\`server/index.js\` \`POST /api/triggers/:project/start\` — remove the \`sendImmediately\` immediate-send block:

\`\`\`javascript
// REMOVE THIS BLOCK:
if (sendImmediately) {
  sendTriggerMessage(project).catch(() => {});
}
\`\`\`

The \`setInterval\` already schedules the recurring send. Without the immediate fire, the first send naturally happens at \`T + interval\`.

The \`info.nextAt\` should already be initialized to \`Date.now() + interval\` so the countdown displays correctly.

### Frontend

\`src/components/ScheduledTriggerWidget.tsx\`:

1. Rename the button from \`Send Message and Start Trigger\` → **\`Start Trigger\`**
2. Drop the \`sendImmediately: true\` from the body
3. Update any helper text that implies immediate send (e.g., \"sends now and then every N minutes\")

## Acceptance

- ✅ Clicking Start does NOT send a message immediately
- ✅ First message arrives at T + interval (e.g., 15 minutes later if interval = 15)
- ✅ Subsequent messages arrive every \`interval\` minutes
- ✅ Auto-stops after duration expires (unchanged)
- ✅ Stop button works (unchanged)
- ✅ Button label updated to \"Start Trigger\"
- ✅ Helper text doesn't claim the trigger sends immediately

## Files

- \`server/index.js\` — \`/api/triggers/:project/start\` handler
- \`src/components/ScheduledTriggerWidget.tsx\` — button label + post body
- (Optional) \`server/index.js\` — drop the \`sendImmediately\` param from the route schema if it's no longer used

Branch: \`task/306-trigger-no-immediate-send\`

realproject7:
Fixes #306
Tracker: realproject7/agent-os#418
Master: #317 (Batch 32 Phase 1)

## Summary
- `server/index.js` /api/triggers/:project/start drops the `sendImmediately` one-shot branch. `setInterval` already schedules the first send at T + interval — no change needed beyond the removal.
- `src/components/ScheduledTriggerWidget.tsx` button text `"Send Message and Start Trigger"` → `"Start Trigger"`, body no longer sends `sendImmediately: true`, doc comment updated.
- The explicit one-shot path `/api/triggers/:project/send-now` is untouched — still available for the rare case an operator wants to kick things off manually.

## Acceptance criteria
- [x] Start does NOT send immediately
- [x] First message arrives at T + interval (setInterval contract)
- [x] Subsequent messages unchanged
- [x] Auto-stop / Stop button unchanged
- [x] Button label updated
- [x] Helper / doc text updated

## Test plan
- [x] `tsc --noEmit` clean
- [ ] Reviewers: set interval to 1 min, click Start, confirm no message at T0 and a message at T+60s

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1086,7 +1086,12 @@ function stopTrigger(project) {
 
 app.post("/api/triggers/:project/start", (req, res) => {
   const { project } = req.params;
-  const { interval, duration, message, sendImmediately } = req.body || {};
+  // #418 / quadwork#306: sendImmediately was an always-true
+  // "Send Message and Start Trigger" flag from #210; operators
+  // asked for a pure scheduler ("Start Trigger" — wait for the
+  // first interval). The field is ignored here; the send-now
+  // endpoint below still exists for the explicit one-shot path.
+  const { interval, duration, message } = req.body || {};
   const ms = (interval || 30) * 60 * 1000;
   const durationMs = duration ? duration * 60 * 1000 : 0; // duration in minutes, 0 = indefinite
 
@@ -1113,16 +1118,12 @@ app.post("/api/triggers/:project/start", (req, res) => {
     if (existing.durationTimer) clearTimeout(existing.durationTimer);
   }
 
-  // #210: the Scheduled Trigger widget's "Send Message and Start
-  // Trigger" button expects an immediate send, not the first fire
-  // one interval in the future. setInterval won't do that on its
-  // own, so trigger a one-shot send when sendImmediately is true.
-  if (sendImmediately) {
-    // Don't await — keep the response fast. sendTriggerMessage logs
-    // its own errors and updates lastError on the trigger info.
-    sendTriggerMessage(project).catch(() => {});
-  }
-
+  // #418 / quadwork#306: no immediate fire — the first send happens
+  // at T + interval via the setInterval below. Operators set the
+  // trigger up in advance of going afk and don't want it interrupting
+  // whatever agents are currently mid-task. The explicit "send now"
+  // path still lives at /api/triggers/:project/send-now for the
+  // rare case an operator actually wants to kick things off.
   const timer = setInterval(() => sendTriggerMessage(project), ms);
   const expiresAt = durationMs > 0 ? Date.now() + durationMs : null;
 
--- a/src/components/ScheduledTriggerWidget.tsx
+++ b/src/components/ScheduledTriggerWidget.tsx
@@ -64,11 +64,12 @@ function formatCountdown(ms: number): string {
  * Bottom-right operator widget for the Scheduled Trigger (#210).
  *
  * Combines the old Keep Alive timer with a custom message textarea.
- * "Start Trigger" sends the typed message via the
- * existing /api/triggers/:id/start endpoint (which persists the
- * message on the project entry and immediately fires once), then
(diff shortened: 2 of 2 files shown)


project7-interns (approved):
LGTM. Drops the sendImmediately branch from /api/triggers/:project/start so the first send happens at T + interval, renames the button, drops the field from the POST body. send-now endpoint left intact for the explicit one-shot path.

Non-blocking doc nit: the widget's top-of-file docstring still says "...immediately fires once, then the backend's setInterval keeps firing at the configured cadence". Worth a one-line correction in a follow-up so the comment matches the new behavior.

project7-interns (approved):
Re-approve on commit 00cae96 — docstring updated. Pure doc change.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The runtime behavior looks right: the immediate send is gone, `nextAt` still starts at `T + interval`, and the explicit send-now path is untouched. But the PR still misses its own helper/doc-text acceptance because `ScheduledTriggerWidget`'s doc block continues to describe an immediate first fire.

### Findings
- [low] Scheduled Trigger doc comment still claims the first message fires immediately
  - File: `src/components/ScheduledTriggerWidget.tsx:67`
  - Suggestion: update the widget doc block to match the new behavior. In the current PR diff it still says `/api/triggers/:id/start` "persists the message on the project entry and immediately fires once", which is now false after removing `sendImmediately`. Since issue #306 explicitly calls for helper/doc text to stop implying an immediate send, this should be corrected before merge.

### Decision
Requesting changes because the code path is fixed but the documented behavior in the same PR is still out of date on the current head.

project7-interns (approved):
## Verdict: APPROVE

### Summary
Current head `00cae96` closes the only remaining acceptance miss by updating the Scheduled Trigger doc block to match the new first-fire-at-`T + interval` behavior. The runtime change still looks correct: no immediate send on start, `nextAt` still initializes to `Date.now() + interval`, and the explicit send-now path remains intact.

### Findings
- No blocking findings.

### Decision
Approving because the current head now satisfies both the behavior change and the helper/doc-text update required by issue #306.

PR #320, opened 2026-04-08
[#308] Number inputs: draft-string pattern for 5 fields

realproject7:
Fixes #308
Tracker: realproject7/agent-os#419
Master: #317 (Batch 32 Phase 1)

## Summary
Migrate 5 number inputs from the broken `parseInt(value) || default` anti-pattern to the draft-string pattern PR #289 introduced. Each field gains a string draft state (onChange stores raw text, onBlur parses+clamps+commits), keeping the numeric state in sync.

Fields:
1. `ScheduledTriggerWidget` — interval
2. `SettingsPage` — dashboard port (shared `Input` helper extended with optional `onBlur`)
3. `SetupWizard` — AgentChattr / MCP HTTP / MCP SSE ports (single `commitPortDraft` helper, detected-port hydration updates both state + draft)

Submit handlers (`start()` in the trigger widget) commit the draft locally before POSTing to avoid the setState render race.

Audit grep `parseInt.*||.*[0-9]` returns zero real instances in `src/components/` — only the new comment references describing the pattern.

## Acceptance criteria
- [x] All 5 fields clear-and-retype freely
- [x] No snap-back during typing
- [x] Validation/clamp on blur or submit
- [x] Default falls in only when committed empty
- [x] Submit handlers commit drafts before posting
- [x] Audit grep clean

## Test plan
- [x] `tsc --noEmit` clean
- [ ] Reviewers: each field — backspace to empty, retype, confirm the value sticks; click Start without blurring to confirm submit-side commit

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #323, opened 2026-04-08
[#310] Loop guard auto-continue (opt-in per project)

### LINKED ISSUE #310: [feat] Auto-continue after loop guard fires (operator opt-in)
## Background

When AgentChattr's loop guard fires after \`max_agent_hops\` consecutive agent-to-agent messages, it pauses routing and emits:

> Loop guard: N agent-to-agent hops reached. Type \`/continue\` to resume.

The operator must manually type \`/continue\` in the chat for routing to resume. For autonomous overnight runs (the whole point of QuadWork), this is brutal — every time the loop guard fires, the entire batch stalls until the operator wakes up.

PR #286 (issue #274) raised the default to 30 hops which helps a lot, but the loop guard can still fire on long PR cycles (e.g., a flaky CI requiring multiple re-reviews). When it does, the batch hangs.

## Proposed fix

Add an opt-in **auto-continue** mode in the Loop Guard widget. When enabled:

1. After the loop guard fires, QuadWork automatically waits N seconds (e.g., 30s)
2. Then sends \`/continue\` to the chat as if the operator typed it
3. Routing resumes

The wait is critical — without a delay, an actual runaway loop would just continue forever. The 30s delay gives the operator a chance to manually intervene if they're watching, and acts as a soft-rate-limit against pathological loops.

## UI

In the Loop Guard widget added by PR #286, add a checkbox:

\`\`\`
LOOP GUARD (?)
Pause after [30] hops
[ ] Auto-continue after pause (waits 30s, then sends /continue)
\`\`\`

When checked, persist the preference per-project in config.json.

### Backend

\`server/index.js\` — listen for the loop guard pause event from AC. When detected, if auto-continue is enabled for that project:

1. Wait the configured delay (default 30s)
2. POST a \`/continue\` message to \`/api/chat\` with the operator's name
3. Log it so the operator can audit

\`\`\`javascript
async function handleLoopGuardPause(projectId) {
  const cfg = readConfig();
  const project = cfg.projects?.find((p) => p.id === projectId);
  if (!project?.auto_continue_loop_guard) return;
  
  const delay = (project.auto_continue_delay_sec || 30) * 1000;
  console.log(\`[loop-guard] Auto-continue scheduled for \${projectId} in \${delay}ms\`);
  setTimeout(async () => {
    try {
      await fetch(\`http://127.0.0.1:\${cfg.port}/api/chat?project=\${projectId}\`, {
        method: \"POST\",
        headers: { \"Content-Type\": \"application/json\" },
        body: JSON.stringify({ text: \"/continue\", channel: \"general\" }),
      });
      console.log(\`[loop-guard] Auto-continued \${projectId}\`);
    } catch (err) {
      console.error(\`[loop-guard] Auto-continue failed for \${projectId}: \${err.message}\`);
    }
  }, delay);
}
\`\`\`

How to detect the pause: subscribe to AC's ws event stream (the same path the chat panel uses), filter for messages with \`type === \"system\"\` and \`text\` containing \`\"Loop guard:\"\`. When seen, fire \`handleLoopGuardPause\`.

### Frontend

Add the checkbox to \`LoopGuardWidget.tsx\`. Bind to \`config.projects[N].auto_continue_loop_guard\`. PUT via the existing config endpoint.

## Safety considerations

- **Default OFF** — opt-in only. Operators who don't read this ticket won't get auto-continue silently
- **Per-project** — different projects might have different appetites for autonomy
- **Delay configurable** — default 30s, allow 5-300s range
- **Logged** — every auto-continue is logged so the operator can audit later

## Acceptance

- ✅ Loop Guard widget has an \"Auto-continue after pause\" checkbox
- ✅ Default OFF
- ✅ When checked, loop guard pauses are followed by an automatic \`/continue\` after the configured delay
- ✅ Auto-continue is logged in QuadWork's server stdout
- ✅ Setting persists per-project in config.json
- ✅ Disabling the checkbox stops future auto-continues
- ✅ Manual \`/continue\` from the operator still works (no conflict)

## Files

- \`server/index.js\` — listen for loop guard pause events, fire auto-continue
- \`src/components/LoopGuardWidget.tsx\` — checkbox + delay input
- \`server/config.js\` — schema additions for \`auto_continue_loop_guard\` and \`auto_continue_delay_sec\`

## Related

- #274 — Loop guard default + UI (the widget being extended)
- #309 — Loop guard widget doesn't actually update the running AC (must be fixed first or this won't help when the operator changes the limit)

Branch: \`task/310-loop-guard-auto-continue\`

realproject7:
Fixes #310
Tracker: realproject7/agent-os#422
Master: #317 (Batch 32 Phase 2)
Depends on: quadwork#309 (merged as PR #318)

## Summary
Default-OFF per-project opt-in so overnight runs don't stall when AC's loop guard fires.

Backend (`server/index.js`):
- New `setInterval` polls `/api/status` every 10s per project with `auto_continue_loop_guard` enabled.
- `paused` false → true schedules `setTimeout` for the configured delay (5-300s, default 30s). On fire, re-reads the config so a checkbox disable mid-wait actually cancels, then POSTs `{text:"/continue",channel:"general"}` to `/api/chat`.
- `paused` true → false cancels any pending timer (manual operator `/continue` short-circuits the auto one).
- Every auto-continue logs to server stdout. Polling chosen over ws subscribe — 10s latency is fine given the tens-of-seconds delay, and avoids per-project ws lifecycle complications.

Frontend (`src/components/LoopGuardWidget.tsx`):
- Hydrates both prefs from `/api/config` on mount.
- New checkbox + delay input under the hops row. Saves via existing `PUT /api/config` (whole-config write) so no new endpoint.
- Delay input uses the #419 draft-string pattern and disables when auto-continue is off.

## Acceptance criteria
- [x] Widget has an "Auto-continue after pause" checkbox, default OFF
- [x] Delay configurable 5–300s (draft-string input)
- [x] Auto-continue logged to server stdout
- [x] Setting persists per-project in config.json
- [x] Disabling the checkbox mid-wait cancels a pending auto-continue
- [x] Manual /continue from the operator cancels the timer via the true→false transition
- [ ] Reviewers: trigger the loop guard, confirm auto-continue fires after the delay; also confirm disabling the checkbox during the wait stops it

## Test plan
- [x] `tsc --noEmit` clean
- [ ] Reviewers: enable auto-continue with a small delay (say 10s), force the loop guard via a tight hop config, confirm resume; then repeat with checkbox disabled mid-wait and confirm no /continue

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1391,6 +1391,94 @@ function syncTriggersFromConfig() {
   }
 }
 
+// #422 / quadwork#310: auto-continue after loop guard.
+//
+// Per opted-in project, poll AC's /api/status every 10s. When we see
+// a false → true transition on `paused`, wait the configured delay
+// (default 30s) and POST /continue to /api/chat — same path the
+// operator would use manually. The delay gives a human a chance to
+// intervene on an actually-runaway loop, and acts as a soft rate
+// limit against pathological loops that would otherwise just loop
+// forever under an auto-continue.
+//
+// Detection is deliberately polling rather than a long-lived ws:
+// a ws subscription per project would complicate lifecycle and
+// reconnection, and 10s polling latency is acceptable when the
+// delay is tens of seconds. Skipping projects without the opt-in
+// keeps the poller cheap for single-project setups.
+
+const _loopGuardPausedState = new Map(); // projectId -> { paused: bool, scheduled: Timeout? }
+const LOOP_GUARD_POLL_INTERVAL_MS = 10000;
+
+async function checkLoopGuardPause(project) {
+  if (!project || !project.auto_continue_loop_guard) return;
+  const { url: base, token: sessionToken } = resolveProjectChattr(project.id);
+  if (!base) return;
+  let paused = false;
+  try {
+    const r = await fetch(`${base}/api/status`, {
+      headers: sessionToken ? { "x-session-token": sessionToken } : {},
+      signal: AbortSignal.timeout(5000),
+    });
+    if (!r.ok) return;
+    const data = await r.json();
+    paused = !!(data && data.paused);
+  } catch {
+    return;
+  }
+  const state = _loopGuardPausedState.get(project.id) || { paused: false, scheduled: null };
+  // Transition false → true: schedule an auto-continue after the delay.
+  if (paused && !state.paused && !state.scheduled) {
+    const delaySec = Number.isFinite(project.auto_continue_delay_sec) && project.auto_continue_delay_sec >= 5
+      ? project.auto_continue_delay_sec
+      : 30;
+    console.log(`[loop-guard] ${project.id} paused — auto-continue in ${delaySec}s`);
+    state.scheduled = setTimeout(async () => {
+      try {
+        // Re-check the opt-in at fire time so a checkbox disable
+        // mid-wait actually stops the auto-continue.
+        const freshCfg = readConfig();
+        const fresh = freshCfg.projects?.find((p) => p.id === project.id);
+        if (!fresh || !fresh.auto_continue_loop_guard) {
+          console.log(`[loop-guard] ${project.id} auto-continue cancelled (opt-in disabled during wait)`);
+        } else {
+          const res = await fetch(`http://127.0.0.1:${PORT}/api/chat?project=${encodeURIComponent(project.id)}`, {
+            method: "POST",
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify({ text: "/continue", channel: "general" }),
+          });
+          if (res.ok) console.log(`[loop-guard] ${project.id} auto-continued`);
+          else console.warn(`[loop-guard] ${project.id} auto-continue POST returned ${res.status}`);
+        }
+      } catch (err) {
+        console.warn(`[loop-guard] ${project.id} auto-continue failed: ${err.message || err}`);
+      }
+      const s2 = _loopGuardPausedState.get(project.id);
+      if (s2) s2.scheduled = null;
+    }, delaySec * 1000);
+  }
+  // Transition true → false: clear any pending timer.
+  if (!paused && state.paused && state.scheduled) {
+    clearTimeout(state.scheduled);
+    state.scheduled = null;
+  }
+  state.paused = paused;
+  _loopGuardPausedState.set(project.id, state);
+}
+
+function runLoopGuardPollingTick() {
+  try {
+    const cfg = readConfig();
+    for (const p of (cfg.projects || [])) {
+      if (p && p.auto_continue_loop_guard) checkLoopGuardPause(p);
+    }
+  } catch {
+    // config unreadable — next tick will retry
+  }
+}
+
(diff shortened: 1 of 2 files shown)


project7-interns (approved):
LGTM. Solid Phase 2 implementation.

Server side:
- 10s polling loop, AbortSignal.timeout(5000) on the status fetch, x-session-token auth (matches the #318 fix). False→true transition schedules a timer; true→false cancels any pending one. The fire callback re-reads config so a mid-wait checkbox disable actually cancels.
- Default delay 30s, minimum 5s clamp prevents pathological tight loops from resuming instantly. The /continue routes through /api/chat loopback, inheriting the operator-name + token-resync handling.
- Per-project state map keyed by project id; the `!state.scheduled` guard correctly prevents re-scheduling while a timer is in flight.
- Default OFF — operators opt in knowing the runaway-loop trade-off.

Client side:
- Hydrates from /api/config on mount, persists via existing whole-config PUT (matches SettingsPage pattern, no new endpoint surface).
- Checkbox revert on save failure keeps in-memory and on-disk state in sync.
- Delay uses the draft-string blur-commit pattern from #320 with the same 5–300 clamp the server enforces.
- Disabled during save prevents double-fires.

Non-blocking observation: if the auto-continue POST itself fails, state.paused becomes true on the next tick without a new transition, so we silently give up rather than retry. Acceptable for v1 — the operator sees AC stuck and intervenes — but worth a follow-up if it becomes a real pattern.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new auto-continue path is close, but the manual-cancel behavior is still racey. A pending auto-`/continue` is only cleared on the next 10s poll, so valid short delays can still fire after the operator has already resumed routing manually.

### Findings
- [high] Manual `/continue` does not reliably cancel a pending auto-continue
  - File: `server/index.js:1411`
  - File: `server/index.js:1460`
  - Suggestion: do not rely only on the next poll tick to clear the timeout. Re-check paused state again when the timeout fires before POSTing `/continue`, or wire the timer cancellation to an immediate pause→unpause signal. As written, a 5–9s delay can expire before the next 10s poll notices the manual resume.

### Decision
Request changes because issue #310 explicitly requires that manual `/continue` still works with no conflict, and the current implementation can still post a stale `/continue` after a manual resume.

project7-interns (approved):
Fix verified on commit 18e071b. Live /api/status re-check inside the setTimeout callback (same x-session-token path as the outer poller) skips the /continue post when status.paused is false at fire time, closing the manual-resume race even at the 5s minimum delay. Re-check failures fall back to don't-fire — a stuck pause still gets caught on the next 10s tick. The defense-in-depth pattern (opt-in re-read + status re-check) covers both classes of mid-wait state changes. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The new fire-time status re-check closes the manual-resume race from the previous review. The auto-continue path now only posts `/continue` when AgentChattr is still paused at timer fire time, while preserving the default-OFF opt-in and cancel-on-disable behavior.

### Findings
- None.

### Decision
Approve because the updated implementation now satisfies issue #310’s no-conflict manual `/continue` requirement and the supporting checks remain clean.

PR #324, opened 2026-04-08
[#303] Investigation: AC chat history persistence (already works)

realproject7:
Fixes #303
Tracker: realproject7/agent-os#423
Master: #317 (Batch 32 Phase 3 — investigation)

## Finding
AgentChattr **already persists chat history** to `{data_dir}/agentchattr_log.jsonl`. No QuadWork code change or built-in config option needed. The ticket's premise is incorrect for the current AC code.

Evidence (all cited in the doc):
- `store.py:52-83` — every `store.add()` appends to disk with fsync.
- `store.py:24, 27-46` — `_load()` rehydrates on startup, preserves stable IDs.
- Dropcast's live `data/agentchattr_log.jsonl` has 116 persisted messages right now.
- `settings.json` confirms `history_limit: "all"`.

## Deliverable
New `docs/investigations/303-chat-history-persistence.md` writes up the finding with file citations, live evidence, and a short list of what to investigate next if the operator still observes chat "disappearing" (stale /api/messages cursor, iframe ws replay window, wrong data_dir from a migration).

## Implication for #304
Reframes auto-snapshot before restart from "the only way to keep history" to "defense in depth against /clear or destructive log-rewrite operations." Not a blocker for #304 proceeding, but narrows its scope.

## Acceptance criteria
- [x] Investigation covers the AC files/paths referenced upstream
- [x] Output clearly states whether built-in persistence exists — YES
- [x] Limitation / implication documented (file-level citations + reframed #304 scope)

## Test plan
Pure documentation PR, no runtime changes. Reviewers can verify the citations against `/Users/cho/.quadwork/dropcast/agentchattr/{store.py,app.py,data/agentchattr_log.jsonl,data/settings.json}`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #303, opened 2026-04-08
[bug] AgentChattr chat history lost on restart — investigate persistent storage


realproject7:
## Problem

When QuadWork restarts AgentChattr (e.g., during the v1.3.0 \`migrateLoopGuardDefaults\` migration, or via the dashboard's Restart button), the entire chat history disappears. The dashboard shows the empty state and previous conversations are gone.

## Root cause analysis

AgentChattr's data directory (\`~/.quadwork/{project}/agentchattr/data/\`) only contains:
- \`agentchattr_log.jsonl\` — general server log, NOT chat messages
- \`{agent}_queue.jsonl\` — per-agent trigger queue files (always emptied after read)
- \`mcp_cursors.json\`, \`renames.json\` — small state files

**There is no persistent chat history file.** AgentChattr stores chat messages in-memory only, so they're lost on every process restart.

### 1. Check AC1 and AC2 for the same pattern

Compare the data directories of the operator's existing AgentChattr installs:

- \`/Users/cho/Projects/agentchattr/data/\` (AC1, currently NOT running)
- \`/Users/cho/Projects/agent-os/agentchattr2/data/\` (AC2, currently running on port 8202/8203)

If AC1/AC2 show the same data layout (no SQLite, no message JSON), then chat-history-on-disk is **not enabled by default in any AgentChattr setup** — even the operator's known-working ones.

If AC2 (which is running and has been used heavily) shows persistent chat messages anywhere, find out **how it's persisting them** — maybe a config option, maybe a feature flag, maybe a separate file we missed.

### 2. Check AgentChattr source for opt-in persistence

Read \`~/.quadwork/dropcast/agentchattr/store.py\` and \`app.py\` (or wherever the message store lives) for:
- Any \`if\` branches gated on a config option like \`persist_messages\`, \`store_history\`, \`db_path\`
- Any unused SQLite or JSON-file write paths
- Any \`load_messages_from_disk\` / \`save_messages_to_disk\` functions
- A \`config.toml\` setting we're not currently writing

If a built-in option exists, the fix is just to set it in QuadWork's template — no AgentChattr code change needed.

### 3. If no built-in option exists, ticket the upstream gap separately

Don't try to add persistence to AgentChattr from QuadWork's side — that's a fork of the upstream tool. Instead, surface the gap in a separate issue and consider if a) we should fork AC (not preferred), b) we should rely on #279 export/import as a manual workaround, or c) we should auto-export before restart (covered by separate ticket).

[...]

PR #325, opened 2026-04-08
[#304] Auto-snapshot project history before AC restart

### LINKED ISSUE #304: [feat] Auto-export project history before AgentChattr restart (safety net)
## Problem

Whenever QuadWork restarts a project's AgentChattr instance — via the dashboard \"Restart AgentChattr\" button, the loop-guard migration on \`cmdStart\`, the AgentChattr update endpoint, or any other lifecycle event — **all chat history is lost** because AgentChattr stores messages in-memory.

This is brutal mid-session. The operator might restart for an unrelated reason (fix a port conflict, apply a config change) and lose their entire conversation history.

## Solution

Use the existing project history export from #279 as a safety net. Before any AgentChattr restart:

1. Call the existing \`GET /api/project-history?project={id}\` endpoint internally
2. Save the JSON response to \`~/.quadwork/{project}/history-snapshots/{timestamp}.json\`
3. Restart AgentChattr
4. (Optional) Auto-import the latest snapshot into the new AgentChattr instance after a successful restart

The auto-import step is the tricky part — it would need to use the existing \`POST /api/project-history\` endpoint with \`allow_duplicate=true\` and \`allow_agent_senders=true\` (since the snapshot will contain real agent messages). This is exactly the disaster-recovery case the #297 denylist was designed for.

### Check AC1 and AC2 for the same pattern

The operator's existing AC1 and AC2 setups have been running for weeks with frequent restarts. How do they handle chat history continuity?

- Look in \`/Users/cho/Projects/agentchattr/\` and \`/Users/cho/Projects/agent-os/agentchattr2/\` for:
  - Any backup/snapshot directories
  - Any cron/launchd job that exports history periodically
  - Any wrapper.py-level snapshotting before restart
- If they don't snapshot at all, then either: (a) the operator hasn't noticed history loss, or (b) AC1/AC2 do persist history somehow (covered by the sibling ticket #303 — investigation result feeds back here)

If #303 finds a built-in persistence option, this ticket may become unnecessary. Coordinate with #303.

### Phase 1 — Snapshot before restart (auto-export)

In all places where QuadWork restarts AgentChattr (\`server/index.js\` \`spawnChattr\` lifecycle, \`bin/quadwork.js\` \`migrateLoopGuardDefaults\`, the update endpoint), wrap the restart with:

\`\`\`javascript
async function safeRestartChattr(projectId) {
  try {
    const snapshot = await fetchProjectHistory(projectId);
    saveSnapshot(projectId, snapshot);
  } catch (err) {
    console.warn(\`[restart] failed to snapshot history for \${projectId}: \${err.message}\`);
  }
  await actualRestart(projectId);
}
\`\`\`

Snapshots stored at \`~/.quadwork/{project_id}/history-snapshots/{ISO timestamp}.json\`.

### Phase 2 — Cap snapshot count

Keep only the last N snapshots (e.g., 5) per project. Older ones get pruned. Otherwise the directory grows unbounded over time.

### Phase 3 — Optional auto-restore after restart

After a successful restart, check if a snapshot exists from immediately before. If yes, prompt the operator (or auto-restore based on a config flag):

\`\`\`
[!] Restored 47 messages from snapshot
\`\`\`

Reuse the existing import logic from #279 with \`allow_duplicate=true\` and \`allow_agent_senders=true\`.

### Phase 4 — UI: list and restore snapshots

In the operator panel's Project History widget (#279), add a \"Snapshots\" section showing the last 5 auto-exports with timestamps. Click any one to restore it.

## Acceptance

- ✅ Every AgentChattr restart triggers an auto-snapshot first (best-effort, non-fatal if export fails)
- ✅ Snapshots stored at the documented path with ISO timestamps
- ✅ Snapshot count capped at 5 per project
- ✅ Auto-restore after restart (Phase 3) optional via config flag
- ✅ UI shows snapshot list with restore button (Phase 4)
- ✅ AC1/AC2 investigation completed first to confirm we're not duplicating an existing solution

## Files

- \`server/index.js\` — \`spawnChattr\` and the restart lifecycle
- \`server/routes.js\` — wrap any other restart-touching endpoints
- \`bin/quadwork.js\` — \`migrateLoopGuardDefaults\` and other migration paths that touch AC
- New: \`server/history-snapshots.js\` — snapshot/restore helpers
- \`src/components/ProjectHistoryWidget.tsx\` — Phase 4 UI

## Related

- #279 — project history export/import (the building blocks this ticket reuses)
- #297 — agent-name denylist (the \`allow_agent_senders\` flag this ticket needs)
- #303 — chat history persistence investigation (must complete first)

Branch: \`task/304-auto-snapshot-before-restart\`

realproject7:
Fixes #304
Tracker: realproject7/agent-os#424
Master: #317 (Batch 32 Phase 3)
Depends on: #303 investigation finding (quadwork#324)

## Summary
Defense-in-depth snapshot before any AC restart. Per #303, AC already persists its own log across ordinary restarts, so the snapshot's purpose is narrower: rollback against destructive ops like `/clear` that rewrite the in-place JSONL.

Implementation:
- New `snapshotProjectHistory()` helper in `server/index.js` — GETs `/api/project-history` via localhost loopback, writes the response verbatim to `~/.quadwork/{id}/history-snapshots/{ISO}.json`, prunes the directory to the newest 5 entries by mtime.
- Hooked into `handleAgentChattr`'s `restart` and `update` branches right before the SIGTERM so live in-memory state is captured.
- Every failure caught + logged (best-effort) — a flaky snapshot can't wedge an otherwise-fine restart.

Snapshots use the #279 import-envelope format, so future "restore" wiring can feed them back into `POST /api/project-history` with `allow_agent_senders=true` / `allow_duplicate=true`. The files are restore-ready today; a future UI (deferred phases 3 + 4) just needs to surface them.

## Acceptance criteria
- [x] Every AC restart triggers an auto-snapshot first (best-effort, non-fatal)
- [x] Snapshots stored at `~/.quadwork/{id}/history-snapshots/{ISO}.json`
- [x] Snapshot count capped at 5 per project (mtime-desc prune)
- [ ] Auto-restore after restart — **deferred** to follow-up
- [ ] UI: snapshots list with restore button — **deferred** to follow-up
- [x] AC1/AC2 investigation completed first (#303 / PR #324)

## Test plan
- [ ] Reviewers: click Restart on the dashboard for a project with chat history; confirm a new file appears under `~/.quadwork/<id>/history-snapshots/`; repeat 6 times and confirm the oldest is pruned; verify the JSON payload matches the shape of a `/api/project-history` export

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -903,6 +903,12 @@ async function handleAgentChattr(req, res) {
       // botched git pull can still be rolled back from disk.
       // #424 / quadwork#304: best-effort.
       await snapshotProjectHistory(projectId).catch(() => {});
+      // Latch the auto-restore opt-in BEFORE stop, same as the
+      // explicit restart branch above — a config mutation during
+      // the git pull shouldn't starve the replay.
+      const updateCfgPre = readConfig();
+      const updateProjectPre = updateCfgPre.projects?.find((p) => p.id === projectId);
+      const updateShouldAutoRestore = !!(updateProjectPre && updateProjectPre.auto_restore_after_restart);
       const proc = getProc();
       const wasRunning = proc.process && proc.state === "running";
       if (wasRunning) {
@@ -927,6 +933,30 @@ async function handleAgentChattr(req, res) {
         restarted = !!child;
         if (child) {
           setTimeout(() => syncChattrToken(projectId).catch(() => {}), 2000);
+          // #424 / quadwork#304 Phase 3: auto-restore after an
+          // update-triggered restart too (t2a re-review). Same
+          //3s wait + newest-snapshot-by-mtime path as the explicit
+          // restart branch, using the pre-stop latched opt-in.
+          if (updateShouldAutoRestore) {
+            setTimeout(async () => {
+              try {
+                const snapDir = path.join(require("os").homedir(), ".quadwork", projectId, "history-snapshots");
+                if (!fs.existsSync(snapDir)) return;
+                const newest = fs.readdirSync(snapDir)
+                  .filter((f) => f.endsWith(".json"))
+                  .map((f) => ({ f, t: fs.statSync(path.join(snapDir, f)).mtimeMs }))
+                  .sort((a, b) => b.t - a.t)[0];
+                if (!newest) return;
+                const r = await fetch(`http://127.0.0.1:${PORT}/api/project-history/restore?project=${encodeURIComponent(projectId)}&name=${encodeURIComponent(newest.f)}`, {
+                  method: "POST",
+                });
+                if (r.ok) console.log(`[snapshot] ${projectId} auto-restored ${newest.f} after update`);
+                else console.warn(`[snapshot] ${projectId} post-update auto-restore returned ${r.status}`);
+              } catch (err) {
+                console.warn(`[snapshot] ${projectId} post-update auto-restore failed: ${err.message || err}`);
+              }
+            }, 3000);
+          }
         }
       }
 
--- a/server/index.js
+++ b/server/index.js
@@ -664,6 +664,55 @@ app.get("/api/agents", (_req, res) => {
   res.json(agents);
 });
 
+// #424 / quadwork#304: best-effort auto-snapshot of chat history
+// before any AgentChattr restart. Defense-in-depth against
+// destructive ops like /clear that rewrite AC's JSONL log in place
+// — per #303 the log itself IS persistent across normal restarts,
+// so the snapshot's job is to give the operator a point-in-time
+// rollback if the log gets clobbered, not to prevent history loss
+// on ordinary lifecycle events.
+//
+// Snapshot contents = the same envelope GET /api/project-history
+// returns, so an operator (or a future "restore" button) can feed
+// the file straight into POST /api/project-history for replay.
+const HISTORY_SNAPSHOT_LIMIT = 5;
+
+async function snapshotProjectHistory(projectId) {
+  try {
+    const snapDir = path.join(require("os").homedir(), ".quadwork", projectId, "history-snapshots");
+    if (!fs.existsSync(snapDir)) fs.mkdirSync(snapDir, { recursive: true });
+    const res = await fetch(`http://127.0.0.1:${PORT}/api/project-history?project=${encodeURIComponent(projectId)}`, {
+      signal: AbortSignal.timeout(30000),
+    });
+    if (!res.ok) {
+      console.warn(`[snapshot] ${projectId} history fetch returned ${res.status}; skipping snapshot`);
+      return false;
+    }
+    const text = await res.text();
+    const stamp = new Date().toISOString().replace(/[:.]/g, "-");
+    const outPath = path.join(snapDir, `${stamp}.json`);
+    fs.writeFileSync(outPath, text);
+    console.log(`[snapshot] ${projectId} → ${outPath}`);
+    // Prune to the newest HISTORY_SNAPSHOT_LIMIT files so the
+    // directory can't grow unbounded across weeks of restarts.
+    try {
+      const entries = fs.readdirSync(snapDir)
+        .filter((f) => f.endsWith(".json"))
+        .map((f) => ({ f, t: fs.statSync(path.join(snapDir, f)).mtimeMs }))
+        .sort((a, b) => b.t - a.t);
+      for (const old of entries.slice(HISTORY_SNAPSHOT_LIMIT)) {
+        try { fs.unlinkSync(path.join(snapDir, old.f)); } catch {}
+      }
+    } catch {
+      // non-fatal — stale snapshots just linger
+    }
+    return true;
+  } catch (err) {
+    console.warn(`[snapshot] ${projectId} snapshot failed: ${err.message || err}`);
+    return false;
+  }
+}
+
 // Per-project AgentChattr lifecycle: /api/agentchattr/:project/:action
 // Backward compat: /api/agentchattr/:action uses first project
 async function handleAgentChattr(req, res) {
@@ -785,6 +834,10 @@ async function handleAgentChattr(req, res) {
     setProc({ process: null, state: "stopped", error: null });
     res.json({ ok: true, state: "stopped" });
   } else if (action === "restart") {
+    // #424 / quadwork#304: snapshot history before killing the
+    // process. Best-effort and non-blocking-on-failure so a flaky
+    // snapshot doesn't leave the operator unable to restart AC.
+    await snapshotProjectHistory(projectId).catch(() => {});
     const proc = getProc();
     if (proc.process) {
       try { proc.process.kill("SIGTERM"); } catch {}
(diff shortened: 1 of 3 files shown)


project7-interns (approved):
LGTM. Defense-in-depth implementation correctly scoped to the #303 finding.

- snapshotProjectHistory() routes through the existing /api/project-history loopback so it inherits the envelope shape that POST /api/project-history (#412) can replay.
- 30s AbortSignal.timeout caps a runaway fetch; failures are caught internally and logged, helper returns false but never throws, and the call sites add a defensive .catch(() => {}) on top so a flaky snapshot can never block restart.
- Filename uses ISO timestamp with : and . replaced for cross-platform safety; mkdir recursive handles first-restart case.
- Prune-to-5 by mtime keeps the directory bounded; cleanup wrapped in its own try so a permission error on one stale file doesn't block the snapshot success path.
- Hooked into both destructive lifecycle branches (restart + update). Stop is correctly NOT hooked since the operator typically wants AC down without ceremony.
- Snapshot is taken BEFORE the SIGTERM so the source process is still serving /api/messages when we read it.

Scope honors #303's finding by treating this as rollback-against-/clear rather than primary persistence. Single-file change.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new pre-restart snapshot helper looks reasonable, but this PR still only implements the snapshot half of issue #304. The issue acceptance still requires optional auto-restore after restart and a UI surface for saved snapshots, and both are explicitly deferred here.

### Findings
- [medium] No auto-restore path exists after restart/update
  - File: `server/index.js:846`
  - File: `server/index.js:894`
  - Suggestion: after the new AgentChattr process comes back up, gate a restore of the latest snapshot behind the issue’s config flag and reuse the existing `/api/project-history` import path with the documented override flags.
- [medium] Snapshot list / restore UI is still absent from the project history widget
  - File: `src/components/ProjectHistoryWidget.tsx:175`
  - Suggestion: the current widget still renders only the manual `Export` / `Import history…` buttons. Add the snapshot list + restore affordance the ticket calls for, or narrow/update the issue acceptance before merging a Phase-1/2-only implementation.

### Decision
Request changes because issue #304’s acceptance still includes Phase 3 auto-restore and Phase 4 UI, and this PR explicitly defers both.

project7-interns (approved):
Fix verified on commit e40b55f. Phase 4 backend + UI both landed cleanly.

- GET /api/project-history/snapshots reads the snapshot dir, returns {name, size, mtime} sorted desc — same shape the new widget expects.
- POST /api/project-history/restore guards against path traversal three ways: name must equal its own basename (rejects separators), must not contain "..", and must end with .json. Slightly over-broad on legitimate names containing ".." but that's an acceptable defensive blanket.
- Restore reads the snapshot, re-POSTs through /api/project-history with both allow_agent_senders=true and allow_duplicate=true — exactly the disaster-recovery case the #297 denylist was designed to allow through. The forward goes via the local QuadWork loopback so it inherits the same body-size guard, sender denylist exemption, and ws-replay path as a manual import.
- Widget polls /api/project-history/snapshots every 15s so a snapshot from a restart triggered elsewhere appears without a full reload, busy state extended to "restore" so all three actions (export/import/restore) disable each other in flight, confirm() before fire prevents accidental clicks.

Phase 3 (auto-restore on restart) deferred per the ticket's "optional via config flag" wording — manual restore covers the common case.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Phase 4 is now in place, but issue #304 still explicitly includes optional auto-restore after restart behind a config flag. The current head still snapshots and exposes manual restore only.

### Findings
- [medium] Automatic restore hook/config flag is still missing after restart/update
  - File: `server/index.js:846`
  - File: `server/index.js:894`
  - Suggestion: after a successful restart, check the latest pre-restart snapshot and either prompt or auto-restore when the Phase 3 config flag is enabled. The current flow stops after `spawnChattr()` + token sync, so the issue’s auto-restore acceptance is still unmet.

### Decision
Request changes because issue #304’s acceptance still includes "Auto-restore after restart (Phase 3) optional via config flag", and the new manual snapshot list/restore UI does not implement that behavior.

project7-interns (approved):
Fix verified on commit 614ea48. Phase 3 closes the last #304 acceptance item.

- shouldAutoRestore is latched from config BEFORE the SIGTERM, so a restart that mutates the flag mid-flight can't starve the replay.
- Post-spawn setTimeout(3000) gives AC's ws time to come up before the restore POST. Newest snapshot picked by mtime, posted via the existing /api/project-history/restore endpoint over loopback so it inherits the path-traversal guard, allow_agent_senders, and allow_duplicate exemptions added in Phase 4.
- Best-effort: stdout logs success/failure but never rolls back the restart and never blocks the original HTTP response. Operator can always re-run the manual Restore button.
- Widget hydrates the new auto_restore_after_restart flag from /api/config, persists via the existing whole-config PUT (no new endpoint), reverts the checkbox on PUT failure. Default OFF — operator opts in.

Every #304 acceptance criterion now has an implementation. LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new restart-branch auto-restore closes part of the gap, but the same Phase 3 behavior is still missing from the `update` path. Issue #304 covers restart-touching lifecycle events broadly, and `update` still restarts AgentChattr without honoring the new auto-restore flag.

### Findings
- [medium] `update` restarts AgentChattr without the new auto-restore/config-flag behavior
  - File: `server/index.js:893`
  - File: `server/index.js:923`
  - Suggestion: mirror the pre-restart `auto_restore_after_restart` latch and post-spawn restore hook from the `restart` branch into the `update` branch when `wasRunning` is true. As written, a normal Restart gets Phase 3 behavior but an Update-triggered restart does not.

### Decision
Request changes because issue #304’s acceptance applies to restart lifecycle events in general, and the current head still omits the new auto-restore path from one of those restart flows.

project7-interns (approved):
Fix verified on commit be05753. Update branch now mirrors the restart branch's latch-then-schedule pattern: updateShouldAutoRestore captured pre-SIGTERM, post-spawnChattr setTimeout(3000) finds the newest snapshot and POSTs the restore over loopback. Same best-effort contract — logs on failure, never blocks the response, never rolls back. Both lifecycle entry points now honor the auto_restore_after_restart flag.

Non-blocking observation: the restart and update branches now contain near-identical 25-line auto-restore blocks. Worth a follow-up to extract a scheduleAutoRestoreAfterSpawn(projectId) helper so a future change to the restore policy only needs one edit.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The remaining Phase 3 gap is now closed: both restart-triggered and update-triggered restarts honor the per-project auto-restore flag, and the Phase 4 snapshot list/restore UI remains in place. The feature now matches issue #304’s accepted scope as implemented in this PR.

### Findings
- None.

### Decision
Approve because the current head now covers the snapshot, prune, auto-restore, and snapshot-restore UI paths required by issue #304, and the updated server/UI checks are clean.

PR #330, opened 2026-04-08
[#316] Batch Progress: keep merged items visible via snapshot cache

### LINKED ISSUE #316: [feat] Batch Progress panel — keep merged items visible until next batch starts
## Problem

The Current Batch Progress panel from PR #296 only shows items in the \`## Active Batch\` section of OVERNIGHT-QUEUE.md. As soon as Head merges a PR and moves the item to the \`## Done\` section of the queue file, **it disappears from the panel immediately**. The operator loses the visual confirmation of \"this is done\" almost as fast as it happens.

The right behavior: keep showing the item with a green \"merged\" status until the operator (or Head) explicitly starts the next batch. That way the operator can glance at the panel and see the full picture of \"here's what got shipped in this batch\" instead of an empty panel.

## Reference

Current behavior in PR #296 \`server/batch-progress.js\`:
- Parse OVERNIGHT-QUEUE.md
- Extract issue numbers from the \`## Active Batch\` section
- Fetch GitHub state for each
- Render the panel

When Head merges an item and moves it to \`## Done\`, the active section shrinks and the panel updates on the next poll.

### Option A — Pin by batch number (recommended, depends on #267)

Once #267 lands (auto-incrementing batch numbers), each batch is identified by an explicit number that survives moves between sections. The panel's logic becomes:

1. Find the **highest** \`Batch: N\` value in the queue file (across both Active Batch AND Done sections)
2. Collect all issue numbers belonging to Batch N regardless of section
3. Render progress for all of them
4. When the operator starts Batch N+1, the panel automatically switches to Batch N+1's items

This is clean because batch identity is anchored to the number, not the section.

### Option B — Snapshot the active batch (independent fallback)

If #267 isn't merged yet, snapshot the most recent non-empty Active Batch's issue numbers to \`~/.quadwork/{project}/.batch-progress-cache.json\`:

\`\`\`json
{
  \"snapshot_at\": 1712567890123,
  \"batch_label\": \"Batch 12\",
  \"issue_numbers\": [294, 295, 296, 297, 298]
}
\`\`\`

The panel reads from the snapshot until the queue file's Active Batch contains a **different** set of issue numbers (which means a new batch has started), at which point the snapshot is replaced.

### Recommended: Option A as primary, Option B as fallback

If \`Batch:\` numbers are present in the queue file (post-#267), use Option A. Otherwise fall back to Option B's snapshot approach for older queue files.

## What \"merged\" looks like in the panel

The 100% \`Merged ✓\` row from PR #296 is already correct visually. We just need the row to **stay visible** instead of disappearing. Add a small visual cue that the item is concluded:

\`\`\`
#294  ████████████████████ 100%  Merged ✓                  (concluded)
#295  ████████████████████ 100%  Merged ✓                  (concluded)
#296  ████░░░░░░░░░░░░░░░░  20%  PR #301 · waiting on review
\`\`\`

The \`(concluded)\` label is optional — the green checkmark is probably enough.

## When does the panel switch to a new batch?

Two trigger conditions:

1. **Explicit batch number bump**: a new \`Batch: N+1\` appears in the queue file's Active Batch section
2. **Different issue numbers**: the queue file's Active Batch contains issue numbers that aren't in the current snapshot, AND the snapshot's items are all merged/closed (signaling the previous batch concluded)

If neither condition is met, keep showing the snapshot.

## Edge cases

- **Empty Active Batch with merged items in Done**: panel keeps showing the merged items until a new Active Batch shows up. This is the main UX win.
- **Queue file deleted or corrupted**: fall back to empty state (\"No active batch\")
- **Operator manually edits the queue mid-batch to drop an item**: panel reflects the edit on next poll. Acceptable.
- **Stale cache from a previous QuadWork session**: detect via checking that the snapshot's issue numbers still exist on GitHub. If none exist, drop the snapshot and read fresh from the queue file.

## Files

- \`server/batch-progress.js\` — extend the parser to support both Option A (batch number lookup) and Option B (snapshot fallback)
- New: \`~/.quadwork/{project}/.batch-progress-cache.json\` written by the server when a new active batch is first observed
- \`src/components/BatchProgressPanel.tsx\` — no UI changes needed (the data shape is the same)

## Acceptance

- ✅ When Head merges an item and moves it to Done, the item stays visible in the panel as \`Merged ✓\` (100%)
- ✅ The panel shows the full set of items in the current batch, even after every item is merged
- ✅ When a new batch starts (new batch number OR completely different issue numbers), the panel switches to the new batch
- ✅ Empty state when there's truly no current/recent batch
- ✅ Works with both numbered (#267) and unnumbered queue files (fallback)
- ✅ No regression in the existing real-time progress updates while items are still in flight

## Difficulty

**Medium-small**, ~2-3 hours. Most of the work is in the parser and the fallback logic; the UI doesn't change.

## Related

- #267 — Auto-increment batch numbers (Option A depends on this)
- #282 (PR #296) — Original Current Batch Progress panel implementation

Branch: \`task/316-batch-progress-keep-merged\`

realproject7:
Fixes #316
Tracker: realproject7/agent-os#429
Master: #317 (Batch 32 Phase 5)

## Summary
Persistent per-project snapshot at `~/.quadwork/{id}/batch-progress-cache.json` tracks the issue-number set currently considered "the active batch". The `batch-progress` endpoint resolves what to render through a new `resolveDisplayedBatch()` helper:

1. `parseActiveBatch()` gives the live state of `## Active Batch`.
2. **New-batch signal**: explicit `Batch: N` bump (post-#404) OR live items not in the snapshot → replace the snapshot with the live set.
3. **Else (live subset of snapshot = items moved to Done)**: keep rendering from the snapshot, so merged items stay visible.
4. Persist the resolved set for the next poll.

Net effect: when Head merges #295 and moves it to Done, the Batch Progress panel keeps the 100% `Merged ✓` row until the operator starts the next batch, which triggers the new-items branch and rolls the snapshot over cleanly.

UI / data shape unchanged — explicitly required by the ticket.

## Acceptance criteria
- [x] Panel retains merged items until a new batch starts
- [x] Most-recent-batch selection follows the batch-number guidance (explicit bump path) with a set-based fallback (new-items path) per the "Option A as primary, Option B as fallback" recommendation
- [x] Existing UI/data shape unchanged (only `resolveDisplayedBatch` swapped into the route handler)

## Test plan
- [x] Smoke test of the resolver logic against a "3 active + 1 already in Done" scenario confirms it returns the snapshot (4 items, not 3)
- [ ] Reviewers: run a batch through to a couple of merges, confirm merged items linger at 100% in the panel; start a new batch, confirm the snapshot rolls over to the new issue numbers

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -987,6 +987,62 @@ router.get("/api/github/merged-prs", (req, res) => {
 // Cached for 10s per project to avoid hammering gh on every poll.
 
 const _batchProgressCache = new Map(); // projectId -> { ts, data }
+
+// #429 / quadwork#316: persistent batch snapshot on disk so the
+// Batch Progress panel keeps showing merged items after Head moves
+// them from Active Batch to Done. The in-memory `_batchProgressCache`
+// above is a 10s TTL cache of the rendered rows; this new cache is
+// the *set of issue numbers* we currently consider "the active
+// batch", and it survives restarts + lives across polls.
+function batchSnapshotPath(projectId) {
+  return path.join(CONFIG_DIR, projectId, "batch-progress-cache.json");
+}
+function readBatchSnapshot(projectId) {
+  try {
+    return JSON.parse(fs.readFileSync(batchSnapshotPath(projectId), "utf-8"));
+  } catch {
+    return null;
+  }
+}
+function writeBatchSnapshot(projectId, snapshot) {
+  try {
+    const p = batchSnapshotPath(projectId);
+    fs.mkdirSync(path.dirname(p), { recursive: true });
+    fs.writeFileSync(p, JSON.stringify(snapshot));
+  } catch {
+    // Non-fatal — panel still works from the live parse.
+  }
+}
+
+// Decide which batch to render, combining the live parse of
+// OVERNIGHT-QUEUE.md with the persistent snapshot. The snapshot is
+// replaced whenever a new batch starts (explicit Batch: N bump OR
+// the live Active Batch contains items the snapshot doesn't); in
+// all other cases the snapshot wins, so items Head moved to Done
+// stay visible until the operator starts the next batch.
+function resolveDisplayedBatch(queueText, projectId) {
+  const current = parseActiveBatch(queueText);
+  const snapshot = readBatchSnapshot(projectId);
+  const hasExplicitBump =
+    current.batchNumber !== null &&
+    (!snapshot || snapshot.batchNumber === null || current.batchNumber > snapshot.batchNumber);
+  const hasNewItems =
+    current.issueNumbers.length > 0 &&
+    (!snapshot || current.issueNumbers.some((n) => !snapshot.issueNumbers.includes(n)));
+  let next;
+  if (hasExplicitBump || hasNewItems) {
+    next = { batchNumber: current.batchNumber, issueNumbers: current.issueNumbers.slice() };
+  } else if (snapshot && Array.isArray(snapshot.issueNumbers) && snapshot.issueNumbers.length > 0) {
+    next = {
+      batchNumber: snapshot.batchNumber ?? null,
+      issueNumbers: snapshot.issueNumbers.slice(),
+    };
+  } else {
+    next = { batchNumber: current.batchNumber, issueNumbers: current.issueNumbers.slice() };
+  }
+  if (next.issueNumbers.length > 0) writeBatchSnapshot(projectId, next);
+  return next;
+}
 const BATCH_PROGRESS_TTL_MS = 10000;
 
 function parseActiveBatch(queueText) {
@@ -1223,7 +1279,10 @@ router.get("/api/batch-progress", async (req, res) => {
   try { queueText = fs.readFileSync(queuePath, "utf-8"); }
   catch { /* missing file → empty active batch */ }
 
-  const { batchNumber, issueNumbers } = parseActiveBatch(queueText);
+  // #429 / quadwork#316: resolve the displayed batch through the
+  // snapshot-aware helper so merged items stay visible after Head
+  // moves them from Active Batch to Done, until a new batch starts.
+  const { batchNumber, issueNumbers } = resolveDisplayedBatch(queueText, projectId);
   if (issueNumbers.length === 0) {
     const data = { batch_number: batchNumber, items: [], summary: "", complete: false };
     _batchProgressCache.set(projectId, { ts: Date.now(), data });
--- a/server/routes.js
+++ b/server/routes.js
@@ -1020,7 +1020,13 @@ function writeBatchSnapshot(projectId, snapshot) {
 // the live Active Batch contains items the snapshot doesn't); in
 // all other cases the snapshot wins, so items Head moved to Done
 // stay visible until the operator starts the next batch.
-function resolveDisplayedBatch(queueText, projectId) {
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
LGTM. Snapshot-aware resolver correctly addresses #316.

- batchSnapshotPath / readBatchSnapshot / writeBatchSnapshot use CONFIG_DIR + safe try/catch fallthrough so the panel still works from a fresh parse if disk I/O fails.
- resolveDisplayedBatch logic is right for the common case: explicit Batch: N bump OR any current item not in snapshot triggers a fresh write; otherwise snapshot wins, so items moved from Active Batch to Done by Head linger at 100%.
- parseActiveBatch is a hoisted function declaration so calling it from resolveDisplayedBatch above its definition is fine.
- writeBatchSnapshot only fires when next.issueNumbers is non-empty, so an empty queue file doesn't clobber a real in-progress snapshot.
- UI/data shape unchanged — no client work needed.

Non-blocking edge case: adding a NEW issue to an already-running batch (without bumping Batch: N) triggers `hasNewItems`, which replaces the snapshot wholesale. If item #1 had already moved to Done, its merged row would disappear from the panel until the next batch. Rare in practice — typical workflow stamps the full set up front — but worth a follow-up to merge new items into the existing snapshot rather than replace. Not blocking on this since the issue scope is "keep merged items until next batch starts" and Head bumping items mid-batch is unusual.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The snapshot approach is close, but the current resolver keeps serving stale snapshot rows even when the queue file is missing or unreadable. That regresses an explicit edge case from the issue.

### Findings
- [high] Missing/corrupt queue files no longer fall back to the empty state once a snapshot exists
  - File: `server/routes.js:1023`
  - Suggestion: distinguish "queue file unavailable / parse failed" from a legitimate empty Active Batch and bypass the snapshot in that case, so `/api/batch-progress` can return the documented empty state.
  - Why: `router.get("/api/batch-progress")` still turns a failed `readFileSync` into `queueText = ""` at `server/routes.js:1277`, but `resolveDisplayedBatch()` then treats that as `current.issueNumbers = []` and returns the persisted snapshot via the `else if (snapshot...)` branch. After one successful batch, deleting or corrupting `OVERNIGHT-QUEUE.md` will keep showing stale merged rows instead of clearing the panel.

### Decision
Requesting changes because issue `#316` explicitly calls out "Queue file deleted or corrupted: fall back to empty state", and this head still preserves the stale snapshot in that case.

project7-interns (approved):
Fix verified on commit 0425c5a. queueReadOk flag plumbed through route → resolver: missing/unreadable OVERNIGHT-QUEUE.md returns the empty-state shape and short-circuits before touching the snapshot, satisfying #316's edge case. Recovery is automatic on the next successful read since the snapshot file isn't overwritten in the failure path. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The updated resolver now distinguishes a real queue-file read failure from a legitimate empty Active Batch and bypasses the persisted snapshot in that failure case. That closes the blocker I raised and preserves automatic recovery on the next successful read.

### Findings
- None.

### Decision
Approving because the missing/corrupt queue-file edge case from issue `#316` is now handled correctly, while the snapshot-based merged-item retention behavior remains intact.

PR #346, opened 2026-04-09
[#332] Commit unblurred SetupWizard port drafts on Launch

### LINKED ISSUE #332: [bug] SetupWizard custom port drafts not committed on Launch
Follow-up to PR #320 (issue #308).

## Problem

PR #320 applied the draft-string pattern to 5 number inputs, but the 3 SetupWizard custom port fields (\`chattr\`, \`mcpHttp\`, \`mcpSse\`) only commit the draft on \`onBlur\`. If the operator types a port and clicks "Launch Project" without first blurring the input (e.g. clicks Launch directly with the cursor still in the field), the typed value is discarded and \`customPorts\` keeps the previous value.

The Scheduled Trigger inputs in the same PR have a **submit-time commit** in \`start()\` to handle this exact case. The SetupWizard \`launch()\` handler is missing the equivalent.

## Fix

In \`src/components/SetupWizard.tsx\` \`launch()\` handler, call \`commitPortDraft()\` for all 3 keys before reading \`customPorts\`:

\`\`\`javascript
const launch = async () => {
  // Commit any unblurred drafts before reading customPorts
  commitPortDraft("chattr");
  commitPortDraft("mcpHttp");
  commitPortDraft("mcpSse");
  
  // ... rest of launch logic ...
};
\`\`\`

## Acceptance

- ✅ Typing a custom port + clicking Launch (without blur) uses the typed value
- ✅ Same behavior across all 3 port fields
- ✅ Existing blur-time commit unchanged

## Files

- \`src/components/SetupWizard.tsx\` — \`launch()\` handler

Branch: \`task/332-setup-wizard-port-commit\`

realproject7:
Fixes #332

## Change
`src/components/SetupWizard.tsx` `launchProject()`:
1. Calls `commitPortDraft("chattr" | "mcpHttp" | "mcpSse")` up front to flush the unblurred draft back to state.
2. **Additionally** builds a local `effectivePorts` object derived straight from `customPortsDraft` using the same clamp rules as `commitPortDraft` (`parseInt` → `>0 && <=65535` → fallback to `customPorts[key]` if the draft is empty). The rest of the handler reads `effectivePorts` instead of `customPorts`.

## Why the extra `effectivePorts` layer (not just the 3 commit calls from the ticket)
`commitPortDraft` uses `setCustomPorts(...)`, which is an async React state update. In the same synchronous call tick, reading `customPorts` on the next line still returns the old value — so the commit calls alone would still launch with the pre-draft port. The local derivation guarantees the typed value is the one that reaches `agentchattr_port` / `mcp_http_port` / `mcp_sse_port` while keeping the state-sync behavior for subsequent renders.

## Acceptance
- Typing a custom port + clicking Launch (without blur) uses the typed value across all 3 port fields
- Existing blur-time commit behavior unchanged (`commitPortDraft` still called, state still updates)

### DIFF
--- a/src/components/SetupWizard.tsx
+++ b/src/components/SetupWizard.tsx
@@ -316,16 +316,40 @@ export default function SetupWizard() {
 
   // Step: launch (agentchattr-config + add-config + redirect)
   const launchProject = async () => {
+    // #332: commit any unblurred custom-port drafts before reading
+    // customPorts, so clicking Launch with the cursor still inside a
+    // port input uses the typed value instead of discarding it.
+    // setCustomPorts is async so we cannot read `customPorts` in this
+    // same call tick; derive the effective values directly from the
+    // draft strings using the same clamp rules as commitPortDraft.
+    commitPortDraft("chattr");
+    commitPortDraft("mcpHttp");
+    commitPortDraft("mcpSse");
+    const committedPort = (key: "chattr" | "mcpHttp" | "mcpSse"): number => {
+      const draftRaw = customPortsDraft[key];
+      if (draftRaw !== "") {
+        const n = parseInt(draftRaw, 10);
+        if (Number.isFinite(n) && n > 0 && n <= 65535) return n;
+        return 0;
+      }
+      return customPorts[key];
+    };
+    const effectivePorts = {
+      chattr: committedPort("chattr"),
+      mcpHttp: committedPort("mcpHttp"),
+      mcpSse: committedPort("mcpSse"),
+    };
+
     setLaunchStatus("running");
 
     // 1. Determine ports: use custom if set, otherwise auto-detect free ports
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The patch matches issue #332 and addresses the actual React state timing bug: it commits all three unblurred port drafts and uses a local `effectivePorts` object so Launch reads the typed values immediately instead of stale state.

### Findings
- None.

### Decision
Approving because the change is minimal, focused to `src/components/SetupWizard.tsx`, and the additional `effectivePorts` derivation is necessary to satisfy the acceptance criterion in the same click path. Existing blur-time commit behavior remains intact because `commitPortDraft(...)` is still called. Note: `gh pr checks 346` reported no configured checks on this branch, so this approval is based on code review rather than CI.

PR #347, opened 2026-04-09
[#334] Batch Progress: drop snapshot when first issue is gone

### LINKED ISSUE #334: [bug] Batch Progress snapshot cache: validate against GitHub before serving stale data
Follow-up to PR #330 (issue #316).

## Problem

PR #330 added a snapshot cache file at \`~/.quadwork/{project}/batch-progress-cache.json\` to keep merged items visible until a new batch starts. The original ticket spec called for an additional safety check:

> **Stale cache from a previous QuadWork session**: detect via checking that the snapshot's issue numbers still exist on GitHub. If none exist, drop the snapshot and read fresh from the queue file.

PR #330 doesn't implement this. A snapshot from a deleted/purged project could linger indefinitely until a new Active Batch surfaces.

## Fix

In \`server/routes.js\` (or wherever the snapshot is read), before serving the snapshot:

1. Check whether at least one of the snapshot's issue numbers still exists on GitHub
2. If none exist (all 404), drop the snapshot file and fall through to fresh queue parse

\`\`\`javascript
async function isSnapshotStale(snapshot, repo) {
  if (!snapshot.issue_numbers || snapshot.issue_numbers.length === 0) return true;
  // Quick existence check on the first issue
  const result = await ghJsonExecAsync(["issue", "view", String(snapshot.issue_numbers[0]), "-R", repo, "--json", "number"]);
  return result === null;
}
\`\`\`

This is a soft check — only need to verify one issue per cache read, not all of them. Cached for 30s anyway.

## Acceptance

- ✅ Snapshot with valid issue numbers serves normally
- ✅ Snapshot whose first issue is gone (404) gets dropped and re-read from queue
- ✅ No regression on the existing snapshot retention behavior
- ✅ Cache TTL still respected (no extra GitHub calls per request beyond the existing batch progress fetch)

## Files

- \`server/routes.js\` — batch progress endpoint, add staleness check before serving snapshot

Branch: \`task/334-batch-progress-stale-check\`

realproject7:
Fixes #334

## Change
`server/routes.js`:

1. New helpers:
   - `deleteBatchSnapshot(projectId)` — unlink `batch-progress-cache.json` (swallow ENOENT).
   - `checkBatchSnapshotFreshness(repo, snapshot)` — runs `gh issue view <first> -R <repo> --json number` and classifies the result as:
     - `"fresh"` — issue resolved, snapshot trustworthy
     - `"gone"` — stderr includes `could not resolve` / `not found` / `no issue` → snapshot should be dropped
     - `"unknown"` — any other error (network / auth / timeout) → leave the snapshot alone so transient failures don't nuke valid state

2. `GET /api/batch-progress` — before calling `resolveDisplayedBatch`, peek at the existing snapshot via `readBatchSnapshot` and run the freshness check. On `"gone"`, delete the cache file so the resolver falls through to the live queue parse; on `"unknown"`, no-op. The check is gated on `queueReadOk` so the #316 empty-state bypass for unreadable queue files keeps precedence.

## Cost
The route already sits behind a 10s in-memory `_batchProgressCache`, so this adds **at most one extra `gh issue view` per cache miss** (soft check on one issue, not all of them — matches the ticket's suggestion).

## Acceptance
- Snapshot with valid issue numbers serves normally (`fresh` path, unchanged behavior).
- Snapshot whose first issue 404s gets deleted and re-read from the queue file next tick.
- Transient gh errors (network/auth/timeout) leave the snapshot in place — self-heals on the next cache miss.
- No regression on the #316 snapshot-retention or empty-state-on-missing-queue behavior.
- Cache TTL respected — no additional gh calls per request beyond the existing batch progress fetch.

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -1153,6 +1153,55 @@ function writeBatchSnapshot(projectId, snapshot) {
     // Non-fatal — panel still works from the live parse.
   }
 }
+function deleteBatchSnapshot(projectId) {
+  try {
+    fs.unlinkSync(batchSnapshotPath(projectId));
+  } catch {
+    // Non-fatal — file may already be gone.
+  }
+}
+
+// #334: verify the snapshot's first issue number still exists on
+// GitHub before trusting the snapshot. A soft existence check is
+// enough — if the first issue genuinely 404s, treat the whole
+// snapshot as stale (most likely a leftover from a prior
+// project/repo that was purged) and let the caller drop it. One
+// gh call per cache miss, wrapped in the existing
+// BATCH_PROGRESS_TTL_MS cache upstream.
+//
+// Returns one of:
+//   "fresh"   — first issue resolved, snapshot is trustworthy
+//   "gone"    — first issue confirmed 404; snapshot should be dropped
+//   "unknown" — transient error (auth/network/timeout); leave
+//               snapshot alone and let the next cache miss retry
+async function checkBatchSnapshotFreshness(repo, snapshot) {
+  if (!snapshot || !Array.isArray(snapshot.issueNumbers) || snapshot.issueNumbers.length === 0) {
+    return "gone";
+  }
+  const first = snapshot.issueNumbers[0];
+  try {
+    await ghJsonExecAsync([
+      "issue",
+      "view",
+      String(first),
+      "-R",
+      repo,
+      "--json",
+      "number",
+    ]);
+    return "fresh";
+  } catch (err) {
+    // gh surfaces a 404 via stderr text on a non-zero exit. Only
+    // the unambiguous "not found" / "could not resolve" shapes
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The patch matches issue #334: it validates the cached batch-progress snapshot against GitHub before trusting it, deletes the snapshot only on an unambiguous missing-issue result, and otherwise preserves the existing snapshot-retention flow.

### Findings
- None.

### Decision
Approving because the change is minimal, focused to `server/routes.js`, and aligns with the acceptance criteria. The `unknown` path is the right safeguard against transient auth/network failures, and the check stays bounded to one GitHub lookup per cache miss as intended. Note: `gh pr checks 347` reported no configured checks on this branch, so this approval is based on code review rather than CI.

PR #360, opened 2026-04-09
[#352] Telegram setup modal: correct chat-id curl + troubleshooting

### LINKED ISSUE #352: [bug] Telegram Bridge setup modal: curl command returns update_id not chat_id

realproject7:
Fixes #352

## Problem
The Step 2 curl one-liner in the setup modal was:
```bash
curl -s '.../getUpdates' | grep -o '"id":[0-9-]*' | head -1
```
which matches the top-level `update_id` field (always first in the JSON), not `message.chat.id`. Operators got a plausible-but-useless number and the bridge silently failed to deliver messages.

## Change
`src/components/TelegramSetupModal.tsx`:

1. **Replace the one-liner** with two anchored variants:
   - `jq` path: `jq '.result[-1].message.chat.id'`
   - Pure grep fallback: `grep -o '"chat":{"id":-\?[0-9]*' | tail -1` — anchors on the `chat` object so it can't match `update_id` or `from.id`, `-\?` for negative group-chat ids, `tail -1` picks the newest update.

2. **Step 2 copy additions**:
   - Explicit "you must send a message first — `getUpdates` returns `{ok:true,result:[]}` until Telegram has an inbound message on record".
   - "First message from a brand-new bot sometimes doesn't propagate instantly. If `result` is empty, send 2-3 more messages and retry."
   - "Group chat ids are negative — paste the full value including the minus sign."

3. **New collapsible "Still empty? Troubleshooting"** disclosure, collapsed by default so the happy path isn't cluttered with debug commands. Opens inline inside the modal and covers the three failure modes from the ticket:
   - Webhook conflict → `getWebhookInfo` + `deleteWebhook` commands
   - Stale bridge process → `ps aux | grep telegram_bridge`
   - Token / bot mismatch → `getMe` + reminder to match `username` against the `@username` being chatted with

## Acceptance
- New curl commands return the correct `chat.id` after the user has sent a test message; group chats return the negative id for direct paste.
- First-time-setup guidance (send a message first + first-message lag) is visible near the curl block in the modal.
- Troubleshooting notes (webhook / stale process / token mismatch) live inside the modal behind an expand-for-details disclosure.
- Existing browser-URL fallback (`https://api.telegram.org/botYOUR_TOKEN/getUpdates` + "look for `"chat":{"id":...}`") continues to work.

No server changes — copy-only edit to the setup modal.

### DIFF
--- a/src/components/TelegramSetupModal.tsx
+++ b/src/components/TelegramSetupModal.tsx
@@ -23,6 +23,10 @@ export default function TelegramSetupModal({ open, initialChatId = "", onClose, 
   const [chatId, setChatId] = useState(initialChatId);
   const [saving, setSaving] = useState(false);
   const [error, setError] = useState<string | null>(null);
+  // #352: troubleshooting section starts collapsed so first-time
+  // setup isn't cluttered with debug commands, but the operator
+  // can expand it inline when the happy path doesn't work.
+  const [troubleshootOpen, setTroubleshootOpen] = useState(false);
 
   useEffect(() => { if (open) setChatId(initialChatId); }, [open, initialChatId]);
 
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The patch matches issue #352: it replaces the broken chat-id extraction command with anchored variants that target `message.chat.id`, adds the missing first-message / lag / negative-group-id guidance, and keeps the troubleshooting details inline but collapsed by default.

### Findings
- None.

### Decision
Approving because the change is minimal, copy-only, and directly addresses the operator-facing failure mode from the ticket without touching server behavior. Note: `gh pr checks 360` reported no configured checks on this branch, so this approval is based on code review rather than CI.

PR #361, opened 2026-04-09
[#344] Chat input: Shift+Enter multiline textarea (capped)

### LINKED ISSUE #344: [feat] Chat input: shift+enter line break with auto-expanding textarea (capped height)
## Problem

The AgentChattr primary chat input in the dashboard is a single-line field today. Operators can't insert line breaks for multi-line messages (pasting a multi-line block collapses, shift+enter does nothing, enter always submits). This makes it painful to compose structured messages like the ones Head/Dev routinely need (queue updates, multi-line instructions, code-ish blocks).

## Proposal

Convert the \`#general\` message input in \`src/components/ChatPanel.tsx\` from a single-line \`<input>\` to a **auto-resizing \`<textarea>\`** with the following behavior:

1. **Enter** → send the message (current behavior preserved).
2. **Shift+Enter** → insert a newline, do NOT send.
3. **Auto-grow** — the textarea height follows content, so 1 line stays compact and N lines expand the field.
4. **Max height cap** — stop growing around ~6 lines (or ~120px). Beyond the cap, the textarea becomes internally scrollable. Do NOT let the input push the chat log off-screen.
5. **Min height** — 1 line, matching the current visual footprint so the default empty state doesn't get taller.
6. **Reset on send** — after a successful send, height collapses back to 1 line.
7. **Paste handling** — pasting a multi-line block expands the field up to the cap and scrolls within if longer.

## Implementation notes

- Track content-driven height via a \`useEffect\` that sets \`el.style.height = 'auto'; el.style.height = Math.min(el.scrollHeight, MAX_PX) + 'px'\` on every change.
- \`MAX_PX\` as a constant near the top of the component (suggest 120).
- Preserve the existing IME / composition-event handling so Korean/Japanese input doesn't submit mid-composition on Enter.
- Keep the Send button position stable — anchor it to the bottom of the textarea, not its center, so it doesn't jump as the field grows.

## Out of scope

- Rich text / markdown preview
- File attachments (tracked separately)
- Slash-command autocomplete
- Applying the same treatment to other inputs (Scheduled Trigger widget already uses a textarea; SetupWizard inputs are single-value). This ticket is scoped to the primary chat input only.

## Acceptance

- Typing a long message without Enter keeps the field at 1 line.
- Pressing Shift+Enter inserts a visible newline and expands the field by one line.
- Continued Shift+Enter presses grow the field until ~6 lines, after which it becomes internally scrollable.
- Pressing Enter (no shift) sends the message and resets the field to 1 line.
- Korean/Japanese IME composition does not accidentally submit when pressing Enter to confirm a candidate.
- No layout shift / chat log clipping when the field is at max height.

## Branch

\`task/XXX-chat-input-multiline\`

realproject7:
Fixes #344

## Change
Single-file: `src/components/ChatPanel.tsx`. Replace the `#general` `<input type="text">` with an auto-resizing `<textarea>`.

**Layout**
- `rows={1}` baseline so the empty state matches the old single-line footprint.
- Wrapper flex changes from `items-center` → `items-end` so the Send button anchors to the bottom of the textarea and doesn't jump as the field grows.
- `resize-none` disables the drag handle; `overflow-y-auto`, `leading-snug`, and inline `style={{ maxHeight: 120 }}` match the JS cap so anything taller scrolls internally.

**Height tracking**
- `CHAT_INPUT_MAX_PX = 120` (≈ 6 lines, per ticket).
- `useEffect(..., [input])` does the classic `el.style.height = 'auto'; el.style.height = Math.min(el.scrollHeight, MAX) + 'px'`. The reset-first step is necessary so deleting text shrinks the field; without it `scrollHeight` would pin at the previous height.
- Existing `send()` already clears `input` via `setInput('')`; the effect then collapses the field on the next tick.

**Keyboard handling**
- Shift+Enter inserts a newline — no explicit handling; it's the default textarea behavior once we stop calling `preventDefault` on Shift+Enter.
- Plain Enter sends, but **only** when the IME isn't mid-composition. Korean/Japanese candidate confirmation fires an Enter keydown that must not submit.
- Composition guard uses two signals: an `isComposingRef` flipped by `onCompositionStart` / `onCompositionEnd`, and the native `KeyboardEvent.isComposing` flag on the event itself. Both must be clean before Enter is treated as submit — belt-and-braces because the React synthetic event timing relative to composition flags isn't guaranteed across browsers.
- Tab / slash-menu / @mention autocomplete logic is untouched; the existing `showSlashMenu` branch still catches Enter before it reaches the new submit gate.

**Ref type**
- `inputRef` changed from `useRef<HTMLInputElement>` to `useRef<HTMLTextAreaElement>`. All existing `inputRef.current?.focus()` call sites remain valid.

## Acceptance
- Typing without Enter stays on one line (min height).
- Shift+Enter inserts a newline and grows the field one row at a time.
- Growth is capped at ~6 lines (120px); beyond that the textarea scrolls internally so the chat log never gets pushed off-screen.
- Enter sends, and `setInput('')` + the height effect reset the field to one line.
- Korean/Japanese IME candidate confirmation via Enter doesn't submit the partial buffer.
- Send button stays at the bottom of the field as it grows.

### DIFF
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -210,7 +210,13 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
   }, []);
   const cursorRef = useRef(0);
   const listRef = useRef<HTMLDivElement>(null);
-  const inputRef = useRef<HTMLInputElement>(null);
+  const inputRef = useRef<HTMLTextAreaElement>(null);
+  // #344: IME composition guard — Korean/Japanese candidate
+  // selection ends with an Enter keystroke that must NOT submit
+  // the message. React synthesizes keydown BEFORE the composition
+  // flags flip, so we track a ref alongside the element's own
+  // isComposing flag and check both in the Enter handler.
+  const isComposingRef = useRef(false);
   const shouldAutoScroll = useRef(true);
   const authRetryRef = useRef(0);
 
@@ -290,6 +296,20 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
 
   const [sending, setSending] = useState(false);
   const [sendError, setSendError] = useState<string | null>(null);
+
+  // #344: auto-grow the chat textarea to fit its content up to
+  // ~6 lines, then scroll internally. Reset to 'auto' first so
+  // shrinking after a deletion works — reading scrollHeight on a
+  // fixed-height element would pin the lower bound to the
+  // previous height. Capped at CHAT_INPUT_MAX_PX to match the
+  // inline style; anything taller scrolls inside the textarea.
+  const CHAT_INPUT_MAX_PX = 120;
+  useEffect(() => {
+    const el = inputRef.current;
+    if (!el) return;
+    el.style.height = "auto";
+    el.style.height = Math.min(el.scrollHeight, CHAT_INPUT_MAX_PX) + "px";
+  }, [input]);
 
   // Send message
   const send = () => {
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The patch matches issue #344: it replaces the single-line chat input with a capped auto-expanding textarea, preserves Enter-to-send, allows Shift+Enter newlines, and adds the necessary IME guards so composition confirmation does not submit.

### Findings
- None.

### Decision
Approving because the change is minimal, focused to `src/components/ChatPanel.tsx`, and preserves the existing autocomplete/send behaviors while adding the requested multiline support. Note: `gh pr checks 361` reported no configured checks on this branch, so this approval is based on code review rather than CI.

PR #362, opened 2026-04-09
[#351] Operator Features: two-column layout at lg+

### LINKED ISSUE #351: [ux] Operator Features panel: two-column layout (Trigger left, stack right)
## Problem

The Operator Features panel (\`src/components/OperatorFeaturesPanel.tsx\`) currently stacks every subsection in a single vertical column. The Scheduled Trigger — the most-used control during an overnight run — gets buried alongside the less-frequently-touched controls (Telegram Bridge, Loop Guard, Project History), forcing operators to scroll to reach the trigger message/start button.

## Proposal

Convert the Operator Features panel body into a **two-column layout**:

\`\`\`
┌─────────────────────┬─────────────────────┐
│                     │  Telegram Bridge    │
│                     ├─────────────────────┤
│  Scheduled Trigger  │  Loop Guard         │
│  (full column       ├─────────────────────┤
│   height)           │  Project History    │
│                     │                     │
└─────────────────────┴─────────────────────┘
\`\`\`

### Requirements

1. **Left column** — Scheduled Trigger only. Spans the full vertical height of the Operator Features panel. This is the primary surface during a run; giving it a dedicated column makes the message textarea + Start/Stop button always reachable without scrolling.
2. **Right column** — vertical stack in this exact order, top to bottom:
   1. Telegram Bridge
   2. Loop Guard
   3. Project History
3. **Column widths** — start with 50/50, adjust if the trigger message textarea looks cramped. Trigger column should not shrink below ~280px.
4. **Gap / dividers** — keep the existing section styling (borders, headers). Add a vertical divider between the two columns, matching the existing \`border-border\` utility.
5. **Responsive fallback** — below a narrow breakpoint (~720px panel width), collapse back to the current single-column stacked layout so nothing clips. Use Tailwind responsive classes or a local width observer, whichever matches existing patterns in this file.
6. **No behavior changes** — this is pure layout. Don't touch state, polling, or the child components' internals. Just wrap them in the new grid/flex structure.
7. **Scrolling** — if the right column's content exceeds the panel height, the right column scrolls independently (the Scheduled Trigger on the left should never scroll). Simplest approach: each column is its own flex child with \`overflow-y-auto\` on the right column only.

## Acceptance

- At normal desktop width, Operator Features shows Scheduled Trigger on the left (full column height) and Telegram Bridge → Loop Guard → Project History stacked on the right.
- Scheduled Trigger's message textarea + Start/Stop button are visible without scrolling at typical desktop heights.
- At narrow widths (mobile / cramped split view), layout gracefully collapses to the current single-column stack.
- No regressions to any of the four child widget behaviors (trigger send/stop, telegram bridge start, loop guard pause, history export).

## Branch

\`task/XXX-operator-features-two-column\`

realproject7:
Fixes #351

## Change
Single-file layout-only change to `src/components/OperatorFeaturesPanel.tsx`. No state, polling, or child-widget internals touched.

**Desktop (`lg` breakpoint and up):**
- Body flips from `flex-col` → `flex-col lg:flex-row`.
- **Left column:** `ScheduledTriggerWidget` in a `lg:flex-1 lg:min-w-[280px] lg:min-h-0 lg:overflow-y-auto` wrapper. `min-w-[280px]` per the ticket so the message textarea doesn't collapse at narrow-but-still-lg widths; `overflow-y-auto` so the trigger itself scrolls if it ever grows past panel height.
- **Vertical divider:** `hidden lg:block w-px self-stretch bg-border` between the two columns.
- **Right column:** `TelegramBridgeWidget → LoopGuardWidget → ProjectHistoryWidget` (order per ticket) in a `flex-col lg:flex-1 lg:min-h-0 lg:overflow-y-auto` wrapper — scrolls independently of the left column.
- `lg:overflow-hidden` on the parent body so the inner columns are the scroll containers at lg+ (otherwise the outer box would catch the overflow and defeat the independent-scroll requirement).

**Below `lg`:**
- Reverts to the current single-column stack (`flex-col`, `overflow-auto`), so cramped split-view / mobile layouts don't clip.

**`min-h-0`** is threaded through the wrappers so flexbox correctly clamps the scrollable children at their parent's height (without it, the children would grow to their natural content height and the `overflow-y-auto` would never trigger).

## Acceptance
- Desktop: Scheduled Trigger left, full panel height; Telegram Bridge → Loop Guard → Project History stacked right; right column scrolls independently.
- Scheduled Trigger's textarea + Start/Stop button are visible without scrolling at typical desktop heights (no `overflow-y-auto` clipping on the Trigger card itself because the column wrapper owns the scroll).
- Narrow widths (< `lg`): layout gracefully collapses to the existing single-column stack.
- No behavior changes — all four child widgets render with the same props and the same order is preserved within the right stack.

### DIFF
--- a/src/components/OperatorFeaturesPanel.tsx
+++ b/src/components/OperatorFeaturesPanel.tsx
@@ -34,8 +34,12 @@ export default function OperatorFeaturesPanel({ projectId }: { projectId: string
         {/* Left column: Scheduled Trigger spans full panel height.
             min-w-[280px] at lg+ keeps the message textarea from
             collapsing below a usable width when the panel is
-            narrow-but-still-lg. */}
-        <div className="lg:flex-1 lg:min-w-[280px] lg:min-h-0 lg:overflow-y-auto">
+            narrow-but-still-lg. Per #351, the Trigger must remain
+            the always-reachable primary surface — no overflow-y
+            on this column. The parent lg:overflow-hidden clamps
+            any overshoot visually; the right column is the only
+            independent scroll container. */}
+        <div className="lg:flex-1 lg:min-w-[280px] lg:min-h-0">
           <ScheduledTriggerWidget projectId={projectId} />
         </div>
         {/* Vertical divider between the two columns, only at lg+. */}
--- a/src/components/OperatorFeaturesPanel.tsx
+++ b/src/components/OperatorFeaturesPanel.tsx
@@ -16,16 +16,37 @@ import ProjectHistoryWidget from "./ProjectHistoryWidget";
  * #226: OVERNIGHT-QUEUE.md viewer/editor moved to a compact row at
  * the bottom of the GitHub panel (bottom-left quadrant) — click Edit
  * there to open the modal.
+ *
+ * #351: two-column layout at lg+ widths — Scheduled Trigger gets
+ * the full-height left column (primary surface during an
+ * overnight run so its textarea + Start/Stop button are always
+ * reachable without scrolling), while Telegram Bridge → Loop
+ * Guard → Project History stack in the right column and scroll
+ * independently if the stack exceeds panel height. Below lg the
+ * layout collapses back to the single-column stack so nothing
+ * clips in cramped split-view / mobile.
  */
 export default function OperatorFeaturesPanel({ projectId }: { projectId: string }) {
   return (
     <div className="flex flex-col h-full min-h-0">
       <PanelHeader label="Operator Features" />
-      <div className="flex-1 min-h-0 flex flex-col gap-2 p-2 overflow-auto">
-        <ScheduledTriggerWidget projectId={projectId} />
-        <LoopGuardWidget projectId={projectId} />
-        <ProjectHistoryWidget projectId={projectId} />
-        <TelegramBridgeWidget projectId={projectId} />
+      <div className="flex-1 min-h-0 flex flex-col lg:flex-row gap-2 p-2 overflow-auto lg:overflow-hidden">
+        {/* Left column: Scheduled Trigger spans full panel height.
+            min-w-[280px] at lg+ keeps the message textarea from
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR mostly matches the two-column layout requested in #351, but it changes the left Trigger column into its own scroll container at `lg+`, which conflicts with the ticket's layout requirement.

### Findings
- [medium] Scheduled Trigger column can scroll independently, but the ticket requires only the right column to own overflow
  - File: `src/components/OperatorFeaturesPanel.tsx`
  - Suggestion: Remove `lg:overflow-y-auto` from the left Trigger wrapper and keep the independent scrolling on the right column only. Issue #351 explicitly calls out that the Scheduled Trigger should span the full panel height and that, if content exceeds height, the right column should scroll independently so the Trigger remains the always-reachable primary surface. With the current `lg:overflow-hidden` parent plus `lg:overflow-y-auto` on the left wrapper, the left column can become scrollable too, which is the opposite ownership model from the ticket.

### Decision
Requesting changes because the current overflow assignment does not fully match the intended operator UX for the left Trigger column.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up resolves the overflow ownership issue from my earlier review. The left Scheduled Trigger column no longer owns a `lg:overflow-y-auto` scroll container, so the right column is now the only independent scroll region at `lg+` as required by issue #351.

### Findings
- None.

### Decision
Approving because the updated patch now matches the intended two-column operator layout: Trigger fixed on the left, Telegram Bridge → Loop Guard → Project History stacked and scrollable on the right, with the existing single-column fallback below `lg`. Note: `gh pr checks 362` still reports no configured checks on this branch, so this approval is based on code review of the updated diff rather than CI.

Issue #343, opened 2026-04-09
[feat] Per-agent model & reasoning-effort configuration in dashboard Settings


realproject7:
## Problem

When a Codex or Claude backend hits an upstream 429/overload (\"Selected model is at capacity\"), the only workaround today is for the user to manually edit \`~/.codex/config.toml\` / \`~/.claude/settings.json\` outside QuadWork and relaunch the worktree. There is no way to:

- See which model / reasoning-effort each agent is currently running
- Switch an individual agent (e.g. Reviewer1) to a cheaper or less-loaded tier without touching external config files
- Persist per-project, per-agent model preferences so different projects can use different defaults

Concrete incident: on 2026-04-09, Reviewer1 (Codex, configured upstream as \`gpt-5.4 xhigh\`) was blocked by an \"at capacity\" error mid-batch. The user's quota was fine — it was a provider-side shed. Dropping to \`medium\` would have unblocked it instantly but required out-of-band config surgery.

## Proposal

Add a **Model Settings** section in the dashboard (under Operator Features or a new collapsible panel) with per-agent rows:

| Agent | Backend | Model | Reasoning |
|-------|---------|-------|-----------|
| Head | codex | gpt-5.4 | high ▾ |
| Dev | claude | opus-4.6 | — |
| Reviewer1 | codex | gpt-5.4 | medium ▾ |
| Reviewer2 | claude | sonnet-4.6 | — |

### Requirements

1. **Dropdowns per agent** for model (backend-specific list) and reasoning effort (Codex only, 4 levels: minimal/low/medium/high; no xhigh default — it's the capacity-failure hot spot).
2. **Default**: \`medium\` reasoning effort for Codex agents, not \`high\`/\`xhigh\`. This matches the user's stated preference and avoids the most-overloaded tier.
3. **Persist** per-project in \`~/.quadwork/{project}/config.json\` (or similar existing config file).
4. **Apply on relaunch** — changes take effect when the affected worktree is restarted. Show a \"Restart required\" badge next to agents whose config has drifted from their running session.
5. **One-click restart** button per agent row to pick up the new config.
6. **Backend wiring**: QuadWork already launches \`codex\` / \`claude\` from \`server/index.js\`. Extend that launch to pass the model/reasoning flags (e.g. Codex: \`-c model_reasoning_effort=medium\`, Claude: \`--model <slug>\`) from the per-project config.
[...]

PR #374, opened 2026-04-09
[#368] Auto-reattach terminal panel after PTY restart

### LINKED ISSUE #368: [bug] Agent terminal view doesn't reattach after /api/agents/:project/:agent/restart
## Problem

After clicking **Restart** in the Agent Models widget (from #343), the backend correctly stops and respawns the agent PTY session — the agent itself comes back online and responds in chat — but the **Agent Terminals panel keeps showing the old session's final output** (e.g. \`[session closed: stopped]\`). The operator sees a dead-looking terminal and assumes the restart failed, even though the new session is healthy.

## Evidence (2026-04-09)

1. Operator changed Head's reasoning effort via Agent Models, \"needs restart\" badge appeared, clicked Restart.
2. Agent Terminals → HEAD pane showed:
   \`\`\`
   > Summarize recent commits
   [session closed: stopped]
   gpt-5.4 medium · 100% left · ~/Projects/quadwork-head
   \`\`\`
   — the final frame of the previous session, static.
3. Operator messaged Head via chat: Head responded immediately (\"head is online\"). The new PTY was alive the whole time; only the terminal view was stale.

## Root cause (suspected)

The restart handler at \`server/index.js:1068\` correctly does:
\`\`\`js
await stopAgentSession(key);
const result = await spawnAgentPty(project, agent);
\`\`\`
so the server-side \`agentSessions\` map gets a new entry under the same key with a new \`term\` instance. But the dashboard terminal panel subscribes to PTY output via WebSocket (or SSE stream) bound to the **old** session/term identity. When the old term ends, the client-side stream closes and flips the panel to \"stopped\" — the client doesn't automatically re-subscribe to the new term under the same key.

Two missing pieces:

1. **Server doesn't broadcast a \"session replaced\" event.** The old session's stream closes cleanly, but no signal is sent to the client that a new session has taken the same key.
2. **Client doesn't retry attach on close.** The Agent Terminals component treats \"stream closed\" as terminal state and stops trying.

## Fix

Pick either or both:

### Option A — client auto-reattach on close (lighter)

In the Agent Terminals panel's WebSocket/SSE close handler, before rendering \`[session closed]\`, poll \`/api/projects\` (or \`/api/sessions\`) once to check whether the \`agentSessions\` map actually has a live entry for this key. If it does — and the state is \`running\` — reopen the stream to the new session. If it doesn't, fall through to the existing \"closed\" render.

### Option B — server pushes a \"restarted\" signal (cleaner)

Have the \`/api/agents/:project/:agent/restart\` handler broadcast a WebSocket event like \`{ type: \"session-restarted\", project, agent, newSessionId }\` to any connected clients. The terminal panel listens for this event and transparently swaps its stream subscription without flashing the closed state.

### Option C — both (recommended)

- Server emits the event for the happy path (fast, no polling).
- Client has the auto-reattach fallback so a reconnecting browser that missed the event still recovers on the next close.

## Also

Clear the terminal buffer on reattach — don't leave the old \`[session closed: stopped]\` line above the new prompt, or operators will think the restart produced a weird dual prompt.

## Acceptance

- Clicking Restart in Agent Models results in the Agent Terminals view transitioning from the running session → briefly showing the stop → the new running session, without manual intervention.
- The \"needs restart\" badge in Agent Models clears as before (existing behavior — #343 already handles this).
- No \"ghost\" stopped-session output lingers after the new prompt appears.
- Restart from other entry points (if any — e.g. a restart-all-agents button) gets the same behavior.

## Out of scope

- Redesigning the terminal streaming transport
- Persisting terminal scrollback across restart (follow-up)

## Files

- \`server/index.js:1068\` — restart handler
- \`src/components/AgentTerminalPanel.tsx\` (or wherever the terminal view lives — grep for \`session closed\`)
- Whatever WebSocket / SSE setup already streams PTY output

## Branch

\`task/XXX-agent-terminal-reattach-after-restart\`

## Related

- quadwork#343 (PR #364) — added the Restart button that triggered this
- quadwork#367 — AgentModelsWidget modal refactor (independent, just collocated)

realproject7:
Fixes #368.

## Problem

After clicking Restart in Agent Models (#343), the backend correctly stops+respawns the PTY but the terminal panel shows a frozen `[session closed: stopped]`. The new session is alive in chat; only the terminal view is stale.

## Fix (Option A — client-side auto-reattach)

`src/components/TerminalPanel.tsx`:

- Factored the WebSocket open path into a `connect()` helper so `ws.onclose` can call it again.
- `ws.onclose` now waits 250ms, probes `/api/sessions` for a live entry under this `{project, agent}`, and if one exists calls `term.reset()` and reopens a fresh WebSocket against the new PTY. Writes `[reattached to new session]` so operators see the transition instead of silence.
- `term.reset()` clears the previous session's final frame so operators don't see a ghost `[session closed: stopped]` hanging above the new prompt (acceptance bullet: "No ghost stopped-session output lingers").
- Hard cap `MAX_REATTACH=5` prevents an infinite reconnect loop when the agent is genuinely dead. Counter resets on every successful `ws.onopen`.
- xterm `onData` handler is now registered once up-front against `wsRef.current` rather than inside each connect() call, so reattach cycles don't stack duplicate keystroke senders that'd each fire on every keypress.

## Scope decisions

- **No server-side broadcast** (Option B): kept out. The 250ms probe of `/api/sessions` is sufficient for the happy-path restart since `stopAgentSession` + `spawnAgentPty` are sequential on the server. A transport-level "session-restarted" event would be cleaner but is a larger surface change and the ticket marks Option A as "lighter" — taking it.
- Restart-badge clearing in AgentModelsWidget already works from #343's own `needsRestart` state — no change here.
- No change to the server restart handler. The PTY lifecycle is correct; only the client was failing to follow it.

## Verification

- `npx tsc --noEmit` clean.
- `next build` clean.

## Test plan

- [ ] Click Restart on any agent in Agent Models → terminal briefly shows `[reattached to new session]`, then the new prompt streams in, with no `[session closed]` line above.
- [ ] Stop an agent from the Stop button (not Restart) → terminal still shows `[session closed: stopped]` (no live session → probe returns false → fallthrough).
- [ ] Kill the backend process entirely → after 5 reattach attempts the terminal settles on `[session closed]` instead of looping forever.
- [ ] Type into the reattached terminal — keystrokes reach the new PTY (onData handler preserved across the reconnect).

### DIFF
--- a/src/components/TerminalPanel.tsx
+++ b/src/components/TerminalPanel.tsx
@@ -208,21 +208,38 @@ export default function TerminalPanel({
 
       ws.onclose = async (e) => {
         if (cancelled) return;
-        // #368: delayed probe of /api/sessions. The restart path on
-        // the server does stop→spawn sequentially, so the fresh
-        // session may need a short beat to register before we ask.
-        // A single 250ms retry window is enough in practice.
-        await new Promise((r) => setTimeout(r, 250));
-        if (cancelled) return;
-        if (reattachAttempts < MAX_REATTACH && await sessionIsLive()) {
-          reattachAttempts += 1;
-          // Clear the stale buffer so the previous session's last
-          // frame (including any "stopped" marker) doesn't linger
-          // above the new prompt.
-          term.reset();
-          term.write(`\x1b[38;2;115;115;115m[reattached to new session]\x1b[0m\r\n`);
-          connect();
-          return;
+        // #368: bounded polling probe of /api/sessions. A single
+        // fixed-delay probe is timing-fragile — if the server-side
+        // stop→spawn sequence takes longer than the delay (slow
+        // PTY spawn, busy event loop, AgentChattr re-registration
+        // stall) the probe sees "no session" and the terminal
+        // permanently falls through to [session closed] even
+        // though the new PTY came up a moment later. Instead,
+        // poll every 200ms for up to 2000ms (10 attempts) and
+        // reattach as soon as a live session appears. The loop
+        // bails immediately on any liveness hit, so the happy
+        // path still completes in ~200-400ms.
+        if (reattachAttempts < MAX_REATTACH) {
+          const PROBE_INTERVAL_MS = 200;
+          const PROBE_WINDOW_MS = 2000;
+          const probeStart = Date.now();
+          let live = false;
+          while (Date.now() - probeStart < PROBE_WINDOW_MS) {
+            await new Promise((r) => setTimeout(r, PROBE_INTERVAL_MS));
+            if (cancelled) return;
+            if (await sessionIsLive()) { live = true; break; }
+          }
+          if (cancelled) return;
+          if (live) {
+            reattachAttempts += 1;
+            // Clear the stale buffer so the previous session's
+            // last frame (including any "stopped" marker) doesn't
+            // linger above the new prompt.
+            term.reset();
+            term.write(`\x1b[38;2;115;115;115m[reattached to new session]\x1b[0m\r\n`);
+            connect();
+            return;
+          }
         }
         term.write(`\r\n\x1b[38;2;115;115;115m[session closed: ${e.reason || e.code}]\x1b[0m\r\n`);
       };
--- a/src/components/TerminalPanel.tsx
+++ b/src/components/TerminalPanel.tsx
@@ -117,34 +117,72 @@ export default function TerminalPanel({
     const observer = new ResizeObserver(() => fit());
     observer.observe(containerRef.current);
 
+    // #368: register the xterm → client data handler ONCE up-front
+    // so reattach cycles don't stack duplicate handlers that would
+    // each try to send the same keystroke to the latest ws.
+    term.onData((data) => {
+      const ws = wsRef.current;
+      if (ws && ws.readyState === WebSocket.OPEN) {
+        ws.send(data);
+      }
+    });
+
     // Connect WebSocket — resolve backend URL from config if not provided
     let cancelled = false;
+    let baseUrl: string | null = null;
 
-    (async () => {
-      let base = wsUrl;
-      if (!base) {
-        // In production, WS is same-origin. In dev mode (next dev on :3000),
-        // resolve the backend port from config since WS isn't proxied by Next.js.
-        const wsProto = window.location.protocol === "https:" ? "wss:" : "ws:";
-        try {
-          const res = await fetch("/api/config");
-          if (res.ok) {
-            const cfg = await res.json();
-            const backendPort = cfg.port || 8400;
-            const currentPort = parseInt(window.location.port, 10);
-            if (currentPort && currentPort !== backendPort) {
-              // Dev mode: connect directly to Express backend
-              base = `${wsProto}//${window.location.hostname}:${backendPort}`;
-            } else {
-              base = `${wsProto}//${window.location.host}`;
-            }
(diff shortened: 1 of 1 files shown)


realproject7:
reviewer2: APPROVE PR #374 ([#368] auto-reattach terminal).

- connect() extraction + baseUrl memoization is clean and avoids refetching /api/config on every reattach.
- onData handler correctly moved to one-time registration using wsRef.current.send — this fixes the duplicate-keystroke bug that would otherwise stack on every reconnect cycle.
- ws.onclose → 250ms delay → sessionIsLive() probe → term.reset() + reconnect is the right sequence; the cancelled guard is checked on both sides of the delay so an unmounting panel can't resurrect itself.
- MAX_REATTACH=5 cap with reset on successful onopen correctly bounds a dead-session reconnect storm while still allowing repeated operator-driven restarts during normal use.
- Fall-through to the existing [session closed: …] line preserves current behavior for real disconnects.
- Option B (server-side broadcast) correctly deferred per ticket framing — Option A is the lighter fix and matches the acceptance criteria.
- tsc + next build clean per dev report.

Non-blocking nit: sessionIsLive probes /api/sessions by projectId+agentId equality, which assumes the sessions list always reflects the freshly-spawned PTY within 250ms. If the stop→spawn sequence on the server ever introduces a longer gap, MAX_REATTACH will burn through quickly. Fine for now; worth a follow-up only if we see retry exhaustion in practice.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The overall direction is right, but the current client-side reattach logic is still race-prone: it only checks for the replacement PTY once, after a fixed 250ms delay.

### Findings
- [high] Single-shot 250ms probing can still strand the terminal on a closed session during slower restarts.
  - File: `src/components/TerminalPanel.tsx:206`
  - Suggestion: `ws.onclose` waits 250ms, calls `sessionIsLive()` once, and permanently falls through to `[session closed: ...]` if the new session is not visible yet. That means any restart where `stopAgentSession` + `spawnAgentPty` + session registration takes longer than 250ms will reproduce the original bug. The ticket acceptance is about automatic reattach after restart, so this needs a bounded retry window across multiple probes, or the cleaner server-side restart signal, rather than a single timing-sensitive check.

### Decision
Requesting changes because the fix can still fail nondeterministically under slower PTY respawns, leaving the operator with the same stale closed terminal the ticket is meant to eliminate.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up closes the race from my prior review by replacing the timing-fragile single probe with a bounded reattach polling window, while preserving the original client-side reattach design.

### Findings
- No blocking findings.

### Decision
Approved. This now satisfies #368's acceptance: the terminal gets a real automatic reattach window after restart instead of depending on a single 250ms timing guess, and the stale closed-session frame is still cleared before the new prompt attaches.

realproject7:
reviewer2: RE-APPROVE (commit 867dca5). Bounded 200ms/2000ms polling loop addresses reviewer1's timing concern cleanly — bails on first liveness hit so happy-path latency stays ~200-400ms, but slow stop→spawn sequences now get a full 2s window before falling through. cancelled guard inside the while loop correctly handles unmount mid-probe. MAX_REATTACH=5 outer cap unchanged so a dead agent still can't storm. Good fix.

Issue #355, opened 2026-04-09
[chore] Head agent prompt: correct Scheduled Trigger button name (no 'Send Message and Start Trigger')


realproject7:
## Problem

The Head agent repeatedly instructs the operator to \"click **Send Message and Start Trigger**\" in the Operator Features panel. That button does not exist. The actual widget has a single **Start Trigger** button (\`src/components/ScheduledTriggerWidget.tsx:316\`) — the message textarea is a separate field above it, not combined into one labeled button.

Example (2026-04-09 chat log):

> Queue is ready. To begin, click **Send Message and Start Trigger** in the Operator Features panel (bottom-right). I will assign \`@dev\` as soon as the trigger fires.

First-time operators look for a button that isn't there and get confused.

## Root cause

Head's seed/system prompt or its example scripts reference a button label that either (a) was never accurate, or (b) was accurate in an earlier UI revision that's since been simplified. Either way, Head is now producing hallucinated UI copy.

## Fix

Grep Head's seed files and agent templates for the phrase \`Send Message and Start Trigger\` (and close variants like \`Start Trigger button\`, \`send and start\`) and replace with the correct instruction:

> Queue is ready. To begin, type your trigger message in the **Scheduled Trigger** section of the Operator Features panel (bottom-right) and click **Start Trigger**. I will assign \`@dev\` as soon as the trigger fires.

Check these likely locations:

- \`bin/workspace-seeds/\` — Head AGENTS.md seed
- Any prompt template inside \`server/\` or \`docs/\` that describes the Operator Features panel
- \`CLAUDE.md\` / \`AGENTS.md\` inside the quadwork project's own Head worktree (may be reseeded)

## Acceptance

- Head no longer mentions \"Send Message and Start Trigger\" in chat output
- Head's kick-off message uses the actual button label (\"Start Trigger\") and clearly separates the \"type message\" and \"click button\" steps
- Seed files contain the corrected wording so new project installs don't inherit the old copy
- (Nice-to-have) A grep of the repo for \`Send Message and Start Trigger\` returns zero hits

## Branch

\`task/XXX-head-prompt-correct-trigger-button-name\`

Issue #348, opened 2026-04-09
[infra] Pin AgentChattr to a known commit/tag instead of cloning HEAD


realproject7:
## Problem

QuadWork installs AgentChattr via a plain \`git clone\` with no ref pinning:

\`\`\`js
// bin/quadwork.js:16
const AGENTCHATTR_REPO = \"https://github.com/bcurts/agentchattr.git\";
// bin/quadwork.js:234
run(\`git clone \"\${AGENTCHATTR_REPO}\" \"\${dir}\" 2>&1\`, { timeout: 60000 });
\`\`\`

Every clone grabs the current HEAD of \`bcurts/agentchattr\`'s default branch. Consequences:

- Two users installing QuadWork a week apart can end up on different AgentChattr versions → hard-to-reproduce bugs
- Per-project clones branch off different HEADs depending on when the project was added → the same machine can run mismatched AgentChattr versions across projects
- Upstream breaking changes to AgentChattr immediately affect all new installs with no warning
- No way to roll back to a known-good AgentChattr when upstream regresses

## Proposal

1. **Add a pinned commit/tag constant** in \`bin/quadwork.js\` alongside \`AGENTCHATTR_REPO\`:

   \`\`\`js
   const AGENTCHATTR_REPO = \"https://github.com/bcurts/agentchattr.git\";
   const AGENTCHATTR_PIN = \"<commit-sha-or-tag>\"; // known-good version shipped with this QuadWork release
   \`\`\`

2. **Clone + checkout the pin** in all three install paths (first-run install, \`quadwork install\`, per-project install):

   \`\`\`js
   run(\`git clone \"\${AGENTCHATTR_REPO}\" \"\${dir}\"\`);
   run(\`git -C \"\${dir}\" checkout \${AGENTCHATTR_PIN}\`);
   \`\`\`

3. **Surface the pin** in \`quadwork --version\` / \`quadwork doctor\` output so users can see which AgentChattr commit they're on.

4. **Update bump procedure**: when QuadWork bumps its own version, update \`AGENTCHATTR_PIN\` deliberately after testing against that upstream commit. Document this in \`docs/RELEASING.md\`.

5. **Existing installs**: on \`quadwork upgrade\` / next relaunch, detect if the per-project clone is off-pin and warn (don't auto-rewrite — operator may have local commits). Offer a \`--reset-agentchattr\` flag for forced re-sync.

[...]

Issue #341, opened 2026-04-09
[bug] Batch Progress parser doesn't handle markdown checkbox items in OVERNIGHT-QUEUE.md


realproject7:
## Problem

The Current Batch panel is empty even when \`OVERNIGHT-QUEUE.md\` has a valid Active Batch section, if Head writes items using GitHub-flavored markdown checkboxes:

\`\`\`
- [ ] #338 — Remove home hero...
- [ ] #337 — Stack SERVER...
\`\`\`

\`parseActiveBatch()\` in \`server/routes.js:1194\` uses:

\`\`\`js
const ITEM_LINE_RE = /^\s*(?:[-*]\s+|\d+\.\s+)?\[?#(\d{1,6})\]?\b/;
\`\`\`

This accepts an optional \`[\` *immediately* before \`#\`, so \`- [#338]\` works but \`- [ ] #338\` does not (space between \`[\` and \`#\`). Result: zero issue numbers parsed, panel shows \"Current Batch: (none)\".

## Repro

1. Ask Head to generate a fresh queue — Head naturally uses \`- [ ]\` checkbox syntax for trackable items.
2. Panel stays empty despite Active Batch section being present and well-formed.

## Fix

Update \`ITEM_LINE_RE\` (and the comment block above it) to also skip an optional \`[ ]\` / \`[x]\` / \`[X]\` checkbox token after the list marker, before the \`#N\`:

\`\`\`js
const ITEM_LINE_RE = /^\s*(?:[-*]\s+|\d+\.\s+)?(?:\[[ xX]\]\s+)?\[?#(\d{1,6})\]?\b/;
\`\`\`

Add a unit/integration test case for a checkbox-style Active Batch line.

## Acceptance

- Queue items written as \`- [ ] #338 — ...\` populate the Current Batch panel with the correct batch number + item list.
- Existing supported shapes (\`- #295\`, \`* #295\`, \`[#295]\`, \`1. #295\`) continue to work.
- Prose references (\`Tracking umbrella: #293\`) still rejected.

## Branch

\`task/XXX-batch-progress-checkbox-parser\`

PR #378, opened 2026-04-09
[#366] Pin AgentChattr on a named 'pinned' branch, not detached HEAD

### LINKED ISSUE #366: [infra] AgentChattr pin leaves clones in detached HEAD — mirrors the AC2 worktree rot risk
## Problem

PR #363 (\`[#348] Pin AgentChattr to a known commit + doctor command\`) implements pinning via:

\`\`\`js
const pinResult = run(\`git -C \"\${dir}\" checkout \${AGENTCHATTR_PIN} 2>&1\`, { timeout: 30000 });
\`\`\`

\`git checkout <sha>\` on a bare commit leaves the clone in **detached HEAD** state. This is the exact failure mode that triggered the AC2 diagnostic session that originally produced tickets #341-#355 — stale/misleading git state on agent worktrees caused Claude agents to quote the branch name as their identity (\"I'm on task/279-agents-md-shared-memory, not a t-agent\") and collapse into a polling loop.

Per-project AgentChattr clones aren't agent worktrees so they don't face the same identity-collision bug, but they *do* have other failure modes downstream:

1. **Silent \`git pull\` failures.** Detached HEAD rejects pull-with-ff-only. Any operator or upstream script that tries to update the clone will fail without a clear \"detached HEAD\" explanation.
2. **Re-checkout on restart ambiguity.** If the operator manually tweaks a file in the clone (e.g. config override) and restarts QuadWork, the install path only re-checks-out the pin when the clone is freshly created. An existing detached clone with local edits won't be rebased onto the pin, so edits on a parent commit silently shadow the pin.
3. **Future automation surprises.** Any future feature that wants to \`git pull\` AgentChattr for minor patches (security updates, hotfixes) will hit the detached-HEAD wall and either (a) fail silently or (b) produce a loud warning that looks like corruption.
4. **Mirrors the AC2 diagnostic pattern.** Even though this specific clone isn't an agent worktree, the **class** of bug (git state drifts, something downstream reads the wrong branch/commit, confusing error messages) is exactly what QuadWork was supposed to get us away from.

## Fix

Instead of a bare \`git checkout <sha>\`, pin to a local tracking branch so the clone has a named HEAD:

\`\`\`js
run(\`git -C \"\${dir}\" checkout -b pinned ${AGENTCHATTR_PIN} 2>&1\`, { timeout: 30000 });
\`\`\`

This gives the clone a stable branch name (\`pinned\`) whose HEAD is the pinned commit. Benefits:

- \`git status\` shows \`On branch pinned\` instead of \`HEAD detached at <sha>\`
- \`git pull\` fails with a clear \"this branch has no upstream\" message instead of a vague detached-HEAD warning
- \`quadwork doctor\` can continue to report SHA vs pin without any change (rev-parse HEAD still works)
- Future upgrade flow can fetch upstream and fast-forward \`pinned\` to the new commit when the pin bumps, which is easier to reason about than re-checking-out in a detached state

### Alternative — use a tag

If bcurts/agentchattr ever starts cutting release tags, the pin can be a tag name instead of a SHA. \`git checkout <tag>\` still produces detached HEAD by default, but \`git checkout -b pinned <tag>\` gives the same named-branch benefit.

## Also check

- \`quadwork doctor\` output should explicitly note whether each clone is on the \`pinned\` branch or genuinely off-pin — helps operators distinguish \"pinned but on a named branch\" from \"drift\".
- Document this in \`docs/RELEASING.md\`'s pin-bump procedure so future bumps keep the branch model intact.

## Acceptance

- Fresh \`npx quadwork init\` produces an AgentChattr clone on a named branch (e.g. \`pinned\`), not detached HEAD
- \`git -C <clone> status\` in the global clone reports \`On branch pinned\` with a clean tree
- \`quadwork doctor\` continues to show the expected pin and the actual SHA correctly
- Existing detached-HEAD clones from pre-fix installs are either auto-migrated on next \`quadwork start\` (soft fix) or flagged by \`quadwork doctor\` so the operator knows to re-clone

## Branch

\`task/XXX-agentchattr-pin-named-branch\`

## Related

- quadwork#348 (PR #363) — added the pinning logic
- quadwork#342 — identity injection ticket that traces back to the same AC2 detached-HEAD class of bug

realproject7:
Fixes #366. The #348 pin path used `git checkout <sha>` which leaves clones in detached HEAD — the same git-state-drift class of bug behind the AC2 worktree-rot diagnostic session that triggered tickets #341-#355. Switch to a named local branch so downstream tooling sees a stable HEAD.

### `bin/quadwork.js _installAgentChattrLocked`
- Replace `git checkout <sha>` with `git checkout -B pinned <sha>`. The clone now lands on a local branch called `pinned` whose HEAD is the pinned commit. Capital `-B` is idempotent and force-updates the branch on re-runs.
- New `else` branch on the existing-clone path: if the clone is currently in detached HEAD **and** its HEAD SHA equals `AGENTCHATTR_PIN`, auto-migrate it onto the `pinned` branch in place. Safe because the SHA is unchanged — only a name is being attached. Drift cases (HEAD ≠ pin) and operator-overridden named branches are left alone; `doctor` surfaces them.

### `bin/quadwork.js cmdDoctor`
- New `cloneBranchAt()` helper.
- New status tags: `[OK  ]` (on pin + on `pinned`), `[BR  ]` (on pin but on a non-`pinned` named branch — operator override), `[DETACH]` (on pin but detached — re-run install to migrate), `[DIFF]` (off-pin — re-clone manually).
- Each report line now appends `branch=<name>` so operators can see the HEAD ref alongside the SHA.
- Updated legend.

### `docs/RELEASING.md`
- New "Branch model (#366)" subsection documenting the `-B pinned` invariant, the auto-migration of pre-#366 detached-HEAD clones at the pin, and the AC2-worktree-rot rationale.

## Verification

```
$ node -c bin/quadwork.js
$ node bin/quadwork.js doctor
QuadWork doctor
===============
AgentChattr repo: https://github.com/bcurts/agentchattr.git
Expected pin:     3e71d4267572579e7ffeb83576645f90932c1849

  [BR  ] global: 3e71d4267572579e7ffeb83576645f90932c1849 branch=main (...)
  [BR  ] project:dropcast: 3e71d4267572579e7ffeb83576645f90932c1849 branch=main (...)
  [BR  ] project:quadwork: 3e71d4267572579e7ffeb83576645f90932c1849 branch=main (...)
```

Existing clones on this dev box are on `main` (named, on-pin) so they get `[BR  ]`, which is the correct fall-through state — no destructive auto-action. A fresh `npx quadwork init` from this commit forward will land on `pinned` and report `[OK  ]`.

## Test plan

- [ ] Fresh install: `rm -rf ~/.quadwork/agentchattr && node bin/quadwork.js start` → clone lands on `pinned` branch, doctor reports `[OK  ]`.
- [ ] Detached-HEAD migration: manually `git -C ~/.quadwork/agentchattr checkout <pin-sha>` to detach, re-run `quadwork start` → clone migrates back to `pinned`, doctor reports `[OK  ]`.
- [ ] Drift case: `git -C ~/.quadwork/agentchattr reset --hard HEAD~1` → doctor reports `[DIFF]`, no auto-action.
- [ ] Operator-override case: `git -C ~/.quadwork/agentchattr checkout -B my-feature` (still on pin) → doctor reports `[BR  ]`, no auto-action.

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -247,12 +247,38 @@ function _installAgentChattrLocked(dir, setError) {
     if (cloneResult === null) return setError(`git clone of ${AGENTCHATTR_REPO} into ${dir} failed`);
     if (!fs.existsSync(runPy)) return setError(`Clone completed but run.py missing at ${dir}`);
     // #348: pin to the known-good AgentChattr commit shipped with
-    // this QuadWork release. On failure (commit unreachable /
-    // force-pushed away), fall back to the default branch with a
-    // loud warning instead of hard-failing the install.
-    const pinResult = run(`git -C "${dir}" checkout ${AGENTCHATTR_PIN} 2>&1`, { timeout: 30000 });
+    // this QuadWork release. #366: use `checkout -B pinned <sha>`
+    // so the clone lands on a named local branch ("pinned") whose
+    // HEAD is the pinned commit, rather than detached HEAD. Named
+    // HEAD avoids the AC2 worktree-rot class of bug: `git status`
+    // says `On branch pinned`, downstream tooling that reads the
+    // current branch sees a stable name, and a future `git pull`
+    // fails with a clear "no upstream" message instead of a vague
+    // detached-HEAD warning. -B (capital) is idempotent — it
+    // force-creates/updates the branch so re-runs are no-ops.
+    // On failure (commit unreachable / force-pushed away), fall
+    // back to the default branch with a loud warning instead of
+    // hard-failing the install.
+    const pinResult = run(`git -C "${dir}" checkout -B pinned ${AGENTCHATTR_PIN} 2>&1`, { timeout: 30000 });
     if (pinResult === null) {
       try { console.warn(`[quadwork] WARNING: could not check out AgentChattr pin ${AGENTCHATTR_PIN} at ${dir}; falling back to default branch. The upstream commit may have been force-pushed away — update AGENTCHATTR_PIN in bin/quadwork.js for reproducible installs.`); } catch {}
+    }
+  } else {
+    // #366: existing clone from a pre-fix install. If the clone
+    // is currently in detached HEAD pointing exactly at the pin,
+    // migrate it onto the named `pinned` branch in place. This
+    // is safe because no commits are lost — the SHA is the same,
+    // we're just attaching a name to it. Skip migration if the
+    // clone is on a different SHA (drift) or already on a named
+    // branch (operator may have set up their own work branch on
+    // top); doctor will flag drift cases.
+    const headSha = (run(`git -C "${dir}" rev-parse HEAD 2>&1`) || "").trim();
+    const headRef = (run(`git -C "${dir}" symbolic-ref --quiet HEAD 2>&1`) || "").trim();
+    if (headSha === AGENTCHATTR_PIN && !headRef) {
+      const migrateResult = run(`git -C "${dir}" checkout -B pinned ${AGENTCHATTR_PIN} 2>&1`, { timeout: 30000 });
+      if (migrateResult === null) {
+        try { console.warn(`[quadwork] WARNING: could not migrate ${dir} from detached HEAD to the 'pinned' branch.`); } catch {}
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


realproject7:
reviewer2: APPROVE PR #378 ([#366] pin AgentChattr on named 'pinned' branch).

- `checkout -B pinned <sha>` is the right primitive: capital B is idempotent and force-updates, so re-runs on a fresh clone are clean no-ops.
- Auto-migration branch for existing pre-fix clones is safe: guarded on both `headSha === AGENTCHATTR_PIN` AND `!headRef` (detached), so drift cases and operator work branches are explicitly left alone.
- The headRef check via `symbolic-ref --quiet HEAD` is the correct way to detect detached HEAD — empty string when detached, ref path otherwise. Stripping `refs/heads/` prefix in cloneBranchAt() is consistent.
- doctor's four-state tag expansion (OK / BR / DETACH / DIFF) matches the possible clone states and surfaces DETACH as actionable ("re-run install to migrate") without auto-mutating unexpectedly.
- Fallback for unreachable pin still works — the existing warn-and-continue path is untouched.
- RELEASING.md documents the -B pinned invariant for future pin bumps, which is the important maintenance note.
- Dev verified doctor output on a real dev box showing 3 on-pin-but-main clones correctly flagged [BR  ] — good sanity check that non-destructive migration rules hold.

project7-interns (approved):
## Verdict: APPROVE

### Summary
This PR fixes the detached-HEAD pinning problem with the intended low-risk branch model: fresh AgentChattr clones land on a named `pinned` branch, and existing on-pin detached clones are migrated in place.

### Findings
- No blocking findings.

### Decision
Approved. The implementation matches #366's acceptance: fresh installs avoid detached HEAD, pre-fix detached clones at the pin are auto-migrated safely, and `quadwork doctor` now reports branch state clearly enough to distinguish pinned-branch health from operator overrides or drift.

PR #382, opened 2026-04-09
[#236] Chat WS send: await broadcast echo, return stored message

### LINKED ISSUE #236: [bug] Chat WebSocket send — fire-and-forget ack, lost response body, attachment regression
Follow-up to PR #231 (#230 hotfix). The WebSocket-based chat send approach is correct at the root-cause level but has three minor issues that should be addressed in a polish pass.

## Issue 1 — Fire-and-forget ack

Current code resolves 250ms after \`ws.send()\` with no real acknowledgment from AgentChattr. If the server validates the payload and rejects it (e.g. wrong shape, banned word, internal error), the dashboard reports success and the user thinks their message went through.

**Fix:** Listen for the echoed broadcast frame on the same ws connection (filter by a client-generated message id) and resolve the promise only after seeing the echo. Surface server-side errors as proper 5xx responses.

## Issue 2 — Lost response body

Old \`/api/send\` proxy returned AgentChattr's response body directly. New code always returns \`{ok: true}\`. Any caller that consumed the response (e.g. for the message id, timestamp, or echo) will get less data than before.

**Fix:** Capture the broadcast echo and return its message object so clients have full parity with the old behavior.

## Issue 3 — Attachment regression

PR #231 forwards \`attachments\` through the ws frame, but it's unverified that AgentChattr's ws \`type:\"message\"\` handler accepts attachments. The old \`/api/send\` route may have been the only path that processed attachment uploads.

**Fix:** Quick repro — send a file via the chat panel and check the AgentChattr channel for the file. If it doesn't appear, either route attachment uploads back through a separate POST endpoint, or fix AgentChattr's ws handler to accept them.

## Files to investigate

- \`server/index.js\` — \`/api/chat\` handler (the new ws-based version from #231)
- AgentChattr source: \`~/.quadwork/{project}/agentchattr/app.py\` — find the ws message handler and see what fields it accepts on \`type:\"message\"\` frames

## Acceptance

- ✅ Chat send returns the actual server response, not a fake \`{ok: true}\`
- ✅ Validation failures from AgentChattr surface as 5xx, not silent success
- ✅ File attachments still work via the chat panel
- ✅ Add a regression test or at least manual verification steps in the PR description

realproject7:
## Summary
- `sendViaWebSocket` now stays subscribed after sending and resolves only when AgentChattr broadcasts our message back (matched on sender + text + channel + reply_to, with a timestamp gate against history replay).
- Returns `{ok, message}` so `/api/chat` regains parity with the old `/api/send` response body — callers get the server-assigned id/timestamp back.
- Premature ws close without an echo now rejects → `/api/chat` returns 502 instead of a silent `{ok:true}`.
- 4003 (bad token) path unchanged — still `EAGENTCHATTR_401` → resync + single retry.
- Attachments already pass through AC's ws `type:"message"` handler (verified in `agentchattr/app.py:1091-1123`); the new test case 1 round-trips an attachments payload to prove it.

## Changes
- `server/routes.js` — rewrite `sendViaWebSocket`; `/api/chat` now returns the echoed message object on success.
- `server/routes.chatWsSend.test.js` — new regression fixture (5 cases) using a fake AC ws server.

## Test plan
- [x] `node server/routes.chatWsSend.test.js` — 5 new cases pass.
- [x] All existing server/*.test.js suites pass (queue-watcher, batchProgress, parseActiveBatch, telegramBridge).
- [ ] Reviewers: confirm the (sender,text,channel,reply_to) echo match is safe in practice (could two near-simultaneous identical messages collide? current behavior resolves on the first matching echo after send — acceptable because the echo IS our own message round-tripped).
- [ ] Reviewers: verify the 1500ms timestamp grace doesn't reject legitimate echoes under load.

Fixes #236

@reviewer1 @reviewer2 please review.

### DIFF
--- a/server/routes.chatWsSend.test.js
+++ b/server/routes.chatWsSend.test.js
@@ -36,9 +36,16 @@ function startFakeAc({ historyBeforeAck = [], rejectWithCode = null, dropBeforeA
     wss.handleUpgrade(req, socket, head, (ws) => {
       // Simulate history replay on connect — these must NOT satisfy
       // the ack, even if they match (sender,text,channel).
+      let historyMaxId = 0;
       for (const msg of historyBeforeAck) {
         ws.send(JSON.stringify({ type: "message", data: msg }));
+        if (typeof msg.id === "number" && msg.id > historyMaxId) historyMaxId = msg.id;
       }
+      // Mirror AC: emit one `type:"status"` frame after history so the
+      // client knows the replay is done. See agentchattr/app.py
+      // `broadcast_status()` call in the /ws handler (line ~1082).
+      ws.send(JSON.stringify({ type: "status", data: { ready: true } }));
+      let nextId = Math.max(9000, historyMaxId + 1);
       ws.on("message", (raw) => {
         const frame = JSON.parse(raw.toString());
         if (frame.type !== "message") return;
@@ -48,7 +55,7 @@ function startFakeAc({ historyBeforeAck = [], rejectWithCode = null, dropBeforeA
         }
         // Mirror store.add: assign id + timestamp, rebroadcast.
         const echoed = {
-          id: 9001,
+          id: nextId++,
           sender: frame.sender,
           text: frame.text,
           channel: frame.channel || "general",
@@ -80,7 +87,8 @@ function startFakeAc({ historyBeforeAck = [], rejectWithCode = null, dropBeforeA
     });
     assert.equal(result.ok, true);
     assert.ok(result.message, "expected echoed message object");
-    assert.equal(result.message.id, 9001);
+    assert.equal(typeof result.message.id, "number");
+    assert.ok(result.message.id >= 9000);
     assert.equal(result.message.text, "hello from test");
     assert.equal(result.message.sender, "user");
     assert.equal(result.message.attachments.length, 1);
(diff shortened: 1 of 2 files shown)


realproject7:
**reviewer2: APPROVE** ✅

Verified:
1. Echo matching on (sender, text, channel, reply_to) + timestamp gate (sentAt - 1500ms) correctly ignores history replay on connect. Case 3 proves it.
2. `sentAt` is only set after `ws.send` succeeds, so the message listener's `!sentAt` guard prevents any pre-send frame from being mistaken for an ack.
3. Premature close path now rejects with a descriptive error instead of silent `{ok:true}` — fixes the core #236 regression. Case 4 covers it.
4. 4003 path preserved with `EAGENTCHATTR_401` → resync + single retry logic intact; retry now also returns `message` for body parity. Case 5 covers it.
5. Attachments round-trip through AC's `type:"message"` envelope — case 1 proves the echo carries them back unchanged.
6. Collision concern (two identical near-simultaneous messages): acceptable. Each /api/chat call owns its own ws connection, resolves on the first matching post-send echo; worst case it resolves on a sibling's echo with the same (sender,text,channel,reply_to), which is functionally indistinguishable anyway.
7. 1500ms timestamp grace: generous enough for load; AC assigns timestamp at store.add, so clock skew within 1.5s is the only risk — fine.

No blockers. Approving.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The WebSocket echo-waiting direction is correct, and the attachment/body regression coverage is useful, but the current ack matcher is still unsafe for duplicate sends. A recent historical message can satisfy the matcher before the newly-sent message is stored, so `/api/chat` can still report success for the wrong record.

### Findings
- [high] The history-replay guard still accepts recent duplicate messages, so the send ack can resolve against an older message instead of the new one.
  - File: `server/routes.js:190`
  - Suggestion: `sendViaWebSocket` needs a true correlation mechanism rather than `(sender,text,channel,reply_to)` plus a 1500ms timestamp grace. Today AC replays history on connect before entering the receive loop, and the filter only rejects echoes older than `sentAt - 1500`. If the user sends the same text twice within that 1.5s window, the replayed older message will pass the filter and resolve the promise before the new send is stored. Fix this by adding an end-to-end correlation id that survives the round trip, or by changing the protocol so the server emits an explicit per-send ack.

### Decision
Request changes. Issue #236 explicitly called out matching the echoed broadcast to the specific client send; the current heuristic still has a false-positive path for rapid duplicate messages, so the fire-and-forget correctness bug is not fully closed. Note: `gh pr checks 382` reports no configured checks on the branch, so this request-changes verdict is based on diff review.

realproject7:
**reviewer2: APPROVE** (round 2) ✅

Round-1 blocker resolved. Verified:
1. History boundary is now definitional: `doSend` is gated on either the first `type:"status"` frame OR a 750ms fallback. Any `type:"message"` frame arriving before `sent=true` is classified as history and only contributes to `maxHistoryId`.
2. Correlation baseline (`maxIdAtSend`) is snapshotted at send time, and the post-send matcher requires `d.id > maxIdAtSend` AND the tuple match — strictly monotonic AC ids make this race-free.
3. Case 3 now stages an identical historical message with `id=42` and a `timestamp: Date.now()`; the live echo uses `id >= 9001`, so the assertion `result.message.id > 42` proves the old heuristic path is dead.
4. 750ms fallback correctly uses `maxHistoryId` as baseline if status never arrives — safe because any frame seen in that window has already bumped `maxHistoryId`.
5. `giveUp` 4s timeout still wraps the whole flow, so a stuck "waiting for status" connection won't hang.
6. Premature close / 4003 / attachments paths unchanged from round 1.

No blockers. Approving.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The round-2 change fixes the false-positive ack path from the first review. `sendViaWebSocket` now waits until AgentChattr finishes history replay, records a message-id baseline at send time, and only accepts a matching echoed message whose id is strictly newer than that baseline.

### Findings
- None.

### Decision
Approve. Waiting for the post-history `type:"status"` frame plus the `id > baseline` check gives this request a real correlation boundary without requiring a protocol change, and the new regression case with a recent identical historical message covers the exact duplicate-send race that blocked round 1. `/api/chat` also now returns the stored message body again and still preserves the 4003 resync path. Note: `gh pr checks 382` reports no configured checks on the branch, so this is a diff-based code review approval.

PR #385, opened 2026-04-09
[#383] Telegram Bridge: fix 4 layered config bugs

### LINKED ISSUE #383: [bug] Telegram Bridge end-to-end broken: wrong URL, wrong TOML schema, missing AC agent declaration (4 layered bugs)
## Context

After quadwork#380 (PR #381) fixed the dedicated venv + interpreter selection, clicking Start on the Telegram Bridge widget successfully spawns the bridge subprocess — but messages never flow in either direction. Investigating the bridge log and the actual TOML / AC config surfaced **four separate bugs layered on top of each other**, all in QuadWork's bridge configuration code path. None of them were caught by #380 because that ticket was scoped to interpreter selection only.

The bridge process itself (upstream \`realproject7/agentchattr-telegram\`) is behaving correctly — every failure below is on QuadWork's side.

### Location
\`server/routes.js:2505\`

\`\`\`js
return {
  bot_token: resolveToken(project.telegram.bot_token || \"\"),
  chat_id: project.telegram.chat_id || \"\",
  agentchattr_url: cfg.agentchattr_url || \"http://127.0.0.1:8300\",
};
\`\`\`

It reads \`cfg.agentchattr_url\` (the **global default** from \`DEFAULT_CONFIG\`, which is \`http://127.0.0.1:8300\`) instead of the project's own \`project.agentchattr_url\` (which is \`http://127.0.0.1:8301\` or similar for every project after the first one).

### Observable symptom
Bridge logs show \`Failed to register with AgentChattr: 400 … http://127.0.0.1:8300/api/register\` on the \`quadwork\` project — it's trying to register with DropCast's AC server (or nothing, if DropCast is stopped) instead of the quadwork project's own AC.

### Fix
\`\`\`js
agentchattr_url: project.agentchattr_url || cfg.agentchattr_url || \"http://127.0.0.1:8300\",
\`\`\`

Prefer per-project, fall back to the global default only for legacy single-project installs.

---

### Location
\`server/routes.js:2664\`

QuadWork writes:
\`\`\`toml
[telegram]
bot_token = \"...\"
chat_id = \"...\"

[agentchattr]
url = \"http://127.0.0.1:8301\"
\`\`\`

But \`~/.quadwork/agentchattr-telegram/telegram_bridge.py\` (upstream bridge script, lines 60-75 of its \`load_config\`) only reads from the \`[telegram]\` section:

\`\`\`python
tg = toml_data.get(\"telegram\", {})
…
if tg.get(\"agentchattr_url\"):
    config[\"agentchattr_url\"] = tg[\"agentchattr_url\"]
\`\`\`

There is no reader for a separate \`[agentchattr]\` section. The bridge silently falls through to its hardcoded default \`DEFAULT_AGENTCHATTR_URL = \"http://127.0.0.1:8300\"\` (\`telegram_bridge.py:32\`), which is why Bug 1 alone didn't produce the expected symptom — the wrong URL came from the bridge's own default, not from QuadWork's \`cfg.agentchattr_url\`.

### Observable symptom
Bridge hits \`127.0.0.1:8300\` no matter what URL QuadWork writes into the \`[agentchattr]\` section.

### Fix
Change the TOML writer to put the URL inside the \`[telegram]\` section as \`agentchattr_url =\`:

\`\`\`js
const tomlContent =
  `[telegram]\\n` +
  `bot_token = \"${tg.bot_token}\"\\n` +
  `chat_id = \"${tg.chat_id}\"\\n` +
  `agentchattr_url = \"${tg.agentchattr_url}\"\\n`;
\`\`\`

Drop the separate \`[agentchattr]\` section entirely. Verify against \`telegram_bridge.py:60-75\` for the exact key names it accepts.

---

### Location
\`bin/quadwork.js\` / \`server/routes.js\` (wherever the per-project AC \`config.toml\` template is written during setup) — needs to be audited. Relevant files that currently declare the 4 default agents:

- \`templates/config.toml\` — the base template
- \`~/.quadwork/<project>/agentchattr/config.toml\` — rendered per project

### What's missing
The per-project AC config only declares:
\`\`\`toml
[agents.head]
[agents.reviewer1]
[agents.reviewer2]
[agents.dev]
\`\`\`

AC's \`registry.register(base, label)\` rejects any base that isn't pre-declared in the config with \`400 unknown base: <name>\`. The bridge tries to register as \`base=\"telegram-bridge\"\`, which isn't declared → 400 → bridge enters the \"re-register next cycle\" loop forever → messages never forward.

For comparison, AC2's \`/Users/cho/Projects/agent-os/agentchattr2/config.toml\` has:

\`\`\`toml
[agents.telegram-bridge]
label = \"Telegram Bridge\"
\`\`\`

### Observable symptom
Bridge logs show \`WARNING: Failed to register with AgentChattr: 400 Client Error\` followed by \`WARNING: AgentChattr token rejected (HTTP 403) — re-registering\` every 2-3 seconds, indefinitely. Bridge is running but can't authenticate because it can't register.

### Fix
Add \`[agents.telegram-bridge]\` with \`label = \"Telegram Bridge\"\` to:

1. \`templates/config.toml\` (so new projects get it automatically)
2. A migration step that appends it to any existing per-project \`~/.quadwork/<id>/agentchattr/config.toml\` that's missing the section (idempotent — skip if already present)

The migration can piggyback on the existing Install Bridge flow: when Install Bridge runs, check the per-project AC config and append the agent declaration if missing. Since the AC server must be restarted for the new agent slug to take effect, either:

- (a) Prompt the operator to click SERVER → Restart after Install Bridge if the AC config was just modified, or
- (b) Auto-restart the AC server from the Install Bridge handler after patching config.toml (intrusive — affects all 4 running agents).

Option (a) is safer; Option (b) is smoother UX but risks interrupting a running batch. Recommend (a) with a clear UI message.

---

### Location
\`~/.quadwork/agentchattr-telegram/telegram_bridge.py:78\` (upstream bridge behavior, not QuadWork code)

The bridge's config loader treats env vars as **higher precedence than TOML values**:

\`\`\`python
# Env var override map
\"TELEGRAM_BOT_TOKEN\": \"telegram_bot_token\",
\`\`\`

So if QuadWork spawns the bridge with any inherited environment that has \`TELEGRAM_BOT_TOKEN\` set (common on operator machines that also run AC1/AC2 with this env var exported in their shell rc or a different bot), the bridge uses the **env var's token** instead of the TOML's token. This produces a bridge running as the wrong bot — the log shows \`Telegram bot: @<wrong_bot>\` even though the QuadWork widget shows the expected \`@quadwork_bot\` from the saved config.

### Observable symptom (reproduced in diagnostic)
When the bridge was manually respawned from a shell that inherited an AC2 \`TELEGRAM_BOT_TOKEN\`, the log said \`Telegram bot: @p7_agent_bot\` instead of the expected \`@quadwork_bot\`. Respawning with \`env -i\` to clear the inherited env produced the correct bot identity.

### Fix
QuadWork's Start handler (\`server/routes.js:2622\`) already calls \`spawn()\` without an \`env\` option, which means the child inherits the Node server process env wholesale. The Node server itself inherits from the shell that ran \`quadwork start\`. If that shell has \`TELEGRAM_BOT_TOKEN\` exported, the bridge picks it up instead of the TOML.

Two possible fixes:

- **Preferred:** explicitly scrub the Telegram-related env vars from the spawn child's environment by passing a filtered \`env\`:
  \`\`\`js
  const childEnv = { ...process.env };
  delete childEnv.TELEGRAM_BOT_TOKEN;
  delete childEnv.TELEGRAM_CHAT_ID;
  delete childEnv.AGENTCHATTR_URL;
  const child = spawn(venvPython, [bridgeScript, \"--config\", tomlPath], {
    detached: true,
    stdio: [\"ignore\", outFd, errFd],
    env: childEnv,
  });
  \`\`\`
  This makes the TOML the single source of truth for the bridge's config regardless of the ambient shell env.

- **Alternative:** patch the upstream bridge to reverse the precedence (TOML > env). Requires an upstream PR to \`realproject7/agentchattr-telegram\`, so slower to land.

**Recommend the preferred fix** — scrub env vars in the spawn child. Zero upstream dependency.

---

## Consolidated acceptance

- Creating a new project via the setup wizard produces a per-project AC config with \`[agents.telegram-bridge]\` declared.
- Existing projects (DropCast, quadwork) get the same section appended idempotently via the Install Bridge migration path.
- Clicking Install Bridge after this fix ships is safe to re-run — no duplicate sections, no config corruption.
- After AC server restart (post-migration), clicking Start on the Telegram Bridge widget:
  - Writes a TOML with the URL **inside** the \`[telegram]\` section as \`agentchattr_url\` (not in a separate \`[agentchattr]\` section)
  - Uses the project's own \`agentchattr_url\` (e.g. \`:8301\` for quadwork), not the global default \`:8300\`
  - Spawns the bridge with a scrubbed env that has no \`TELEGRAM_BOT_TOKEN\` / \`TELEGRAM_CHAT_ID\` / \`AGENTCHATTR_URL\` inherited from the shell
  - Bridge successfully registers (\`POST /api/register\` → 200) and enters the normal poll loop
  - Messages sent to the Telegram bot appear in the QuadWork AgentChattr chat under the \`telegram-bridge\` sender
  - Messages posted to the AgentChattr chat by any agent are forwarded to the Telegram chat
- Bridge log file at \`~/.quadwork/telegram-bridge-<project>.log\` contains no \`WARNING: Failed to register\` or \`WARNING: AgentChattr token rejected\` lines during normal operation.
- No regression to the venv + interpreter fix from #380.
- Existing error-surface pipeline from #353 / #372 still catches real crash modes unchanged.

## Test coverage

Extend \`server/routes.telegramBridge.test.js\`:

- **Case 12:** \`getProjectTelegram()\` with a project whose \`agentchattr_url\` is \`:8301\` returns \`:8301\`, not the global default \`:8300\`.
- **Case 13:** The TOML writer produces a string with \`agentchattr_url\` inside the \`[telegram]\` section, no separate \`[agentchattr]\` section. Assert absence of \`\\n[agentchattr]\\n\`.
- **Case 14:** Install Bridge's config-patching step is idempotent — running it against a config that already has \`[agents.telegram-bridge]\` does not duplicate the section.
- **Case 15:** Spawn child env has \`TELEGRAM_BOT_TOKEN\` / \`TELEGRAM_CHAT_ID\` / \`AGENTCHATTR_URL\` stripped even if the parent env has them set.

## Files to change

- \`server/routes.js\` — \`getProjectTelegram()\` (Bug 1), TOML writer (Bug 2), Install Bridge config-patch migration (Bug 3), Start handler env scrub (Bug 4)
- \`templates/config.toml\` — add \`[agents.telegram-bridge]\` so new projects get it (Bug 3)
- \`server/routes.telegramBridge.test.js\` — 4 new test cases
- \`bin/quadwork.js\` — if the CLI wizard writes per-project AC config directly, mirror the template change

## Operator-side current state (do NOT auto-revert in the fix)

The operator manually patched their \`~/.quadwork/telegram-quadwork.toml\` and \`~/.quadwork/quadwork/agentchattr/config.toml\` during diagnosis:

- \`~/.quadwork/telegram-quadwork.toml\` now has \`agentchattr_url = \"http://127.0.0.1:8301\"\` inline in \`[telegram]\` section (correct format)
- \`~/.quadwork/quadwork/agentchattr/config.toml\` now has \`[agents.telegram-bridge]\` appended at the bottom

The fix should:
- Be idempotent — NOT undo these manual patches
- Recognize the hand-patched config as already-correct during the Install Bridge migration
- Warn loudly if the existing config differs from what the new writer would produce in a way that suggests a manual patch (don't clobber silently)

## Out of scope

- Windows portability for the bridge (same note as #380)
- Bridge script parity between AC2's copy and QuadWork's clone
- Upstream PR to \`realproject7/agentchattr-telegram\` to reverse the env/TOML precedence (Bug 4 alternative) — leave the upstream behavior alone
- Re-designing the per-project AC config template (out of this ticket's scope)

## Branch

\`task/XXX-telegram-bridge-config-four-bugs\`

## Related

- quadwork#380 (PR #381) — venv + interpreter fix that exposed these follow-on bugs
- quadwork#379 — original research ticket (closed as research-complete)
- quadwork#372 (PR #373) — error persistence surface that let us see these bugs at all
- quadwork#353 (PR #359) — earlier crash-visibility work
- AC2 reference config: \`/Users/cho/Projects/agent-os/agentchattr2/config.toml\` (has the correct \`[agents.telegram-bridge]\` declaration)
- Bridge config reader: \`~/.quadwork/agentchattr-telegram/telegram_bridge.py:60-90\`

realproject7:
## Summary
- Fixes the four layered bugs in #383 that kept the Telegram Bridge from exchanging messages end-to-end after the #381 venv fix.
- Bug 1: `getProjectTelegram()` now prefers `project.agentchattr_url` (per-project port) over the global default.
- Bug 2: Both bridge TOML writers (`server/routes.js` start handler + `bin/quadwork.js` setup wizard) now emit `agentchattr_url` inside `[telegram]`. The upstream bridge's `load_config` never read a separate `[agentchattr]` section.
- Bug 3: `templates/config.toml` declares `[agents.telegram-bridge]`, and the Install Bridge handler runs an idempotent migration that appends the section to every existing per-project AC `config.toml`. Hand-patched configs are recognized as already-correct and not clobbered. Operator must click SERVER → Restart after Install Bridge for AC to pick up the new slug.
- Bug 4: Start handler scrubs `TELEGRAM_BOT_TOKEN` / `TELEGRAM_CHAT_ID` / `AGENTCHATTR_URL` from the spawn child env so an ambient shell can't override the TOML.
- Extracts four pure helpers (`resolveProjectAgentchattrUrl`, `buildTelegramBridgeToml`, `patchAgentchattrConfigForTelegramBridge`, `buildTelegramBridgeSpawnEnv`) and exports them for unit tests.

## Test plan
- [x] `node server/routes.telegramBridge.test.js` — 15 cases pass (4 new: 12-15 cover all four bugs).
- [x] `node server/routes.batchProgress.test.js` — no regression.
- [x] `node server/routes.parseActiveBatch.test.js` — no regression.
- [x] `node server/routes.chatWsSend.test.js` — no regression.
- [x] `node -e \"require('./server/routes')\"` — module loads clean.
- [ ] Manual: click Install Bridge on a machine with hand-patched configs, verify no duplicate sections and no clobbering.
- [ ] Manual: click Start on quadwork project, verify bridge registers 200 against :8301 and forwards messages both ways.

Fixes #383

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1122,9 +1122,18 @@ bridge_sender = "telegram-bridge"
           const bridgeTomlContent = `[telegram]\nbot_token = "${botToken}"\nchat_id = "${chatId}"\nagentchattr_url = "${projectChattrUrl}"\n`;
           fs.writeFileSync(bridgeToml, bridgeTomlContent, { mode: 0o600 });
           fs.chmodSync(bridgeToml, 0o600);
+          // #383 Bug 4: scrub TELEGRAM_*/AGENTCHATTR_URL from the
+          // child env so an ambient shell that exported a different
+          // bot's token can't silently override the TOML we just
+          // wrote. Same fix as server/routes.js Start handler.
+          const bridgeEnv = { ...process.env };
+          delete bridgeEnv.TELEGRAM_BOT_TOKEN;
+          delete bridgeEnv.TELEGRAM_CHAT_ID;
+          delete bridgeEnv.AGENTCHATTR_URL;
           const bridgeProc = spawn("python3", [bridgeScript, "--config", bridgeToml], {
             stdio: "ignore",
             detached: true,
+            env: bridgeEnv,
           });
           bridgeProc.unref();
           if (bridgeProc.pid) {
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1116,7 +1116,10 @@ bridge_sender = "telegram-bridge"
         if (fs.existsSync(bridgeScript)) {
           log("Starting Telegram bridge...");
           const bridgeToml = path.join(CONFIG_DIR, `telegram-${setup.projectName}.toml`);
-          const bridgeTomlContent = `[telegram]\nbot_token = "${botToken}"\nchat_id = "${chatId}"\n\n[agentchattr]\nurl = "${projectChattrUrl}"\n`;
+          // #383 Bug 2: the bridge only reads agentchattr_url from
+          // inside [telegram]. A separate [agentchattr] section is
+          // silently ignored and the bridge falls back to :8300.
+          const bridgeTomlContent = `[telegram]\nbot_token = "${botToken}"\nchat_id = "${chatId}"\nagentchattr_url = "${projectChattrUrl}"\n`;
           fs.writeFileSync(bridgeToml, bridgeTomlContent, { mode: 0o600 });
           fs.chmodSync(bridgeToml, 0o600);
           const bridgeProc = spawn("python3", [bridgeScript, "--config", bridgeToml], {
--- a/server/routes.js
+++ b/server/routes.js
@@ -2385,6 +2385,63 @@ function telegramConfigToml(projectId) {
   return path.join(CONFIG_DIR, `telegram-${projectId}.toml`);
 }
 
+// #383: path to a project's AgentChattr config.toml. The install
+// handler patches this file to declare the `telegram-bridge` agent
+// so AC's registry accepts the bridge's register call.
+function projectAgentchattrConfigPath(projectId) {
+  return path.join(CONFIG_DIR, projectId, "agentchattr", "config.toml");
+}
+
+// #383 Bug 1: prefer the per-project agentchattr_url. Every project
+// after the first uses a distinct port (8301, 8302, ...), so reading
+// the global default silently routed bridge traffic to the wrong AC
+// instance.
+function resolveProjectAgentchattrUrl(cfg, project) {
+  return (
+    (project && project.agentchattr_url) ||
+    (cfg && cfg.agentchattr_url) ||
+    "http://127.0.0.1:8300"
+  );
+}
+
+// #383 Bug 2: the upstream bridge only reads `agentchattr_url` from
+// inside `[telegram]`. A separate `[agentchattr]` section is silently
+// ignored and the bridge falls back to its hardcoded :8300 default.
+function buildTelegramBridgeToml(tg) {
+  return (
+    `[telegram]\n` +
+    `bot_token = "${tg.bot_token}"\n` +
+    `chat_id = "${tg.chat_id}"\n` +
+    `agentchattr_url = "${tg.agentchattr_url}"\n`
+  );
+}
+
+// #383 Bug 3: AC's registry rejects any base name not pre-declared
+// in config.toml with `400 unknown base`. The bridge registers as
+// `telegram-bridge`, so every per-project AC config must declare it.
+// Idempotent: only appends if the section is not already present.
+function patchAgentchattrConfigForTelegramBridge(tomlText) {
+  if (/^\[agents\.telegram-bridge\]\s*$/m.test(tomlText)) {
+    return { text: tomlText, changed: false };
+  }
+  const sep = tomlText.length === 0 || tomlText.endsWith("\n") ? "" : "\n";
+  const block = `\n[agents.telegram-bridge]\nlabel = "Telegram Bridge"\n`;
+  return { text: tomlText + sep + block, changed: true };
+}
+
+// #383 Bug 4: the upstream bridge treats env vars as higher
+// precedence than TOML values. If the parent shell exported
+// TELEGRAM_BOT_TOKEN for a different bot, the bridge silently ran
+// as the wrong identity. Scrub those keys from the child's env so
+// the TOML is the single source of truth.
+function buildTelegramBridgeSpawnEnv(parentEnv) {
+  const env = { ...parentEnv };
+  delete env.TELEGRAM_BOT_TOKEN;
+  delete env.TELEGRAM_CHAT_ID;
+  delete env.AGENTCHATTR_URL;
+  return env;
+}
+
 // #353: per-project log file for the bridge subprocess. The start
 // handler redirects stdout + stderr here so crashes (ImportError,
 // config parse, auth failure) are recoverable instead of
@@ -2502,7 +2559,8 @@ function getProjectTelegram(projectId) {
     return {
       bot_token: resolveToken(project.telegram.bot_token || ""),
       chat_id: project.telegram.chat_id || "",
-      agentchattr_url: cfg.agentchattr_url || "http://127.0.0.1:8300",
+      // #383 Bug 1: prefer per-project URL over the global default.
+      agentchattr_url: resolveProjectAgentchattrUrl(cfg, project),
     };
   } catch {
     return null;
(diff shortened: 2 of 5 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The main `/api/telegram?action=start` path is fixed, but the PR still leaves one bridge spawn path inheriting Telegram env vars and does not surface the required post-migration restart instruction to operators after `Install Bridge` patches existing AgentChattr configs.

### Findings
- [medium] Setup wizard still starts the bridge with inherited `TELEGRAM_*` / `AGENTCHATTR_URL` env, so Bug 4 remains reproducible on the `npx quadwork init` path.
  - File: `bin/quadwork.js:1125`
  - Suggestion: reuse the same env-scrubbing logic here that was added in `server/routes.js`, so the setup wizard and Settings page behave consistently.

- [medium] The install migration returns `patched_projects`, but the UI still discards that and only alerts `Installed`, so operators never get the explicit "restart AgentChattr" instruction required by the issue when existing configs were patched.
  - File: `server/routes.js:2723`
  - File: `src/components/SettingsPage.tsx:597`
  - Suggestion: when `patched_projects.length > 0`, show a clear success message telling the operator to click `SERVER -> Restart` for the affected projects, or auto-restart if you intentionally choose that route.

### Decision
Requesting changes because the ticket is about an end-to-end fix, and these two remaining paths still leave the bridge misconfigured or the operator uninformed in scenarios explicitly called out by #383.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up closes the remaining end-to-end gaps from my prior review. The setup wizard now scrubs inherited Telegram-related env vars before spawning the bridge, and the UI now surfaces the required restart notice when Install Bridge patches existing AgentChattr configs.

### Findings
- None.

### Decision
Approving because the PR now covers the server start path, the setup-wizard spawn path, the config migration for existing projects, and the operator guidance required after patching configs.

PR #389, opened 2026-04-10
[#386] Fix restart button not killing AC process

### LINKED ISSUE #386: [bug] SERVER → Restart button doesn't kill the existing AC Python process
## Problem

Clicking **SERVER → Restart** in the ControlBar does not kill the existing per-project AgentChattr Python process. The old PID stays alive and continues listening on the same port with stale config. The operator believes the restart happened (the UI may briefly show a transition) but the AC server never picks up config changes like new \`[agents.*]\` declarations.

## Evidence (2026-04-09, quadwork project)

1. AC process PID 86769 started at **09:47** on port 8301.
2. Operator added \`[agents.telegram-bridge]\` to \`~/.quadwork/quadwork/agentchattr/config.toml\` at **18:05**.
3. Operator clicked SERVER → Restart in the QuadWork dashboard.
4. \`lsof -iTCP:8301\` still showed **PID 86769** — same process, never killed.
5. \`curl POST /api/register {base: "telegram-bridge"}\` → \`400 unknown base\` — AC didn't reload the config.
6. Manually running \`kill 86769\` followed by \`POST /api/agentchattr/quadwork/restart\` spawned a new PID (28132) that correctly loaded the updated config and accepted the telegram-bridge registration.

So the **API endpoint** (\`POST /api/agentchattr/:project/restart\`) works correctly — it kills the old process and spawns a new one. The **UI button** either calls a different endpoint, doesn't await the kill, or targets the wrong process.

## Investigation scope

1. **ControlBar.tsx** — find the Restart button's click handler. What endpoint does it call? Does it pass the correct project ID? Does it await the response?
2. **server/index.js** — find the restart endpoint the button hits. Is it the same \`/api/agentchattr/:project/restart\` that worked when called via curl? Or is there a different endpoint?
3. **Process lifecycle** — does the restart handler call \`process.kill(pid, 'SIGTERM')\` and wait for the process to exit before spawning? Or does it fire-and-forget the kill and immediately respawn, potentially racing the old process for the port?
4. **PID file** — does the restart handler read the correct PID file to identify which process to kill? Could it be reading a stale PID file that points to a dead process while the real AC runs under a different PID?

## Acceptance

- Clicking SERVER → Restart kills the existing AC Python process (verified by PID change in \`lsof\`).
- The new AC process starts on the same port with freshly-loaded \`config.toml\`.
- Config changes (e.g. new \`[agents.*]\` sections) take effect after the restart without manual intervention.
- All 4 agent sessions reconnect to the new AC instance (existing behavior from the PTY reattach work in #368).
- No race condition where old and new AC processes fight for the port.

## Branch

\`task/XXX-server-restart-kill-ac\`

## Related

- quadwork#383 — the Telegram Bridge 4-bug fix that exposed this issue (AC needed a restart to pick up the new agent declaration, and the Restart button didn't deliver)
- quadwork#368 (PR #374) — terminal reattach after restart (depends on a real restart actually happening)

realproject7:
Fixes #386

## Summary
- The SERVER → Restart button's handler only killed the tracked `proc.process` reference, which is `null` when QuadWork itself has restarted (AC child processes are spawned `detached: true` + `unref()` and survive Node restarts)
- Added `killProcessOnPort(port)` — uses `lsof` to find and SIGTERM any process listening on the AC port, regardless of whether it's tracked in memory
- Added `waitForPortFree(port, timeout)` — polls until the port is released (up to 3s) before spawning the replacement process, replacing the blind 500ms `setTimeout` that could race
- Applied the same fix to the `stop` and `update` handlers for consistency

## Test plan
- [ ] Start AC via dashboard, note PID via `lsof -iTCP:<port>`
- [ ] Click SERVER → Restart in ControlBar
- [ ] Verify old PID is gone and new PID is listening on the same port
- [ ] Change `config.toml` (e.g. add `[agents.test]`), click Restart, verify new config loads
- [ ] Kill QuadWork (Ctrl+C), restart it, verify Restart button still kills the orphaned AC process

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -834,6 +834,54 @@ async function handleAgentChattr(req, res) {
     return child;
   }
 
+  // #386: Kill any process listening on the AC port. Handles orphaned
+  // processes that survive QuadWork restarts (detached + unref'd spawns
+  // lose their tracked reference when the Node process recycles).
+  function killProcessOnPort(port) {
+    try {
+      const pids = execSync(`lsof -ti TCP:${port} -sTCP:LISTEN`, {
+        encoding: "utf-8",
+        timeout: 5000,
+        stdio: ["pipe", "pipe", "pipe"],
+      }).trim();
+      if (!pids) return;
+      for (const line of pids.split("\n")) {
+        const pid = parseInt(line, 10);
+        if (pid > 0) {
+          try { process.kill(pid, "SIGTERM"); } catch {}
+        }
+      }
+    } catch {
+      // lsof exits non-zero when no matching process — expected
+    }
+  }
+
+  // #386: Poll until the port is free or timeout expires.
+  function waitForPortFree(port, timeoutMs = 3000) {
+    const start = Date.now();
+    return new Promise((resolve) => {
+      function check() {
+        try {
+          execSync(`lsof -ti TCP:${port} -sTCP:LISTEN`, {
+            encoding: "utf-8",
+            timeout: 2000,
+            stdio: ["pipe", "pipe", "pipe"],
+          });
+          // Still occupied — retry if within budget
+          if (Date.now() - start < timeoutMs) {
+            setTimeout(check, 200);
+          } else {
+            resolve(false);
+          }
+        } catch {
+          // lsof found nothing — port is free
+          resolve(true);
+        }
+      }
+      check();
+    });
+  }
+
   if (action === "start") {
     const proc = getProc();
     if (proc.state === "running" && proc.process) {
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The restart handler change addresses the real failure mode in this branch: detached AgentChattr children can outlive QuadWork and leave `proc.process` unusable, so killing by listening port plus waiting for the port to clear is a reasonable fix.

### Findings
- None.

### Decision
Approved. Residual risk: GitHub reports no checks on this branch, and local `npm run build` in this checkout fails with an existing Next.js/Turbopack workspace-root configuration error unrelated to this diff, so behavioral validation is still mainly manual for this PR.

PR #392, opened 2026-04-10
[#388] Fix sender overflow CSS + bridge deregister on stop

### LINKED ISSUE #388: [bug] Chat sender column overflows + stale -2/-3 agent slots accumulate on restarts
### Issue A — Sender column has no width cap or overflow prevention

Long agent names like \`telegram-bridge-2\`, \`reviewer2-2\`, or even the base \`telegram-bridge\` overflow the sender column and bleed into the message text and timestamp columns. The screenshot shows multiple overlapping lines where the sender name runs into the message content, making the chat unreadable.

This is an AC frontend CSS issue. The sender column (\`.msg-sender\` in \`static/style.css\`) has no \`max-width\`, no \`overflow: hidden\`, no \`text-overflow: ellipsis\`, and no \`white-space: nowrap\`. It grows with the text, pushing everything else.

**Fix (in QuadWork's per-project AC clone static files, not upstream):**

Option A — CSS truncation in the sender column:
\`\`\`css
.msg-sender {
  max-width: 80px;       /* fits ~8 chars at 11px font */
  overflow: hidden;
  text-overflow: ellipsis;
  white-space: nowrap;
  display: inline-block;
  vertical-align: middle;
}
\`\`\`

Option B — apply the slug shortening from #371 so names are naturally short (\`head\`, \`dev\`, \`re1\`, \`re2\`, \`tg\`) and the CSS cap is rarely hit. Still add the CSS as a safety net.

**Recommend both** — #371 makes names short, CSS cap prevents any future overflow regardless of name length.

**Implementation approach:** QuadWork already clones AC's static files per-project at \`~/.quadwork/<project>/agentchattr/static/\`. Either:
1. Patch the CSS during the per-project AC install step (append the CSS rules to \`style.css\`), or
2. Ship a QuadWork-specific CSS override file that the AC clone's \`index.html\` includes via a \`<link>\` tag injected during install

Option 1 is simpler but fights future AC upgrades / pin bumps (#348, #366). Option 2 is cleaner but requires an HTML edit. Recommend option 1 for now — the CSS is small and the pin bump procedure already requires re-testing.

### Issue B — Stale -2, -3 suffixed agent slots accumulate on server restarts

Every time QuadWork restarts agents or the AC server, agents re-register. If the old registration slot isn't cleanly deregistered first, AC's registry assigns a new slot (\`head-2\`, \`reviewer1-2\`, \`telegram-bridge-3\`, etc.). These suffixed names:

- Clutter the sender column (even more overflow)
- Confuse operators (which \`telegram-bridge\` is the real one?)
- Persist in AC's registry until timeout (60s) or explicit deregister
- Show up as \"disconnected (timeout)\" messages that spam the chat log

**Root cause:** QuadWork's agent spawn / restart flow doesn't always cleanly deregister before re-registering. The deregister call may fail silently, race the spawn, or not happen at all on certain restart paths (e.g. the SERVER → Restart button path, which we already identified as buggy in #386).

**Fix options:**

1. **QuadWork-side: ensure clean deregister before every register.** The \`agentchattr-registry.js\` already has deregister logic — audit all spawn/restart code paths to ensure it's called AND awaited before the new register. Cover:
   - Initial spawn on \`quadwork start\`
   - Agent restart via the Agent Models widget
   - SERVER → Restart (the #386 flow)
   - Telegram bridge start/stop/restart

2. **AC-side: reduce the timeout window.** AC's registry holds stale slots for 60s before auto-expiring. Reducing this to 10-15s would make the -2/-3 suffixes clear faster. But this is an upstream change (bcurts/agentchattr) or a config.toml tweak.

3. **AC-side: replace-on-register.** If an agent with the same base registers while an old slot is still active, AC could kill the old slot immediately instead of creating a -2 suffix. This is the cleanest behavior but requires an upstream change.

**Recommend fix 1** (QuadWork-side deregister) as the primary fix, with fix 2 (shorter timeout in config.toml) as a belt-and-suspenders.

### Issue A
- No sender name overflows into the message text column in the embedded AC chat iframe, regardless of agent name length.
- Names longer than ~8 chars are truncated with ellipsis in the sender column.
- Tooltip on hover shows the full name (if CSS \`text-overflow: ellipsis\` is used, the \`title\` attribute should be set).

### Issue B
- After a clean SERVER → Restart, agents re-register under their original names (\`head\`, \`dev\`, etc.) without -2/-3 suffixes.
- After clicking Stop → Start on the Telegram Bridge widget, the bridge registers as its base name without accumulating suffixed slots.
- The chat log does not show a burst of \"<agent>-N disconnected (timeout)\" messages after every restart cycle.
- Stale slots from before the fix still expire naturally (no manual cleanup needed).

## Out of scope

- Upstream PRs to bcurts/agentchattr (nice-to-have but not a dependency)
- Agent slug renaming (tracked in #371)
- The SERVER → Restart button not killing AC (tracked in #386)

### Issue A
- \`~/.quadwork/<project>/agentchattr/static/style.css\` — sender column CSS
- \`server/install-agentchattr.js\` or equivalent — post-install CSS patch step (if approach 1)

### Issue B
- \`server/agentchattr-registry.js\` — deregister logic
- \`server/index.js\` — agent spawn/restart lifecycle
- \`server/routes.js\` — telegram bridge start/stop handlers
- \`templates/config.toml\` — AC crash timeout setting (if configurable)

## Branch

\`task/XXX-chat-sender-overflow-stale-slots\`

## Related

- quadwork#371 — agent slug rename (makes names short, addresses root cause of overflow for the 4 main agents + bridge)
- quadwork#386 — SERVER → Restart not killing AC (related lifecycle bug)
- quadwork#383 — telegram bridge config bugs (same restart that triggered the -3 suffix accumulation)

realproject7:
Fixes #388

## Summary
- **Issue A (CSS overflow):** Added `patchAgentchattrCss()` to `install-agentchattr.js` that appends `max-width`, `overflow: hidden`, `text-overflow: ellipsis`, and `white-space: nowrap` to AC's `.msg-sender` class. Called after install and after update (git pull overwrites style.css). Idempotent via marker comment.
- **Issue B (stale slots):** Added deregister call to AC's `/api/deregister/telegram-bridge` in the telegram bridge "stop" handler so the slot clears immediately instead of lingering for 60s as a stale `-2`/`-3` suffix. Regular agent sessions already deregister on stop via `stopAgentSession`.

## Test plan
- [ ] Install AC fresh → verify `.msg-sender` CSS patch in `~/.quadwork/<project>/agentchattr/static/style.css`
- [ ] Update AC (SERVER → Update) → verify CSS patch survives git pull
- [ ] Long agent names in chat are truncated with ellipsis in the sender column
- [ ] Stop telegram bridge → verify `telegram-bridge` slot deregisters from AC immediately (no 60s linger)
- [ ] Restart cycle: no `-2`/`-3` suffixed slots accumulate

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -9,6 +9,7 @@ const { spawn } = require("child_process");
 const { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr, resolveChattrSpawn, syncChattrToken, CONFIG_PATH } = require("./config");
 const routes = require("./routes");
 const { waitForAgentChattrReady, registerAgent, deregisterAgent, startHeartbeat, stopHeartbeat } = require("./agentchattr-registry");
+const { patchAgentchattrCss } = require("./install-agentchattr");
 const { startQueueWatcher, stopQueueWatcher } = require("./queue-watcher");
 
 const net = require("net");
@@ -1006,6 +1007,8 @@ async function handleAgentChattr(req, res) {
       }
 
       const pullResult = execSync("git pull 2>&1", { cwd: acDir, encoding: "utf-8", timeout: 30000 }).trim();
+      // #388: re-apply sender-overflow CSS patch after git pull
+      patchAgentchattrCss(acDir);
       const venvPython = path.join(acDir, ".venv", "bin", "python");
       let pipResult = "";
       const reqFile = path.join(acDir, "requirements.txt");
--- a/server/install-agentchattr.js
+++ b/server/install-agentchattr.js
@@ -192,6 +192,8 @@ function _installAgentChattrLocked(dir, setError) {
       if (pipResult === null) return setError(`pip install -r ${reqFile} failed`);
     }
   }
+  // #388: patch sender-column overflow CSS after clone/install
+  patchAgentchattrCss(dir);
   return dir;
 }
 
@@ -207,9 +209,37 @@ function chattrSpawnArgs(dir, extraArgs) {
   return { command: venvPython, spawnArgs: ["run.py", ...(extraArgs || [])], cwd: dir };
 }
 
+/**
+ * #388: Patch AgentChattr's style.css to cap the sender column width.
+ * Idempotent — skips if the marker comment is already present.
+ * Called after install and after update (git pull overwrites style.css).
+ */
+function patchAgentchattrCss(dir) {
+  if (!dir) return;
+  const cssPath = path.join(dir, "static", "style.css");
+  if (!fs.existsSync(cssPath)) return;
+  try {
+    const content = fs.readFileSync(cssPath, "utf-8");
+    if (content.includes("/* quadwork#388 sender-overflow fix */")) return;
+    const patch = `
+/* quadwork#388 sender-overflow fix */
+.msg-sender {
+  max-width: 80px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  display: inline-block;
+  vertical-align: middle;
+}
+`;
+    fs.writeFileSync(cssPath, content + patch);
+  } catch {}
+}
+
 module.exports = {
   AGENTCHATTR_REPO,
   findAgentChattr,
   installAgentChattr,
   chattrSpawnArgs,
+  patchAgentchattrCss,
 };
--- a/server/install-agentchattr.js
+++ b/server/install-agentchattr.js
@@ -210,18 +210,22 @@ function chattrSpawnArgs(dir, extraArgs) {
 }
 
 /**
- * #388: Patch AgentChattr's style.css to cap the sender column width.
- * Idempotent — skips if the marker comment is already present.
- * Called after install and after update (git pull overwrites style.css).
+ * #388: Patch AgentChattr's static files for sender-column overflow.
+ * Idempotent — skips if the marker is already present.
+ * Called after install and after update (git pull overwrites static/).
+ *
+ * CSS: cap .msg-sender width with ellipsis truncation.
+ * JS:  add title attribute to .msg-sender spans for hover tooltip.
  */
 function patchAgentchattrCss(dir) {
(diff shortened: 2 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR is headed in the right direction, but it does not fully satisfy issue #388 yet. One acceptance criterion is still unimplemented, and the telegram-bridge cleanup remains racy.

### Findings
- [medium] Full sender names are still not accessible on hover after truncation.
  - File: `server/install-agentchattr.js:220`
  - Suggestion: The issue acceptance explicitly requires a hover tooltip for truncated names. The CSS patch only adds ellipsis styling; it does not add a `title` attribute or any equivalent UI. In the current AgentChattr markup, sender spans are rendered as `<span class="msg-sender">...</span>` with no title attribute, so users still cannot recover the full name on hover.
- [medium] Telegram bridge deregistration is fire-and-forget, so stale `-2`/`-3` slots can still accumulate on fast restart cycles.
  - File: `server/routes.js:2837`
  - Suggestion: The new `fetch(.../api/deregister/telegram-bridge).catch(() => {})` is not awaited before the bridge process is killed and before a subsequent start may re-register. That leaves the same race the issue calls out: the new registration can arrive before the old slot is cleared. Await the deregister request (or otherwise confirm completion) before proceeding with stop/start.

### Decision
Requesting changes because the PR does not yet meet all stated acceptance criteria for #388.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The updated PR closes the two gaps from the prior review: sender names now have a hover tooltip after truncation, and telegram-bridge deregistration is awaited before the process is killed.

### Findings
- None.

### Decision
Approved. The change now satisfies the stated acceptance criteria for issue #388 within the scope of this PR.

PR #395, opened 2026-04-10
[#393] Add orphan cleanup to AC start path

### LINKED ISSUE #393: [bug] SERVER -> Start still misses orphaned AgentChattr cleanup after PR #389
## Scope

Batch 6 PR #389 fixed the orphaned-AgentChattr cleanup on `stop`, `restart`, and `update`, but the plain `start` path still goes straight to `spawnChattr()` without cleaning up an orphaned process already listening on the project AgentChattr port.

That leaves the original bug partially reproducible:
- restart QuadWork so the detached AgentChattr child survives but in-memory process state is lost
- click `SERVER -> Start`
- QuadWork attempts a fresh spawn while the orphan is still bound to the port

## Required fix

Apply the same occupied-port protection on the `action === "start"` path that PR #389 added elsewhere:
- detect and terminate any orphaned AgentChattr process holding the project port
- wait briefly for the port to clear before spawning the replacement
- preserve the existing happy path when nothing is already listening

## Acceptance Criteria

- After QuadWork restarts while AgentChattr is still alive in the background, clicking `SERVER -> Start` succeeds without leaving the old orphan on the port.
- The `start` path uses the same cleanup/wait strategy as the existing `restart` path, rather than spawning blindly.
- Normal first-start behavior still works when no process is currently listening on the AgentChattr port.

## Likely files

- `server/index.js`

## Related

- PR #389
- Issue #386

realproject7:
Fixes #393

## Summary
- The `action === "start"` path in `handleAgentChattr` now calls `killProcessOnPort()` + `waitForPortFree()` before spawning, matching the restart/stop/update paths from PR #389
- Fixes the case where QuadWork restarts but a detached AC process survives — clicking Start now cleans up the orphan first

## Test plan
- [ ] Kill QuadWork while AC is running, restart QuadWork, click Start — old orphan is killed, new AC starts on same port
- [ ] Normal first start (no orphan) still works without delay
- [ ] Port occupied by non-AC process: start proceeds after 3s timeout with warning

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -888,6 +888,13 @@ async function handleAgentChattr(req, res) {
     if (proc.state === "running" && proc.process) {
       return res.json({ ok: true, state: "running", message: "Already running" });
     }
+    // #393: kill any orphaned process on the port before spawning
+    // (same pattern as restart/stop from #386).
+    killProcessOnPort(chattrPort);
+    const portFree = await waitForPortFree(chattrPort, 3000);
+    if (!portFree) {
+      console.warn(`[agentchattr] ${projectId} port ${chattrPort} still occupied after 3s — spawning anyway`);
+    }
     try {
       const child = spawnChattr();
       if (!child) {


project7-interns (approved):
## Verdict: APPROVE

### Summary
This PR applies the same orphan-cleanup and port-wait strategy from PR #389 to the plain `start` path, which is exactly what issue #393 requires.

### Findings
- None.

### Decision
Approved. The change is minimal, targeted, and closes the remaining occupied-port gap on `SERVER -> Start`.

PR #406, opened 2026-04-10
[#391] Global notification sound across all pages

### LINKED ISSUE #391: [feat] Notification sound should play for all projects regardless of which page is displayed
## Problem

The notification chime only plays when the operator is viewing the specific project page whose chat panel is active. If the operator is on the Home page, the Settings page, or a different project's page, incoming agent messages on other projects produce no sound.

This defeats the purpose of the notification — operators often leave the dashboard on the Home page or switch between projects, and rely on the chime to know when an agent has posted. Missing a chime means missing a review verdict, a merge request, or a batch completion that needs operator action.

## Expected behavior

When the Sound toggle is **on** for any project, the notification chime should play whenever ANY project with sound enabled receives a new agent message, regardless of which page the operator is currently viewing in the dashboard.

## Current architecture (why it doesn't work)

The notification sound logic lives inside the per-project \`ChatPanel\` component, which is only mounted when the operator navigates to that project's page (\`/project/<id>\`). When the operator is on Home (\`/\`), Settings (\`/settings\`), or a different project's page, the ChatPanel for the other project is unmounted → its WebSocket connection is closed → no messages received → no chime.

## Proposal

Move the notification listener out of \`ChatPanel\` and into a **global component** that:

1. **Runs at the app shell level** (e.g. in \`_app.tsx\` / \`layout.tsx\` / a top-level provider) so it's always mounted regardless of which page the operator is on.
2. **Maintains a lightweight WebSocket or polling connection per project** that has sound enabled. Does NOT need to render chat messages — only needs to detect \"new message from a non-self sender\" events.
3. **Reads each project's sound preferences** from localStorage (same keys the ControlBar's Notification Sound subsection already uses).
4. **Plays the configured chime** when a new agent message arrives on any sound-enabled project.
5. **Avoids duplicate playback** — if the operator IS on the project page and the ChatPanel's own listener already played the chime, the global listener should not double-play. Options:
   - The ChatPanel defers to the global listener entirely (simplest — remove sound logic from ChatPanel)
   - A shared \"last played\" timestamp prevents double-fire within a short window

## Implementation sketch

\`\`\`
GlobalNotificationListener (always mounted in layout)
  ├── for each project with sound enabled:
  │     ├── lightweight WS to /ws/chat?project=<id> (or poll /api/chat?since=<cursor>)
  │     ├── on new message from non-self sender → play chime
  │     └── respect sound choice (Warm Bell / Pluck / etc.) from that project's localStorage prefs
  └── shared Audio instance (reuse across projects to avoid overlapping chimes)
\`\`\`

## Acceptance

- Operator is on the Home page. An agent posts a message on the quadwork project. Chime plays.
- Operator is on the DropCast project page. An agent posts on the quadwork project. Chime plays (quadwork's configured sound).
- Operator is on Settings. Agent posts on any project. Chime plays.
- Sound toggle is off for DropCast but on for quadwork. DropCast messages are silent; quadwork messages chime. Correct per-project behavior.
- No double chime when the operator IS on the project page whose ChatPanel would have also played.
- Chime does not play for the operator's own messages (sender = \"user\").
- Chime does not play for system events (disconnected, timeout, etc.) — only real agent messages.
- Performance: the global listener does not noticeably increase network traffic or CPU. One lightweight WS per project is fine for 2-5 projects; for larger installs, consider polling instead.

## Out of scope

- Desktop OS notifications (system-level toast / banner)
- Per-project sound customization UI changes (already exists in ControlBar)
- Mobile push notifications

## Files likely involved

- New: \`src/components/GlobalNotificationListener.tsx\` (or similar)
- \`src/app/layout.tsx\` or \`src/app/app-shell.tsx\` — mount the global listener
- \`src/components/ChatPanel.tsx\` — remove or defer the per-panel sound logic to avoid double-play
- Notification utils module (wherever \`getNotificationEnabled\` / \`getNotificationChoice\` live)

## Branch

\`task/XXX-global-notification-sound\`

realproject7:
## Summary
- New `GlobalNotificationListener` component mounted in root `layout.tsx` — always active regardless of which page is displayed.
- Polls `/api/chat` per project on a 3s interval, tracks per-project cursors, filters self/system messages, and plays the configured chime on new agent messages.
- Removed notification detection + playback from `ChatPanel` (deferred entirely to the global listener to avoid double-play).
- Reads the same localStorage prefs (sound choice, enabled, background-only) that ControlBar already manages.

## Test plan
- [ ] On Home page: agent posts on any project → chime plays
- [ ] On different project page: agent posts on another project → chime plays
- [ ] On Settings page: agent posts → chime plays
- [ ] Sound toggle off → no chime
- [ ] Background-only mode: tab focused → no chime; tab blurred → chime plays
- [ ] Operator's own messages and system events do not trigger chime
- [ ] No double chime when on the project page (ChatPanel no longer plays)
- [ ] Initial page load does not burst historical chimes (cursor seeding)

Fixes #391

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -3,6 +3,7 @@ import { Geist_Mono } from "next/font/google";
 import "./globals.css";
 import Sidebar from "@/components/Sidebar";
 import TopHeader from "@/components/TopHeader";
+import GlobalNotificationListener from "@/components/GlobalNotificationListener";
 
 const geistMono = Geist_Mono({
   variable: "--font-geist-mono",
@@ -22,6 +23,7 @@ export default function RootLayout({
   return (
     <html lang="en" className={`${geistMono.variable} h-full`}>
       <body className="h-full flex flex-col">
+        <GlobalNotificationListener />
         <TopHeader />
         <div className="flex flex-1 min-h-0">
           <Sidebar />
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -2,7 +2,6 @@
 
 import React, { useState, useEffect, useRef, useCallback } from "react";
 import ProjectChatEmptyState from "./ProjectChatEmptyState";
-import { playNotificationSound } from "../lib/notificationSound";
 
 interface Message {
   id: number;
@@ -192,22 +191,6 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
   // #397 / quadwork#262: tracks the message the next send will be a
   // threaded reply to. Cleared after a successful send or on cancel.
   const [replyTo, setReplyTo] = useState<{ id: number; sender: string } | null>(null);
-  // #409 / quadwork#273 + #405 / quadwork#278: track the operator's
-  // configured display name so the notification-sound filter
-  // suppresses self-messages even when the operator renamed
-  // themselves (e.g. operator_name = "alice"). Held in a ref so the
-  // hot poll path doesn't need a dep on the state value.
-  const operatorNameRef = useRef<string>("user");
-  useEffect(() => {
-    fetch("/api/config")
-      .then((r) => (r.ok ? r.json() : null))
-      .then((cfg) => {
-        if (cfg && typeof cfg.operator_name === "string" && cfg.operator_name) {
-          operatorNameRef.current = cfg.operator_name;
-        }
-      })
-      .catch(() => {});
-  }, []);
   const cursorRef = useRef(0);
   const listRef = useRef<HTMLDivElement>(null);
   const inputRef = useRef<HTMLTextAreaElement>(null);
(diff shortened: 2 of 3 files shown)


realproject7:
**@reviewer2 verdict: APPROVE**

Well-structured change. The global listener in root layout correctly ensures chimes fire regardless of page. Key points verified:

- `playNotificationSound()` handles enabled/background-only/sound-choice prefs internally — no duplication needed.
- Cursor seeding (skip when `prevCursor === 0`) prevents burst of chimes on initial load.
- Operator name filter matches the old ChatPanel logic.
- ChatPanel notification removal is clean — no orphaned refs or dead imports.

Minor non-blocking note: if multiple projects have new messages in the same poll cycle, `playNotificationSound()` fires once per project. In practice this is unlikely to be jarring (polls are async and the browser usually coalesces near-simultaneous audio), but worth noting for future consideration if users report it.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The global listener is the right direction, but the current cursor-seeding logic drops the first real notification for any project whose chat was empty when QuadWork loaded.

### Findings
- [high] First incoming message on a previously empty project never plays a chime.
  - File: `src/components/GlobalNotificationListener.tsx:55`
  - Suggestion: distinguish between "initial historical backfill" and "first real post-start message". Right now `prevCursor === 0` is treated as a blanket skip after `msgs.length > 0`, so when a project had no messages at boot, the first new message sets the cursor and immediately returns without playing.

### Decision
Request changes. This regresses the core acceptance for quieter/newer projects because the first agent message after startup is exactly the notification the operator needs to hear.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The updated global listener now seeds per-project cursors before polling begins, which fixes the missed-first-message regression on previously empty projects. The listener remains mounted at the app shell level and still removes the double-play risk from `ChatPanel`.

### Findings
- No blocking findings.
- Note: `gh pr checks 406` currently reports no automated checks on this branch.

### Decision
Approve. The rework addresses the requested fix and now matches the intended global-notification behavior for new messages across pages.

PR #409, opened 2026-04-10
[#391] Fix global notification resilience + project refresh

realproject7:
## Summary
Follow-up to merged PR #406, addressing two findings from @head's Batch 8 review:

1. **Config fetch failure resilience**: `refreshConfig()` now retries every 30s instead of being a one-shot. A transient startup failure no longer permanently disables notifications for the session.

2. **Periodic project list refresh**: Project list is re-fetched every 30s so newly added projects start chiming and removed projects stop polling — without requiring a page reload.

Also improved cursor seeding to be per-project and skip already-seeded projects, so new projects discovered mid-session get proper cursor initialization.

## Test plan
- [ ] Kill the QuadWork server briefly on page load, then restart → notifications recover within 30s
- [ ] Add a new project while dashboard is open → chimes start within 30s without reload
- [ ] Remove a project → polling stops for it within 30s
- [ ] Existing behavior unchanged: chimes on new agent messages, no chime on self/system

### DIFF
--- a/src/components/GlobalNotificationListener.tsx
+++ b/src/components/GlobalNotificationListener.tsx
@@ -17,6 +17,9 @@ interface Project {
   agentchattr_url?: string;
 }
 
+// How often to re-fetch config to pick up added/removed projects.
+const CONFIG_REFRESH_MS = 30_000;
+
 export default function GlobalNotificationListener() {
   // Per-project cursors so we only fire on genuinely new messages.
   const cursorsRef = useRef<Record<string, number>>({});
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The retry and periodic refresh parts are good, but the new per-project cursor seeding still leaves a correctness hole for newly added projects.

### Findings
- [high] A project is added to `projectsRef.current` before its cursor seeding finishes, and `pollAll()` skips that project until `cursorsRef.current[project.id]` is set. If the first real agent message arrives in that window, the seed request can fetch it and record its id as already seen, so no chime will fire for that first message.
  - File: `src/components/GlobalNotificationListener.tsx:33`
  - Suggestion: Make project activation atomic with cursor initialization, or seed from a boundary that cannot consume the first post-add message as historical state.

### Decision
Requesting changes because PR #409 still fails the newly-added-project edge case that this follow-up is meant to cover.

project7-interns (approved):
## Verdict: APPROVE

### Summary
Re-review complete. The cursor initialization is now synchronous and the first poll is explicitly chime-suppressed, which removes the async seeding race for newly added projects while preserving the no-false-chime behavior for historical messages.

### Findings
- None blocking.

### Decision
Approving. This revision addresses the newly-added-project notification gap I previously flagged and still covers the retry and periodic refresh fixes in scope for PR #409.

PR #411, opened 2026-04-10
[#410] Fix chat panel welcome screen showing over existing messages

### LINKED ISSUE #410: [bug] Chat panel shows 'Ready when you are' welcome screen despite existing messages in the channel
## Problem

When navigating to a project page, the AgentChattr primary chat panel displays the empty/welcome state (\"Ready when you are\" with example prompts) even when the channel already has messages from agents, the Telegram bridge, or prior conversations.

The welcome screen only disappears after the operator sends a message directly via the chat input — at that point the full message history loads and renders correctly. Messages that arrived via Telegram bridge, agent-to-agent chat, or the Scheduled Trigger do not cause the panel to transition out of the welcome state.

## Expected behavior

On page load, the chat panel should immediately poll for existing messages and display them if any exist. The welcome screen should only appear when the channel has **zero messages** — a genuinely fresh project with no chat history.

## Likely root cause

The chat panel is an embedded AgentChattr iframe (\`ChatPanel.tsx\`, iframe mode). The iframe loads AC's own web UI, which starts with the welcome/empty state and only fetches messages after a user interaction or WebSocket connection event. If the iframe's initial load doesn't trigger a message fetch (e.g. because the WS auth token is stale, or the initial fetch returns before any messages are rendered), the welcome state persists.

Alternatively, in API fallback mode (\`ChatPanelAPI\`), the component might have a similar issue where the initial \`/api/chat\` poll returns messages but the render logic still shows the welcome screen until a local send event occurs.

## Investigation hints

- Check whether the iframe mode or the API fallback mode is active when this happens (inspect the DOM for an \`<iframe>\` vs the API-rendered message list)
- If iframe: AC's own \`static/chat.js\` may gate the message display behind a user interaction or a specific WebSocket event
- If API fallback: \`ChatPanelAPI\`'s initial \`useEffect\` fetch may have a race condition where the welcome state renders before the fetch completes, and the fetch result doesn't trigger a re-render

## Acceptance

- Navigating to a project page with existing chat messages immediately shows those messages (no welcome screen)
- Navigating to a project page with zero messages shows the welcome screen
- Messages arriving via Telegram bridge while the page is open appear in real-time (no need to send a message first to \"unlock\" the chat view)
- No regression to the chat input, @mention autocomplete, or the iframe/API fallback switching logic

## Out of scope

- Redesigning the welcome screen content
- Chat panel performance / pagination for very large histories

## Files likely involved

- \`src/components/ChatPanel.tsx\` — iframe vs API fallback detection, initial fetch logic
- \`~/.quadwork/<project>/agentchattr/static/chat.js\` — if iframe mode, the AC frontend's message loading logic

## Branch

\`task/XXX-chat-welcome-screen-existing-messages\`

realproject7:
## Summary
- **Bug**: Chat panel displayed the "Ready when you are" welcome screen on page load even when the channel had existing messages (from bridge, agents, or prior conversations). The welcome screen only disappeared after the user sent a message locally.
- **Root cause**: `messages` state starts as `[]`, so the empty-state guard (`messages.length === 0`) was true until the first poll returned — and the welcome screen rendered in that window. For channels with messages, the poll would fill `messages`, but the initial flash was jarring and confusing.
- **Fix**: Added a `loaded` flag (default `false`) that flips to `true` after the first successful `/api/chat` poll. The welcome/empty state only renders when `loaded && messages.length === 0`. A "Loading messages..." placeholder shows during the initial fetch.

Fixes #410

## Test plan
- [ ] Navigate to a project with existing chat messages — should show messages immediately, no welcome screen flash
- [ ] Navigate to a project with zero messages — should show "Loading messages..." briefly, then the welcome screen
- [ ] Switch between projects — loading state resets correctly per project
- [ ] Messages arriving via Telegram bridge while page is open appear in real-time
- [ ] No regression to chat input, @mention autocomplete, or slash command menu

### DIFF
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -228,7 +228,6 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
       })
       .then((data) => {
         setAuthError(null);
-        setLoaded(true);
         const msgs: Message[] = Array.isArray(data) ? data : data.messages || [];
         if (msgs.length > 0) {
           setMessages((prev) => {
@@ -240,7 +239,8 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
           if (maxId > cursorRef.current) cursorRef.current = maxId;
         }
       })
-      .catch(() => {});
+      .catch(() => {})
+      .finally(() => { setLoaded(true); });
   }, [channel, projectId]);
 
   useEffect(() => {
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -175,6 +175,9 @@ export default function ChatPanel({ projectId }: ChatPanelProps) {
 function ChatPanelAPI({ projectId }: { projectId?: string }) {
   const channel = "general";
   const [messages, setMessages] = useState<Message[]>([]);
+  // #410: track whether the initial fetch has completed so we don't
+  // flash the welcome/empty state while messages are still loading.
+  const [loaded, setLoaded] = useState(false);
   const [input, setInput] = useState("");
   const [showMentions, setShowMentions] = useState(false);
   const [mentionFilter, setMentionFilter] = useState("");
@@ -207,6 +210,7 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
   useEffect(() => {
     cursorRef.current = 0;
     setMessages([]);
+    setLoaded(false);
   }, [projectId]);
 
   // Poll messages via proxy
@@ -224,6 +228,7 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
       })
       .then((data) => {
         setAuthError(null);
+        setLoaded(true);
         const msgs: Message[] = Array.isArray(data) ? data : data.messages || [];
         if (msgs.length > 0) {
           setMessages((prev) => {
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
This change is narrow and low-risk, but it only updates the API fallback renderer. The main `ChatPanel` component still prefers the iframe path first, so the PR does not clearly fix the issue on the primary chat path.

### Findings
- [high] The fix only applies to `ChatPanelAPI`, which is rendered only after the iframe path fails. `ChatPanel` still attempts iframe mode first and returns early when that succeeds, so operators hitting the bug in the normal iframe path will still see the existing behavior.
  - File: `src/components/ChatPanel.tsx:38`
  - Suggestion: Either reproduce and fix the iframe mode / AgentChattr UI path, or establish in code and test evidence that the affected production path is always the API fallback before merging.

### Decision
Requesting changes because the current patch does not demonstrate that issue #410 is fixed in the component's primary render path.

project7-interns (approved):
## Verdict: APPROVE

### Summary
Re-review complete. The additional explanation and current code path show that the user-facing QuadWork chat panel normally resolves to `ChatPanelAPI` in standard deployments, so this patch is targeting the correct path for issue #410.

### Findings
- None blocking.

### Decision
Approving. The `loaded` gate fixes the premature empty-state render in the API fallback path, and the follow-up change to set it in `.finally()` also avoids leaving the panel stuck in loading on failed polls.

PR #412, opened 2026-04-10
[#408] Auto-trigger toggle for batch lifecycle

### LINKED ISSUE #408: [feat] Auto-trigger: toggle switch that auto-starts/stops Scheduled Trigger on batch lifecycle
## Problem

Running an overnight autonomous batch requires the operator to:
1. Manually click **Start Trigger** when a batch begins
2. Manually click **Stop Trigger** when the batch completes
3. Manually click **Start Trigger** again when the next batch begins

If the operator forgets step 1, agents sit idle. If they forget step 2, the trigger keeps firing into a completed batch, wasting tokens. If they go to sleep after starting a batch and Head finishes + starts the next batch while they're away, the trigger doesn't restart for the new batch.

## Proposal

Add an **Auto-Trigger** toggle switch to the Scheduled Trigger widget:

\`\`\`
┌─────────────────────────────────────────────┐
│ SCHEDULED TRIGGER                           │
│                                             │
│ Auto-Trigger: [ON ■□ OFF]                   │
│                                             │
│ [Message textarea...]                       │
│ Send every [15] min for [3] hours           │
│ [Start Trigger]                             │
└─────────────────────────────────────────────┘
\`\`\`

### Behavior when Auto-Trigger is ON

1. **Batch starts** (Current Batch panel transitions from \"no active batch\" or \"COMPLETE\" to a new batch with items) → **Scheduled Trigger auto-starts** with the configured message, interval (default 15 min), and duration (default 3 hours). Equivalent to clicking Start Trigger automatically.

2. **Batch completes** (Current Batch panel shows \"✅ COMPLETE — All N items merged\") → **Scheduled Trigger auto-stops**. The running trigger is cancelled. No more periodic messages fire. But the Auto-Trigger toggle **stays ON** — it's waiting for the next batch to start.

3. **New batch starts** (Head writes a new Active Batch to OVERNIGHT-QUEUE.md, Current Batch transitions from COMPLETE to a fresh batch) → **Scheduled Trigger auto-starts again** with the same defaults.

This creates a hands-free loop: the operator turns on Auto-Trigger once, starts the first batch, and goes to sleep. The trigger fires while the batch runs, stops when it completes, and restarts if Head begins a new batch overnight.

### Trigger defaults when auto-started

- **Message:** the current textarea content (whatever the operator typed or the default template)
- **Interval:** the current interval field value (default 15 min)
- **Duration:** the current duration field value (default 3 hours)

The operator can customize all three before enabling Auto-Trigger. The auto-start uses whatever values are currently set.

### Detection mechanism

The auto-trigger logic needs to detect batch lifecycle transitions. Two approaches:

**Option A (recommended) — Poll the batch-progress endpoint:**
The widget already knows the project ID. Poll \`GET /api/batch-progress?project=<id>\` on the same cadence as the existing batch panel (30s). Track the previous state and detect transitions:
- \`complete: false\` → \`complete: true\` = batch just completed → stop trigger
- \`items.length === 0\` → \`items.length > 0\` = new batch started → start trigger
- \`complete: true\` → \`complete: false\` (new batch with items) = new batch started → start trigger

**Option B — Server-side event:**
Add a webhook/callback in the batch-progress resolver that fires on state transitions. Cleaner but more invasive. Save for a future iteration.

### Persistence

- Auto-Trigger ON/OFF state persists in the project config (\`~/.quadwork/config.json\` under \`project.trigger_auto\`) so it survives server restarts.
- When QuadWork starts and Auto-Trigger is ON, check the current batch state immediately — if there's an active non-complete batch, auto-start the trigger right away.

### UI

- Toggle switch in the Scheduled Trigger widget header, styled consistently with other toggles (similar to Sound toggle in Notification Sound)
- When ON: show a small indicator like \`Auto ●\` in accent color
- When the trigger is auto-started, show \`(auto)\` next to the Running pill so the operator knows it wasn't manually started
- When the trigger is auto-stopped on batch complete, briefly flash a status message: \`Batch complete — trigger paused. Waiting for next batch.\`

## Acceptance

- Auto-Trigger toggle appears in the Scheduled Trigger widget
- Toggle ON + active batch running → trigger starts automatically within one poll cycle (~30s)
- Batch completes (all items merged) → trigger stops automatically, toggle stays ON
- New batch starts while toggle is ON → trigger restarts automatically
- Toggle OFF → trigger behaves exactly as before (manual Start/Stop only), no lifecycle detection
- Auto-trigger state persists across server restarts
- Manual Start/Stop still works when Auto-Trigger is ON (operator can override at any time)
- When both auto-start and manual start could race, the trigger that's already running wins (don't double-start)
- Interval/duration/message fields are editable while Auto-Trigger is ON — changes apply to the next auto-start cycle (current running trigger keeps its original values until it stops)

## Edge cases

- **Batch completes but Auto-Trigger was turned OFF mid-batch:** trigger keeps running until duration expires or operator clicks Stop (no auto-stop since toggle is OFF)
- **No active batch when toggle is turned ON:** nothing happens immediately. Trigger waits for a batch to appear.
- **Server restart during an active batch with Auto-Trigger ON:** on startup, detect active batch + auto-trigger enabled → auto-start trigger immediately
- **Duration expires before batch completes:** trigger stops (duration-based expiry, same as today). Does NOT auto-restart until the batch actually completes AND a new batch begins. Otherwise it would loop indefinitely on a stuck batch.

## Out of scope

- Auto-configuring the trigger message content based on batch context
- Per-batch duration customization (always uses the widget's current duration setting)
- Webhook/server-push-based detection (save for Option B follow-up)
- Auto-adjusting interval based on batch size or agent activity

## Files likely involved

- \`src/components/ScheduledTriggerWidget.tsx\` — toggle UI, auto-start/stop logic, batch state polling
- \`server/routes.js\` or \`server/index.js\` — persist \`trigger_auto\` in project config
- \`src/components/BatchProgressPanel.tsx\` — may need to expose batch state to sibling components (or the trigger widget polls independently)

## Branch

\`task/XXX-auto-trigger-batch-lifecycle\`

## Related

- quadwork#210 — original Scheduled Trigger feature
- quadwork#306 / quadwork#418 — \"first message fires at T + interval, not on click\" behavior (preserved — auto-start follows the same rule)

realproject7:
## Summary
- Adds an **Auto-Trigger** toggle button to the Scheduled Trigger widget header
- When ON, polls `/api/batch-progress` every 30s to detect batch lifecycle transitions:
  - **Batch starts** (items appear, `complete: false`) → auto-starts trigger with current message/interval/duration
  - **Batch completes** (all items merged/closed) → auto-stops trigger, shows status flash
  - **New batch starts** after completion → auto-restarts trigger
- Toggle state persists in project config (`trigger_auto` field) across server restarts
- On startup with toggle ON + active non-complete batch → auto-starts immediately
- Shows `(auto)` in header when trigger was auto-started vs `(running)` for manual
- Manual Start/Stop still works at any time; already-running trigger wins (no double-start)

Fixes #408

## Test plan
- [ ] Auto toggle appears in Scheduled Trigger widget header, styled like Sound toggle
- [ ] Toggle ON + active batch running → trigger auto-starts within ~30s
- [ ] Batch completes (all merged) → trigger auto-stops, flash message appears
- [ ] New batch starts while toggle ON → trigger auto-restarts
- [ ] Toggle OFF → no lifecycle detection, manual-only behavior preserved
- [ ] Toggle state persists across page reload / server restart
- [ ] Manual Stop during auto-triggered run → clears "(auto)" label
- [ ] Manual Start while toggle ON → works normally, shows "(running)" not "(auto)"
- [ ] Duration expires before batch completes → trigger stops, does NOT auto-restart until next batch
- [ ] No active batch when toggle ON → nothing happens, waits for batch

### DIFF
--- a/src/components/ScheduledTriggerWidget.tsx
+++ b/src/components/ScheduledTriggerWidget.tsx
@@ -129,7 +129,15 @@ export default function ScheduledTriggerWidget({ projectId }: ScheduledTriggerWi
   const prevBatchRef = useRef<{ complete: boolean; hasItems: boolean } | null>(null);
 
   // #408: load auto-trigger setting from project config on mount.
+  // Reset refs on projectId change to avoid stale state across projects.
   const autoTriggerLoadedRef = useRef(false);
+  useEffect(() => {
+    autoTriggerLoadedRef.current = false;
+    prevBatchRef.current = null;
+    setAutoTrigger(false);
+    setAutoTriggered(false);
+    setAutoStatus(null);
+  }, [projectId]);
   useEffect(() => {
     if (autoTriggerLoadedRef.current) return;
     fetch("/api/config")
--- a/src/components/ScheduledTriggerWidget.tsx
+++ b/src/components/ScheduledTriggerWidget.tsx
@@ -7,6 +7,16 @@ interface ScheduledTriggerWidgetProps {
   projectId: string;
 }
 
+// #408: batch progress shape from /api/batch-progress
+interface BatchState {
+  complete: boolean;
+  items: { issue_number: number; status: string }[];
+  batch_number: number | null;
+}
+
+// How often the auto-trigger polls batch progress (same as BatchProgressPanel).
+const AUTO_TRIGGER_POLL_MS = 30_000;
+
 interface TriggerInfo {
   enabled: boolean;
   interval: number;  // ms (active timer interval, 0 when idle-with-saved-state)
@@ -112,6 +122,28 @@ export default function ScheduledTriggerWidget({ projectId }: ScheduledTriggerWi
   const [countdown, setCountdown] = useState("");
   const [expiresCountdown, setExpiresCountdown] = useState("");
   const initialMessage = useMemo(() => defaultMessage(projectId), [projectId]);
+  // #408: auto-trigger state
+  const [autoTrigger, setAutoTrigger] = useState(false);
+  const [autoTriggered, setAutoTriggered] = useState(false); // true when current run was auto-started
+  const [autoStatus, setAutoStatus] = useState<string | null>(null); // flash message
+  const prevBatchRef = useRef<{ complete: boolean; hasItems: boolean } | null>(null);
+
+  // #408: load auto-trigger setting from project config on mount.
+  const autoTriggerLoadedRef = useRef(false);
+  useEffect(() => {
+    if (autoTriggerLoadedRef.current) return;
+    fetch("/api/config")
+      .then((r) => (r.ok ? r.json() : null))
+      .then((cfg) => {
+        if (!cfg) return;
+        const entry = (cfg.projects || []).find((p: { id: string }) => p.id === projectId);
+        if (entry?.trigger_auto) {
+          setAutoTrigger(true);
+        }
+        autoTriggerLoadedRef.current = true;
+      })
+      .catch(() => {});
+  }, [projectId]);
 
   const load = useCallback(async () => {
     try {
@@ -228,6 +260,7 @@ export default function ScheduledTriggerWidget({ projectId }: ScheduledTriggerWi
     try {
       const r = await fetch(`/api/triggers/${encodeURIComponent(projectId)}/stop`, { method: "POST" });
       if (!r.ok) throw new Error(`${r.status}`);
+      setAutoTriggered(false);
       setTrigger({
         ...(trigger || {
           interval: 0,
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The batch lifecycle logic is close, but the persisted auto-toggle state is loaded incorrectly across project switches.

### Findings
- [high] `autoTriggerLoadedRef` is a one-way latch that never resets when `projectId` changes. After the first project sets it to `true`, navigating to a different project skips the `/api/config` read entirely, so the new project can inherit stale `autoTrigger` state from the previous one instead of its own `trigger_auto` setting.
  - File: `src/components/ScheduledTriggerWidget.tsx:131`
  - Suggestion: Re-read project config on every `projectId` change, or key the loaded flag by project id instead of using a single sticky boolean.

### Decision
Requesting changes because the per-project persistence requirement is broken when the widget is reused across project navigation.

project7-interns (approved):
## Verdict: APPROVE

### Summary
Re-review complete. The added `projectId` reset effect clears the sticky auto-trigger state before reloading config, so the widget now reads `trigger_auto` per project instead of carrying state across navigation.

### Findings
- None blocking.

### Decision
Approving. This resolves the per-project persistence bug I previously flagged and keeps the batch lifecycle automation behavior intact.

PR #427, opened 2026-04-11
[#414] Fix tooltip popover: OVERNIGHT-QUEUE.md clipped + Loop Guard inline

### LINKED ISSUE #414: [bug] Tooltip fixes: OVERNIGHT-QUEUE.md tooltip inactive + Loop Guard tooltip renders inline instead of popover
### Bug 1 — OVERNIGHT-QUEUE.md tooltip is not clickable

The \`(?)\` button next to \"OVERNIGHT-QUEUE.md\" in the GitHub panel footer does not respond to clicks. The popover never appears.

**Likely cause:** the \`(?)\` button may be inside a parent element with \`pointer-events: none\`, or the click handler is being swallowed by the Edit button's container, or the \`InfoTooltip\` component is not mounted at all (rendering issue).

**Location:** \`src/components/GitHubPanel.tsx\` — the Overnight Queue row at the bottom of the GitHub panel.

### Bug 2 — Loop Guard tooltip renders inline, not as a popover

The Loop Guard section shows its help text as an **expanded inline block** below the header, pushing the controls down. It should render as a floating popover (absolute-positioned, \`z-30\`, overlay) like the other tooltips — matching the \`InfoTooltip\` component pattern.

**Likely cause:** the Loop Guard section (\`LoopGuardWidget.tsx\`) is using the old #311-style inline help pattern (a \`showHelp\` state that conditionally renders a \`<div>\` in the document flow) instead of the new \`InfoTooltip\` component which uses absolute positioning.

**Location:** \`src/components/LoopGuardWidget.tsx\` — the help text rendering near the section header.

### Fix

**Bug 1:** inspect why the \`InfoTooltip\` in the Overnight Queue row doesn't respond. Check z-index, pointer-events, event propagation. Test the click in isolation.

**Bug 2:** replace the Loop Guard's inline help pattern with the \`InfoTooltip\` component (same as all other sections). Remove the old \`showHelp\` state + inline \`<div>\` and use \`<InfoTooltip>\` in the header instead.

### Acceptance

- Clicking \`(?)\` next to OVERNIGHT-QUEUE.md opens a floating tooltip popover
- Loop Guard tooltip appears as a floating popover, not an inline expanded block
- Both dismiss on click-outside and Escape
- No layout shift when tooltips open

## Branch

\`task/XXX-tooltip-followup-fixes\`

## Related

- quadwork#407 (PR #413) — original tooltip implementation

realproject7:
## Summary
Fixes #414

- **Bug 1 — OVERNIGHT-QUEUE.md tooltip invisible:** The `(?)` popover opened downward but the parent panel has `overflow-hidden`, clipping it entirely. Added a `position="above"` prop to `InfoTooltip` so bottom-of-panel tooltips render upward.
- **Bug 2 — Loop Guard tooltip renders inline:** Replaced the old `showHelp` state + inline `<div>` pattern with the shared `InfoTooltip` component so it renders as a floating popover matching all other sections.

## Changes
- `InfoTooltip.tsx` — new optional `position` prop (`"below"` default / `"above"`)
- `GitHubPanel.tsx` — use `position="above"` on the OVERNIGHT-QUEUE.md tooltip
- `LoopGuardWidget.tsx` — remove `showHelp` state, use `<InfoTooltip>` instead

## Test plan
- [ ] Click `(?)` next to OVERNIGHT-QUEUE.md → floating popover appears **above** the button
- [ ] Click `(?)` next to Loop Guard → floating popover appears (not inline block)
- [ ] Both dismiss on click-outside and Escape
- [ ] No layout shift when tooltips open
- [ ] Other existing tooltips (Issues, Pull Requests, Batch Progress) still work unchanged

### DIFF
--- a/src/components/GitHubPanel.tsx
+++ b/src/components/GitHubPanel.tsx
@@ -307,7 +307,7 @@ export default function GitHubPanel({ projectId }: GitHubPanelProps) {
       <div className="shrink-0 flex items-center justify-between px-3 py-1.5 border-t border-border">
         <div className="flex items-center gap-1.5">
           <span className="text-[11px] text-text-muted font-mono">OVERNIGHT-QUEUE.md</span>
-          <InfoTooltip>
+          <InfoTooltip position="above">
             <b>Overnight Queue</b> — the task queue file Head reads to pick the next ticket. Click Edit to modify batch contents and ordering.
           </InfoTooltip>
         </div>
--- a/src/components/InfoTooltip.tsx
+++ b/src/components/InfoTooltip.tsx
@@ -4,6 +4,8 @@ import { useState, useRef, useEffect } from "react";
 
 interface InfoTooltipProps {
   children: React.ReactNode;
+  /** Open popover above the button instead of below. */
+  position?: "below" | "above";
 }
 
 /**
@@ -14,7 +16,7 @@ interface InfoTooltipProps {
  * AgentTerminalsGrid — consistent `w-3.5 h-3.5 rounded-full`
  * styling with accent hover.
  */
-export default function InfoTooltip({ children }: InfoTooltipProps) {
+export default function InfoTooltip({ children, position = "below" }: InfoTooltipProps) {
   const [open, setOpen] = useState(false);
   const ref = useRef<HTMLDivElement>(null);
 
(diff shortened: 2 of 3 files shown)


realproject7:
## RE2 Review — **APPROVE** ✅

All four acceptance criteria met:

1. **OVERNIGHT-QUEUE.md tooltip** — `position="above"` uses `bottom-5` instead of `top-5`, so the popover opens upward and won't be clipped by the parent's `overflow-hidden`. Correct fix.

2. **Loop Guard tooltip** — Old inline `showHelp` state + `<div>` replaced with the shared `InfoTooltip` component. Help text content preserved verbatim.

3. **InfoTooltip `position` prop** — Clean addition. Defaults to `"below"` so all existing usages are unaffected.

4. **Net -6 lines** — good simplification by reusing the shared component.

No issues found.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR is tightly scoped to #414 and fixes both tooltip regressions without introducing unnecessary changes. The shared tooltip now supports upward rendering for the footer row, and Loop Guard correctly switches from inline help to the shared floating popover.

### Findings
- None.

### Decision
Approving because the implementation matches the issue acceptance criteria and preserves the existing tooltip behavior elsewhere.

PR #428, opened 2026-04-11
[#424] Fix Discord bridge echo prevention on AC slot rename

### LINKED ISSUE #424: [bug] Discord bridge echo prevention fails when AC renames slot
## Found in
Audit of PR #420 (issue #397) — `bridges/discord/discord_bridge.py`

## Bug
When AgentChattr renames the bridge slot (e.g. `discord-bridge` → `discord-bridge-2` on restart/conflict), the echo prevention at line 262 partially breaks:

```python
if sender == bridge_sender or sender == ac["name"]:
```

- `bridge_sender` is the static config value (e.g. `"discord-bridge"`) — never updates
- `ac["name"]` correctly reflects the registered name from AC (e.g. `"discord-bridge-2"`)
- If AC assigns `discord-bridge-2`, messages from that sender are correctly caught by `ac["name"]`

However, the reverse case is the real problem: if the bridge re-registers and gets back slot 1 (`discord-bridge`), but there are still old messages in AC from a previous `discord-bridge-2` session, those would NOT be filtered. The `bridge_sender` check partially covers this but only for the exact configured name.

## Expected
Echo prevention should filter messages from any sender identity the bridge has ever used in this session.

## Actual
Only filters the configured `bridge_sender` and the current `ac["name"]`. Stale slot names from previous registrations can echo.

## Fix
Track all registered names in a set and check against it:
```python
if sender in ac["known_names"] or sender == bridge_sender:
```

## Severity
Medium — creates duplicate messages in Discord under AC restart scenarios.

realproject7:
## Summary
Fixes #424

When AgentChattr renames the bridge slot on restart (e.g. `discord-bridge` → `discord-bridge-2`), messages from previous slot names were not filtered by echo prevention, causing duplicate messages in Discord.

- Added `known_names` set to the `ac` state dict
- `ac_register` now adds each assigned name to the set
- Echo prevention checks `sender in ac["known_names"]` instead of only `sender == ac["name"]`

## Test plan
- [ ] Start bridge, note AC assigns `discord-bridge`
- [ ] Restart AC so bridge re-registers as `discord-bridge-2`
- [ ] Verify messages from both `discord-bridge` and `discord-bridge-2` are filtered (no echo in Discord)
- [ ] Verify normal agent messages still relay to Discord

### DIFF
--- a/bridges/discord/discord_bridge.py
+++ b/bridges/discord/discord_bridge.py
@@ -145,7 +145,7 @@ def save_cursor(path):
 
 # Mutable dict so heartbeat thread sees re-registration updates.
 # bridge_sender is set from cfg during main() so all callers can read it.
-ac = {"token": "", "name": "", "bridge_sender": "discord-bridge"}
+ac = {"token": "", "name": "", "bridge_sender": "discord-bridge", "known_names": set()}
 
 
 def ac_register(url, base=None, label="Discord Bridge"):
@@ -161,7 +161,8 @@ def ac_register(url, base=None, label="Discord Bridge"):
     data = resp.json()
     ac["name"] = data["name"]
     ac["token"] = data["token"]
-    log.info("Registered with AC as %s", ac["name"])
+    ac["known_names"].add(data["name"])
+    log.info("Registered with AC as %s (known: %s)", ac["name"], ac["known_names"])
     return data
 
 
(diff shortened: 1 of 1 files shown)


realproject7:
## RE2 Review — **APPROVE** ✅

Correct fix for the echo prevention gap:

1. **`known_names` set** — accumulates every slot name AC assigns during the session. Initialized empty in the module-level `ac` dict, populated in `ac_register`.

2. **Echo check** — `sender in ac["known_names"]` catches all past slot names (e.g. `discord-bridge`, `discord-bridge-2`). The `bridge_sender` fallback covers the configured name even before first registration.

3. **Logging** — includes the full `known_names` set, useful for debugging slot rename scenarios.

4. **Scope** — 1 file, +6 −4, no unrelated changes.

No issues found.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR is narrowly scoped to #424 and fixes the echo-prevention gap when AgentChattr reassigns the bridge sender name across re-registrations. Tracking all registered names in-session matches the issue and keeps the rest of the relay flow unchanged.

### Findings
- None.

### Decision
Approving because the implementation addresses the stale-slot echo case described in #424 without introducing scope creep or breaking the existing Discord send/poll paths.

PR #429, opened 2026-04-11
[#425] Fix Discord bridge dropping messages under high volume

### LINKED ISSUE #425: [bug] Discord bridge AC poll missing pagination — drops messages under high volume
## Found in
Audit of PR #420 (issue #397) — `bridges/discord/discord_bridge.py`

## Bug
The AC → Discord poll loop (line 225) requests `?limit=50&since_id=<cursor>` but does NOT handle pagination. If AgentChattr returns exactly 50 messages, there may be more available, but the loop sleeps for `poll_interval` seconds before the next fetch — silently dropping the overflow.

## Expected
When the response contains exactly `limit` messages, immediately re-fetch with the updated `since_id` to drain the backlog before sleeping.

## Actual
Always sleeps after processing one batch, regardless of whether more messages are available.

## Impact
Under high message volume (e.g. multi-agent batch with rapid PR review cycles), the bridge silently skips messages between polls. The Telegram bridge has the same limitation but its traffic is typically lower.

## Fix
After processing a batch, check `if len(messages) >= 50:` and loop immediately with updated cursor instead of sleeping.

## Severity
Medium — only manifests under high message volume, but silently drops messages with no error.

realproject7:
## Summary
Fixes #425

The AC → Discord poll loop fetched one batch of 50 messages per interval and slept, silently dropping any overflow. Now the loop drains all available messages by immediately re-fetching when it gets a full batch (≥50), only sleeping once the backlog is empty.

## Changes
- `bridges/discord/discord_bridge.py` — wrapped the fetch+process block in an inner `while True` loop that `continue`s on full batches and `break`s otherwise

## Test plan
- [ ] Under normal volume (<50 msgs/interval): behavior unchanged, single fetch per cycle
- [ ] Under high volume (>50 msgs queued): bridge drains all pages before sleeping
- [ ] Auth failure (401/403) breaks out of inner loop and retries on next cycle
- [ ] Cursor persisted after each page, not just after the final page

### DIFF
--- a/bridges/discord/discord_bridge.py
+++ b/bridges/discord/discord_bridge.py
@@ -224,74 +224,82 @@ async def poll_ac_to_discord(cfg, channel):
 
     while True:
         try:
-            params = {"limit": 50}
-            if _cursor["last_seen_id"]:
-                params["since_id"] = _cursor["last_seen_id"]
-            headers = {}
-            if ac["token"]:
-                headers["Authorization"] = f"Bearer {ac['token']}"
+            # Drain all available messages before sleeping. When AC
+            # returns a full batch (limit messages), immediately
+            # re-fetch with the updated cursor to avoid dropping
+            # overflow under high volume.
+            while True:
+                params = {"limit": 50}
+                if _cursor["last_seen_id"]:
+                    params["since_id"] = _cursor["last_seen_id"]
+                headers = {}
+                if ac["token"]:
+                    headers["Authorization"] = f"Bearer {ac['token']}"
 
-            resp = requests.get(
-                f"{url}/api/messages",
-                params=params,
-                headers=headers,
-                timeout=10,
-            )
+                resp = requests.get(
+                    f"{url}/api/messages",
+                    params=params,
+                    headers=headers,
+                    timeout=10,
+                )
 
-            if resp.status_code in (401, 403):
-                log.warning("AC poll %d — re-registering", resp.status_code)
-                try:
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR is tightly scoped to #425 and fixes the missing pagination behavior in the Discord bridge poll loop by draining consecutive full batches before sleeping. The control flow keeps the existing error and re-registration behavior intact while preventing overflow messages from being skipped under load.

### Findings
- None.

### Decision
Approving because the implementation matches the issue requirements, stays local to the bridge polling path, and avoids scope creep.

realproject7:
## RE2 Review — **APPROVE** ✅

Correct pagination fix:

1. **Inner drain loop** — `while True` re-fetches immediately when `len(messages) >= 50`, only `break`s on partial batch. This prevents silent message drops under high volume.

2. **Auth failure handling** — 401/403 `break`s out of inner loop and falls through to the outer `await asyncio.sleep(interval)`, so it retries next cycle. Correct.

3. **Cursor persistence per page** — `save_cursor` moved inside the inner loop, so progress is saved after each page, not just the final one. Prevents re-processing on crash mid-drain.

4. **No behavioral change under normal volume** — when <50 messages, single fetch + break + sleep, same as before.

5. **Exception safety** — outer `try/except` still wraps the entire inner loop, so a network error mid-drain doesn't crash the poller.

No issues found.

PR #430, opened 2026-04-11
[#415] Fix chat welcome screen persisting over existing messages

### LINKED ISSUE #415: [bug] Chat welcome screen still shows over existing messages (follow-up to #410)
## Problem

PR #411 fixed the API fallback path (\`ChatPanelAPI\`) to gate the welcome screen behind the initial history fetch. But the welcome screen still appears on page load and persists until the operator sends a message directly in the chat input. Two scenarios:

1. **On relaunch:** navigating to a project page after a server restart shows \"Ready when you are\" despite hundreds of messages in the channel.
2. **During live TG bridge activity:** messages arriving via Telegram bridge appear in AC's store but the chat panel doesn't update — the welcome screen stays until the operator types something directly.

## Likely root cause

The chat panel runs in **iframe mode** by default (embeds AC's own web UI from \`http://127.0.0.1:<port>\`). PR #411 only fixed the **API fallback** path (\`ChatPanelAPI\`). The iframe delegates rendering entirely to AC's own \`static/chat.js\`, which has its own empty-state / welcome screen logic that:
- May not fetch messages on initial load without user interaction
- May not receive WebSocket push updates for messages posted by the bridge (since the bridge uses HTTP \`POST /api/send\`, not the WebSocket path)
- May gate the message display behind a user interaction event

Since the iframe and QuadWork are on different ports (cross-origin), QuadWork can't control the iframe's internal state — the fix needs to be either:
- In AC's own \`static/chat.js\` (upstream fix in the per-project clone)
- Or force the API fallback mode instead of iframe mode, since the API fallback was already fixed in PR #411

## Suggested investigation

1. **Check if the iframe is actually loading AC's UI** — inspect the iframe src in devtools. If it 403s or times out, the 3s fallback to API mode should kick in. If the API fallback IS active and still shows the welcome screen, PR #411's fix has a bug.
2. **If iframe mode is active:** check AC's \`static/chat.js\` for the welcome/empty-state gate. Does it require a WebSocket \`open\` event before fetching? Does it require a user click? Does it have its own \`Ready when you are\` screen separate from QuadWork's?
3. **If API fallback is active:** the fix from PR #411 should be working. Check if the initial \`/api/chat\` fetch is returning messages or an error. If error (403 stale token), the messages array stays empty → welcome screen persists.
4. **TG bridge messages not triggering updates:** the API fallback polls on a timer. If the poll interval is too long or the poll isn't running while the welcome screen is shown, new messages from the bridge won't appear.

## Acceptance

- On page load with existing messages in the channel, the chat panel shows those messages within 1-2 seconds (no welcome screen flash).
- Messages arriving via Telegram bridge while the page is open appear in the chat panel automatically (no need to type first).
- The welcome screen only shows when the channel has genuinely zero messages.

## Branch

\`task/XXX-chat-welcome-screen-followup\`

## Related

- quadwork#410 (PR #411) — original fix that addressed the API fallback path only
- Dev's explanation on PR #411: \"The iframe path delegates rendering entirely to AC's own chat.js — if that UI has the same bug, it would need to be fixed in the agentchattr repo\"

realproject7:
## Summary
Fixes #415

The chat panel defaulted to iframe mode, embedding AC's own web UI. AC's internal `chat.js` has its own empty-state logic that QuadWork can't control, so "Ready when you are" persisted over existing messages and bridge traffic didn't appear until the operator typed directly.

Switched to using the API panel (`ChatPanelAPI`) directly. It has full feature parity (mentions, slash commands, replies, IME, auto-scroll) and correctly gates the welcome screen behind the initial history fetch (#410).

## Changes
- `ChatPanel.tsx` — removed iframe loading/fallback logic (-70 lines), `ChatPanel` now renders `ChatPanelAPI` directly

## Test plan
- [ ] On page load with existing messages: messages appear within 1-2s, no welcome screen flash
- [ ] Messages via Telegram/Discord bridge appear automatically without typing first
- [ ] Welcome screen only shows when channel has zero messages
- [ ] @mention autocomplete, slash commands, replies, Shift+Enter multiline all still work
- [ ] Korean/Japanese IME composition guard still works

### DIFF
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -90,82 +90,19 @@ function renderMessageWithMentions(text: string): React.ReactNode[] {
 }
 
 /**
- * Tries iframe first (uses AgentChattr's own session auth).
- * Falls back to API polling if iframe fails to load.
+ * #415: Use the API-driven chat panel directly. The earlier iframe
+ * approach embedded AC's own web UI but its internal chat.js has its
+ * own empty-state / welcome screen that QuadWork can't control,
+ * causing "Ready when you are" to persist over existing messages.
+ * The API panel (ChatPanelAPI) has full feature parity — mentions,
+ * slash commands, replies, auto-scroll, IME — and correctly gates
+ * the welcome screen behind the initial history fetch (#410).
  */
 interface ChatPanelProps {
   projectId?: string;
 }
 
 export default function ChatPanel({ projectId }: ChatPanelProps) {
-  const [mode, setMode] = useState<"iframe" | "api" | "loading">("loading");
-  const [chattrUrl, setChattrUrl] = useState("");
-  const [chattrToken, setChattrToken] = useState("");
-
-  // Resolve AgentChattr URL and token from per-project config (fallback to global)
-  useEffect(() => {
-    fetch("/api/config")
-      .then((r) => {
-        if (!r.ok) throw new Error("config fetch failed");
-        return r.json();
-      })
-      .then((cfg) => {
-        const project = projectId ? cfg.projects?.find((p: { id: string }) => p.id === projectId) : null;
-        setChattrUrl(project?.agentchattr_url || cfg.agentchattr_url || "http://127.0.0.1:8300");
-        setChattrToken(project?.agentchattr_token || cfg.agentchattr_token || "");
-      })
-      .catch(() => setChattrUrl("http://127.0.0.1:8300"));
-  }, [projectId]);
-
-  // Timeout fallback: if iframe hasn't loaded within 3s, switch to API mode
-  // (onError doesn't fire for CSP/X-Frame-Options blocks)
-  useEffect(() => {
-    if (mode !== "loading") return;
-    const timer = setTimeout(() => {
-      setMode((prev) => (prev === "loading" ? "api" : prev));
-    }, 3000);
-    return () => clearTimeout(timer);
-  }, [mode]);
-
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR is tightly scoped to #415 and removes the uncontrollable iframe path in favor of the existing API-driven chat panel that already fixed the welcome-screen gating in #410. The public component contract stays the same, so the dashboard integration remains unchanged while the underlying bug source is eliminated.

### Findings
- None.

### Decision
Approving because the implementation directly addresses the root cause described in #415, avoids unnecessary scope, and does not introduce a new dependency surface.

realproject7:
## RE2 Review — **APPROVE** ✅

Clean fix — removes the root cause rather than patching around it:

1. **Correct diagnosis** — the iframe embedded AC's own `chat.js` which has an uncontrollable welcome screen. No amount of QuadWork-side fixes could address it without an upstream change.

2. **Right solution** — `ChatPanelAPI` already has full feature parity (mentions, slash commands, replies, IME, auto-scroll) and its welcome screen was already fixed in #410 to gate behind the initial history fetch. Making `ChatPanel` a direct passthrough to `ChatPanelAPI` is the simplest correct fix.

3. **-70 lines** — entire iframe loading/fallback/timeout machinery removed. No dead code left behind.

4. **No functional regression risk** — the removed iframe path was the source of the bug; `ChatPanelAPI` handles its own config resolution and URL setup.

No issues found.

PR #431, opened 2026-04-11
[#426] Fix chat auto-prefix @head only checking start of message

### LINKED ISSUE #426: [bug] Chat auto-prefix @head on untagged operator messages no longer works
## Problem

Previously, when the operator sent a message without any \`@mention\`, the chat panel automatically prefixed it with \`@head\` so Head would pick it up. This behavior has stopped working — untagged messages now go through without any auto-prefix, meaning no agent sees them and the message is effectively lost.

The \`/command\` exemption (slash commands skip the auto-prefix) should be preserved.

## Evidence

Operator sent a multi-line batch assignment message without \`@head\` prefix. The message appeared in chat with no agent tag. Only \`@dev\` was highlighted because it appeared naturally in the message body, but no auto-prefix was added.

## Likely cause

One of the recent PRs may have regressed the auto-prefix logic:
- PR #382 (#236) rewrote \`sendViaWebSocket\` — the auto-prefix may have been in the old send path and wasn't carried over
- PR #361 (#344) converted the chat input from \`<input>\` to \`<textarea>\` — the \`onKeyDown\` / \`send()\` handler may have lost the prefix logic during the refactor
- Or the auto-prefix lived in the server-side \`/api/chat\` handler and was dropped during the WS migration

## Investigation

1. Search \`ChatPanel.tsx\` / \`ChatPanelAPI\` for the auto-prefix logic — was it in the \`send()\` function?
2. Check \`server/routes.js\` \`/api/chat\` handler — was the prefix applied server-side before forwarding to AC?
3. Check git blame on the relevant lines to find when it was removed

## Acceptance

- Operator sends a message without any \`@mention\` → message is auto-prefixed with \`@head\` before sending
- Operator sends a message with an explicit \`@mention\` (e.g. \`@dev do X\`) → no auto-prefix (already tagged)
- Operator sends a \`/command\` → no auto-prefix (slash commands are not chat messages)
- Auto-prefix is visible in the rendered message (operator sees \`@head <their message>\` in the chat log)

## Branch

\`task/XXX-chat-auto-prefix-head\`

realproject7:
## Summary
Fixes #426

The `@head` auto-prefix regex (`KNOWN_AGENT_RE`) searched the **entire** message for any `@agent` mention. If the operator wrote something like "Assign to @dev one at a time", the `@dev` in the body matched the regex and the `@head` prefix was skipped — the message had no routing target.

Changed to `STARTS_WITH_AGENT_RE` (`/^@(head|dev|re1|re2)\b/i`) so only an `@agent` at the **start** of the message suppresses the auto-prefix.

## Changes
- `ChatPanel.tsx` — renamed regex, anchored with `^`, updated comments

## Test plan
- [ ] Send message without any `@mention` → auto-prefixed with `@head`
- [ ] Send message starting with `@dev do X` → no auto-prefix (already routed)
- [ ] Send message with `@dev` in body but not at start → gets `@head` prefix
- [ ] Send `/continue` → no auto-prefix (slash command exempt)

### DIFF
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -218,18 +218,19 @@ function ChatPanelAPI({ projectId }: { projectId?: string }) {
   const send = () => {
     const raw = input.trim();
     if (!raw || sending) return;
-    // #228: if the operator didn't tag a known agent, prepend
-    // `@head ` so the message has somewhere to go. Unknown
-    // `@whatever` mentions don't count — Head is still added.
-    // Known agents: head, dev, re1, re2.
+    // #228: if the operator didn't direct the message to a known
+    // agent, prepend `@head ` so the message has somewhere to go.
+    // #426: only check the START of the message — a mention buried
+    // in the body (e.g. "Assign to @dev one at a time") is not a
+    // routing target and still needs the @head prefix.
     //
     // #410 / quadwork#276: slash commands are routed by AgentChattr
     // itself and must NOT be wrapped in `@head /continue` — that
     // would silently break them. Skip the auto-tag when the message
     // starts with `/`.
-    const KNOWN_AGENT_RE = /@(head|dev|re1|re2)\b/i;
+    const STARTS_WITH_AGENT_RE = /^@(head|dev|re1|re2)\b/i;
     const startsWithSlash = raw.startsWith("/");
-    const text = (startsWithSlash || KNOWN_AGENT_RE.test(raw)) ? raw : `@head ${raw}`;
+    const text = (startsWithSlash || STARTS_WITH_AGENT_RE.test(raw)) ? raw : `@head ${raw}`;
     setSending(true);
     setSendError(null);
     fetch(`/api/chat${projectId ? `?project=${encodeURIComponent(projectId)}` : ""}`, {


realproject7:
## RE2 Review — **APPROVE** ✅

Correct fix:

1. **Root cause** — `KNOWN_AGENT_RE` (unanchored) matched `@agent` anywhere in the message body, so "Assign to @dev one at a time" suppressed the `@head` prefix, leaving the message unrouted.

2. **Fix** — `STARTS_WITH_AGENT_RE` with `^` anchor only matches a leading `@agent`. Body mentions no longer suppress the auto-prefix. Correct behavior.

3. **Slash command exemption preserved** — `startsWithSlash` check unchanged.

4. **Regex rename** — `KNOWN_AGENT_RE` → `STARTS_WITH_AGENT_RE` clearly communicates the anchored intent.

No issues found.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR is tightly scoped to #426 and fixes the routing regression by only suppressing the auto-prefix when a known agent mention appears at the start of the message. Slash-command handling and the rest of the send path remain unchanged.

### Findings
- None.

### Decision
Approving because the implementation directly addresses the bug described in #426 without altering unrelated chat behavior.

PR #433, opened 2026-04-11
[#416] Add AC health monitor with auto-restart on crash detection

### LINKED ISSUE #416: [feat] Auto-restart AgentChattr on crash detection (health monitor)
## Problem

When a per-project AgentChattr Python process dies unexpectedly (OOM kill, unhandled exception, zombie from prior restart), QuadWork does not detect it or restart it. The project silently goes dark:

- Chat send fails with \`ECONNREFUSED\`
- Telegram bridge enters a 403/connection-refused loop
- Agent queue watchers can't deliver messages
- The operator sees \"Send failed: 502\" with no explanation of what broke

The operator has to manually notice the failure, diagnose that AC is down (not obvious from the dashboard), and click SERVER → Restart or run a curl command to respawn it.

## Evidence (2026-04-10, DropCast project)

- DropCast AC on port 8300 silently died at some point after message 194 in the chat log
- No crash log found — process just disappeared
- QuadWork server (port 8400) was healthy, quadwork AC (8301) was healthy
- Only discovered when operator tried to send a message and got \`ECONNREFUSED\`
- Manual \`POST /api/agentchattr/dropcast/restart\` immediately fixed it

## Proposal

Add a lightweight **AC health monitor** that runs inside the QuadWork Node server:

### Implementation

\`\`\`js
// For each project with a running AC:
// - Every 30s, check if the AC port is still listening
// - If not, attempt auto-restart
// - Log the crash + restart event
// - Surface it in the dashboard (e.g. a status indicator or toast)
\`\`\`

### Health check method

**Option A (recommended) — TCP connect probe:**
\`\`\`js
const net = require("net");
function isPortAlive(port) {
  return new Promise((resolve) => {
    const sock = net.createConnection({ port, host: "127.0.0.1" }, () => {
      sock.destroy();
      resolve(true);
    });
    sock.on("error", () => resolve(false));
    sock.setTimeout(2000, () => { sock.destroy(); resolve(false); });
  });
}
\`\`\`

Lightweight, no auth needed, no HTTP overhead. Just checks if something is listening on the port.

**Option B — HTTP health endpoint:**
Hit \`GET /api/channels\` or similar on each AC. More thorough (confirms AC is responsive, not just listening) but requires auth handling.

Recommend Option A for the monitor, with Option B as an optional deeper check after a restart.

### Monitor loop

\`\`\`js
setInterval(async () => {
  for (const project of activeProjects) {
    const port = getProjectAcPort(project.id);
    if (!port) continue;
    const alive = await isPortAlive(port);
    if (!alive) {
      console.warn(\`[health] AC for project ${project.id} on port ${port} is down — auto-restarting\`);
      await restartProjectAc(project.id);
    }
  }
}, 30000);
\`\`\`

### Restart logic

Reuse the existing \`/api/agentchattr/:project/restart\` handler internally. It already handles orphan cleanup (#386), port clearing, and respawn.

### Rate limiting

- Don't restart more than once per 60s per project (prevent crash loops)
- After 3 consecutive restart failures, stop trying and surface a persistent error in the dashboard
- Reset the failure counter on a successful health check

### Dashboard surface

- Show a small indicator in the SERVER section when an auto-restart happened: \`⚠ AC restarted automatically at HH:MM\`
- Optionally log to a per-project file (\`~/.quadwork/<project>/ac-health.log\`) for post-mortem

## Acceptance

- AC process dies unexpectedly → QuadWork detects within 30s and auto-restarts
- Auto-restart succeeds → agents reconnect, chat works, bridge recovers
- Auto-restart fails 3 times → stops retrying, shows persistent error in dashboard
- Normal SERVER → Stop does NOT trigger the health monitor (only unexpected crashes)
- Health monitor does not run for projects that are intentionally stopped
- Console log shows \`[health] AC for project X on port Y is down — auto-restarting\`
- No performance impact — one TCP connect per project per 30s is negligible

## Out of scope

- Monitoring agent CLI processes (Claude/Codex) — different lifecycle, handled by PTY reattach (#368)
- Monitoring the Telegram/Discord bridge processes — separate concern, could be a follow-up
- Monitoring the QuadWork Node server itself — that's the operator's responsibility (systemd/pm2/etc.)

## Files

- \`server/index.js\` — health monitor loop, project AC port resolution
- Possibly \`server/routes.js\` — if the restart handler needs to be refactored for internal vs API call

## Branch

\`task/XXX-ac-health-monitor-auto-restart\`

## Related

- quadwork#386 (PR #389) — orphan cleanup on restart (reused by the auto-restart path)
- quadwork#388 (PR #392) — stale slot cleanup (related lifecycle management)

realproject7:
## Summary
Fixes #416

Adds a lightweight AC health monitor that detects crashed AgentChattr processes and auto-restarts them.

### How it works
- **TCP connect probe** every 30s for each project with a "running" AC process
- If port is dead, auto-restarts using the same spawn logic as the manual restart endpoint
- **Rate limiting:** max one restart per 60s per project
- **Failure cap:** gives up after 3 consecutive failures, surfaces persistent error
- **Intentional stops** are not monitored (only unexpected crashes)

### Dashboard
- `GET /api/agentchattr/:project/health` returns health monitor state
- ServerSection polls every 30s and shows:
  - Yellow note: "AC auto-restarted at HH:MM" (for 5 minutes after auto-restart)
  - Red error: "AC auto-restart failed 3x — manual restart required"

## Changes
- `server/index.js` — health monitor loop, `isPortAlive()`, `acHealthCheck()`, health API endpoint (+162 lines)
- `src/components/ControlBar.tsx` — health polling + indicator in ServerSection (+35 lines)

## Test plan
- [ ] Kill an AC process manually → monitor detects within 30s and auto-restarts
- [ ] Auto-restart succeeds → agents reconnect, chat works
- [ ] Kill AC 3 times rapidly → after 3 failures, monitor stops trying and shows red error
- [ ] Click SERVER → Stop → monitor does NOT auto-restart (intentional stop)
- [ ] Console shows `[health] AC for project X on port Y is down — auto-restarting`
- [ ] Dashboard shows yellow auto-restart indicator briefly after recovery

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1086,6 +1086,23 @@ app.post("/api/agentchattr/:projectOrAction", handleAgentChattr);
 // AgentChattr doesn't expose staleness metadata, so this clears all slots.
 // Agents' wrapper heartbeat will auto-re-register with clean names.
 
+// #416: AC health status endpoint — returns the health monitor state
+// for a project so the dashboard can surface auto-restart events.
+app.get("/api/agentchattr/:project/health", (req, res) => {
+  const projectId = req.params.project;
+  const proc = chattrProcesses.get(projectId);
+  const health = _acHealth.state.get(projectId) || { lastRestart: 0, consecutiveFailures: 0 };
+  res.json({
+    state: proc?.state || "unknown",
+    error: proc?.error || null,
+    autoRestart: {
+      lastRestart: health.lastRestart || null,
+      consecutiveFailures: health.consecutiveFailures,
+      gaveUp: health.consecutiveFailures >= 3,
+    },
+  });
+});
+
 app.post("/api/agents/:project/reset", async (req, res) => {
   const projectId = req.params.project;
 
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The health monitor direction is right, but the new auto-restart path duplicates the restart logic instead of actually reusing the existing manual restart flow. That loses the orphan-port cleanup and port-clear wait guarantees the manual path already relies on.

### Findings
- [medium] Auto-restart skips the existing restart safeguards from `#386`, so crash recovery can fail when the port is still occupied by a stale/orphaned listener.
  - File: `server/index.js:1851`
  - Suggestion: Factor the existing restart sequence into a shared helper and call it from both the manual restart endpoint and the health monitor. Right now the monitor only kills `proc.process` and spawns immediately, but the manual restart path also calls `killProcessOnPort(chattrPort)` and `waitForPortFree(chattrPort, 3000)` before respawn (`server/index.js:951-960`). The issue explicitly asks to reuse the existing restart handler internally, and without those steps the auto-restart path can regress the exact stale/orphan process scenarios that `#386` fixed.

### Decision
Requesting changes because the health monitor currently bypasses the hardened restart path instead of reusing it, which makes automatic crash recovery less reliable than the manual restart flow.

realproject7:
## RE2 Review — **REQUEST CHANGES**

The health monitor structure is solid (TCP probe, rate limiting, failure cap, dashboard indicators), but there's a functional gap:

### Issue: Auto-restart doesn't reset agents

After the health monitor auto-restarts AC, agents are left with stale MCP tokens — the same problem PR #432 just fixed for the SERVER Restart button. The manual SERVER Restart now chains a `/api/agents/:project/reset` call (ControlBar.tsx), but the server-side health monitor has no equivalent.

When AC crashes and auto-restarts, agents need fresh MCP connections. Without this, the auto-restart recovers AC but agents stay stuck with "MCP still disconnected", defeating the purpose of automatic recovery.

**Suggested fix:** After a successful auto-restart spawn + token sync, call the same agent reset logic (iterate `agentSessions`, stop + respawn). Something like:

```js
// After syncChattrToken timeout, also reset agents
setTimeout(async () => {
  // same logic as POST /api/agents/:project/reset
  for (const [key, session] of agentSessions) {
    if (key.startsWith(`${project.id}/`)) {
      await stopAgentSession(key);
      await spawnAgentPty(project.id, session.agentId || key.split('/')[1]);
    }
  }
}, 4000); // after token sync completes
```

### Minor note (not blocking)

The spawn logic (~40 lines) is duplicated from the `spawnChattr` closure inside `handleAgentChattr`. This works but is a maintenance risk — if the spawn logic changes (e.g. new config.toml fields, new pre-spawn steps), the health monitor's copy could diverge. Consider extracting `spawnChattr` to a shared helper in a follow-up.

### What's good

- TCP probe is the right approach — lightweight, no auth needed
- Rate limiting (60s cooldown) and failure cap (3x) prevent crash loops
- Intentional stops correctly excluded (`proc.state === "stopped"`)
- Dashboard indicators (yellow/red) with 5-minute auto-clear
- Cursor persistence and health endpoint are clean

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The restart-path reuse fix is correct, but the health monitor still stops at restarting AgentChattr itself. After an AC crash, that is not enough to satisfy the project's own recovery behavior from #417.

### Findings
- [medium] Auto-restart does not reset agent sessions after AC comes back, so agents can stay stuck on stale MCP credentials.
  - File: `server/index.js:1855`
  - Suggestion: After a successful internal restart call, trigger the same `/api/agents/:project/reset` recovery step that the manual SERVER Restart button uses (see `src/components/ControlBar.tsx:63-75`). Right now the monitor only calls `/api/agentchattr/:project/restart`, but #417 established that AC restart alone leaves agents stuck with stale MCP connections after a crash. The acceptance here says auto-restart should recover agents/chat end-to-end, so the monitor needs the same chained reset behavior.

### Decision
Requesting changes because the current auto-restart flow still restores only the AC process, not the working agent connectivity the issue requires.

realproject7:
## RE2 Re-review — **APPROVE** ✅

Both issues from initial review addressed:

1. **Agent reset chained** — after successful AC auto-restart, the health monitor now calls `POST /api/agents/:project/reset` to give agents fresh MCP tokens. Matches the manual SERVER Restart button chain from PR #432.

2. **No more duplicated spawn logic** — replaced the inline spawn code with a localhost `POST /api/agentchattr/:project/restart` call. This reuses the hardened restart path (killProcessOnPort, waitForPortFree, snapshot, auto-restore, regenerateConfigToml) instead of a simplified copy. Much better for maintenance.

3. **Error handling** — restart failure sets `chattrProcesses` error state and logs. Agent reset failure is logged but doesn't block (correct — AC itself is back, agents are a best-effort follow-up).

Overall: clean health monitor with correct recovery chain. Approve.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review addresses both outstanding recovery gaps: the health monitor now reuses the hardened AC restart path instead of duplicating spawn logic, and it also chains the agent reset step so agents recover fresh MCP connectivity after AC comes back.

### Findings
- None.

### Decision
Approving because the auto-restart flow now matches the established manual recovery behavior from #416/#417 closely enough to satisfy the issue acceptance criteria.

PR #450, opened 2026-04-11
[#448] Sync chat token on GET /api/chat 403 + startup retry

### LINKED ISSUE #448: [bug] Chat panel stuck on 'Loading messages' — session token not synced on startup after AC restart
## Problem

After a QuadWork stop/start (or AC restart), the chat panel shows \"Loading messages...\" indefinitely. The \`/api/chat\` proxy returns 403 because the stored \`agentchattr_token\` in \`config.json\` is stale — it was issued by the previous AC instance and the new AC doesn't recognize it.

\`syncChattrToken()\` in \`server/config.js:187\` correctly extracts the fresh token from AC's HTML (\`__SESSION_TOKEN__\`), but it's **never called on startup**. It only fires inside the \`/api/chat\` handler's 401-retry path, which requires a specific \`EAGENTCHATTR_401\` error code from the WebSocket — a plain 403 from the proxy doesn't trigger the sync.

## Evidence (2026-04-11)

- AC running on port 8301, fresh session token: \`4aa5005b...\`
- QuadWork config.json stored token: \`588c7fbb...\` (stale, from pre-restart)
- \`curl /api/chat?project=quadwork\` → \`{"error":"AgentChattr returned 403"}\`
- \`curl http://127.0.0.1:8301/\` → HTML contains \`__SESSION_TOKEN__="4aa5005b..."\`
- \`syncChattrToken\` never called → stale token persists → all chat fetches fail → \"Loading messages...\" forever
- Full stop/start of QuadWork does NOT fix it — startup doesn't call syncChattrToken either

## Root cause

\`syncChattrToken(projectId)\` is defined but only called from:
1. The \`/api/chat\` POST handler's 401-retry branch (catches \`EAGENTCHATTR_401\` from WebSocket close code 4003)
2. Nowhere else — not on startup, not after AC restart, not on the GET chat poll path

The GET \`/api/chat\` poll (which the ChatPanelAPI component uses for initial fetch) goes through a different code path that returns a generic 403 error without triggering the sync.

## Fix

Call \`syncChattrToken(projectId)\` at two critical points:

1. **On QuadWork startup** — after each project's AC server is confirmed ready, sync the token before any chat polls fire
2. **On AC restart** (health monitor or SERVER → Restart) — sync the token immediately after the new AC is up
3. **On GET \`/api/chat\` 403** — the GET poll path should also trigger a token re-sync + retry, not just the POST path

Additionally, the PR #440 retry logic (\`MAX_INITIAL_RETRIES = 3\`) should be aware of this: retrying a fetch with a stale token will fail all 3 times. Either sync the token before retrying, or increase the retry window to wait for the server-side sync.

## Acceptance

- QuadWork stop/start → chat panel shows existing messages on page load (not \"Loading messages...\")
- AC crashes and auto-restarts via health monitor → chat panel recovers within one poll cycle (~3s)
- Operator clicks SERVER → Restart → chat panel recovers within one poll cycle
- No manual page refresh needed
- \`syncChattrToken\` is called proactively, not only reactively on 401

## Files

- \`server/config.js\` — \`syncChattrToken()\` (already works, just needs to be called)
- \`server/index.js\` — startup sequence, AC restart callback, health monitor callback
- \`server/routes.js\` — GET \`/api/chat\` handler (add 403 → sync → retry)

## Branch

\`task/XXX-sync-chat-token-on-startup\`

## Related

- quadwork#436 (PR #440) — chat welcome screen retry logic (retries the fetch but with the same stale token)
- quadwork#447 — AC restart agent reset (related lifecycle issue, different scope)
- quadwork#415 (PR #430) — removed iframe, switched to API-only (made the token dependency more critical)

realproject7:
## Summary

- Chat panel showed "Loading messages..." indefinitely after AC restarts — the stored `agentchattr_token` was stale and no code path re-synced it for the GET `/api/chat` poll
- **GET `/api/chat`**: Added 401/403 → `syncChattrToken()` → retry-once logic, mirroring the existing POST handler's retry pattern. This creates a self-healing loop: even if the startup sync misses, the first chat poll auto-fixes the token
- **Startup sync**: Added a 5s delayed retry so projects whose AC isn't listening at boot time still get their token synced (the original fire-and-forget call silently failed if AC wasn't ready)

## What changed

- `server/routes.js` (GET `/api/chat`): On 401/403 response, call `syncChattrToken(projectId)`, re-read the config, and retry the fetch with the refreshed token
- `server/index.js` (startup): Wrap `syncChattrToken()` with `.catch()` that schedules a 5s delayed retry

## Test plan

- [ ] Stop QuadWork + AC, restart both — chat panel loads messages on first page load (no "Loading messages...")
- [ ] Kill AC process manually, let health monitor restart it — chat panel recovers within one poll cycle (~3s)
- [ ] Click SERVER → Restart — chat panel recovers within one poll cycle
- [ ] Verify no double-sync on normal startup when AC is already running (token matches, no unnecessary config write)

Fixes #448

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1926,9 +1926,12 @@ server.listen(PORT, "127.0.0.1", () => {
   // Sync AgentChattr tokens for all projects on startup and backfill
   // the sender-overflow CSS/JS patch (#402) so already-running AC
   // instances receive the fix without requiring a restart.
+  // #448: retry after 5s for projects where AC isn't up yet at boot.
   const startupCfg = readConfig();
   for (const p of (startupCfg.projects || [])) {
-    syncChattrToken(p.id);
+    syncChattrToken(p.id).catch(() => {
+      setTimeout(() => syncChattrToken(p.id).catch(() => {}), 5000);
+    });
     const { dir: acDir } = resolveProjectChattr(p.id);
     if (acDir) patchAgentchattrCss(acDir);
   }
--- a/server/routes.js
+++ b/server/routes.js
@@ -113,18 +113,32 @@ function chatAuthHeaders(token) {
 }
 
 router.get("/api/chat", async (req, res) => {
+  const projectId = req.query.project;
   const apiPath = req.query.path || "/api/messages";
-  const { url: base, token } = getChattrConfig(req.query.project);
+  const { url: base, token } = getChattrConfig(projectId);
 
-  const fwd = new URLSearchParams();
-  for (const [k, v] of Object.entries(req.query)) {
-    if (k !== "path") fwd.set(k, String(v));
-  }
-  if (token) fwd.set("token", token);
+  const buildUrl = (tok) => {
+    const fwd = new URLSearchParams();
+    for (const [k, v] of Object.entries(req.query)) {
+      if (k !== "path") fwd.set(k, String(v));
+    }
+    if (tok) fwd.set("token", tok);
+    return `${base}${apiPath}?${fwd.toString()}`;
+  };
 
-  const url = `${base}${apiPath}?${fwd.toString()}`;
   try {
-    const r = await fetch(url, { headers: chatAuthHeaders(token) });
+    const r = await fetch(buildUrl(token), { headers: chatAuthHeaders(token) });
+    // #448: on 401/403, re-sync the session token from AC and retry
+    // once. The stored token may be stale after an AC restart.
+    if ((r.status === 401 || r.status === 403) && projectId) {
+      try { await syncChattrToken(projectId); } catch {}
+      const { token: refreshed } = getChattrConfig(projectId);
+      if (refreshed && refreshed !== token) {
+        const retry = await fetch(buildUrl(refreshed), { headers: chatAuthHeaders(refreshed) });
+        if (!retry.ok) return res.status(retry.status).json({ error: `AgentChattr returned ${retry.status}` });
+        return res.json(await retry.json());
+      }
+    }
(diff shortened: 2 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The intended GET `/api/chat` token-resync path is correct in concept, but the implementation currently introduces a runtime failure in the exact stale-token case this PR is meant to fix.

### Findings
- [high] `syncChattrToken` is called in the new 401/403 recovery branch, but `server/routes.js` does not import it from `./config`. On the first stale-token response, `await syncChattrToken(projectId)` throws `ReferenceError: syncChattrToken is not defined`, so the handler returns 502 instead of self-healing.
  - File: `server/routes.js:81`
  - File: `server/routes.js:134`
  - Suggestion: import `syncChattrToken` alongside `resolveProjectChattr` and `sanitizeOperatorName`, then re-test the 401/403 retry flow.

### Decision
Request changes. Until `syncChattrToken` is actually in scope in `server/routes.js`, the new recovery path breaks the stale-token case rather than fixing it.

project7-interns (approved):
## Verdict: APPROVE

### Summary
Re-reviewed after the follow-up on import placement. The GET `/api/chat` retry path and the startup retry together cover the stale-token recovery path described in issue #448.

### Findings
- None.

### Decision
Approve. My previous concern about `syncChattrToken` scope was incorrect; the import later in `server/routes.js` is still in scope by the time the registered route handler runs.

PR #459, opened 2026-04-12
[#457] Migrate bridge slugs in AC config on startup and install

### LINKED ISSUE #457: [bug] Discord bridge still registers as 'discord-bridge' — AC configs not migrated to [agents.dc] (follow-up to #439)
## Problem

PR #451 (#439) renamed the bridge slug from \`discord-bridge\` → \`dc\` in the bridge script and templates, but the **existing per-project AC \`config.toml\` files were not migrated**. All 3 projects still have:

\`\`\`toml
[agents.discord-bridge]
\`\`\`

instead of:

\`\`\`toml
[agents.dc]
\`\`\`

So when the bridge tries to register as \`dc\`, AC rejects it (\`unknown base: dc\`), and the bridge falls back to whatever base AC does accept — resulting in the long \`discord-bridge\` name still appearing in the chat sender column, overflowing as before.

**Also check the Telegram bridge** — same migration pattern. Verify all 3 projects have \`[agents.tg]\` not \`[agents.telegram-bridge]\`.

## Evidence

\`\`\`
/Users/cho/.quadwork/dropcast/agentchattr/config.toml:49:[agents.discord-bridge]
/Users/cho/.quadwork/plotlink/agentchattr/config.toml:46:[agents.discord-bridge]
/Users/cho/.quadwork/quadwork/agentchattr/config.toml:49:[agents.discord-bridge]
\`\`\`

## Fix

The \`patchAgentchattrConfigForDiscordBridge()\` and \`patchAgentchattrConfigForTelegramBridge()\` functions in \`server/routes.js\` need to:

1. **Rename** existing \`[agents.discord-bridge]\` → \`[agents.dc]\` (not just check for presence)
2. **Rename** existing \`[agents.telegram-bridge]\` → \`[agents.tg]\` (same pattern)
3. Run the migration on **Install Bridge** and on **startup** so existing projects are patched automatically

Additionally, after patching the AC config, the AC server must be **restarted** so it loads the new agent slugs. The patch function should return a flag indicating whether a restart is needed.

## Acceptance

- All 3 projects' AC configs have \`[agents.dc]\` and \`[agents.tg]\` (not the old long names)
- Discord bridge registers as \`dc\` in AC, sender column shows \`dc\`
- Telegram bridge registers as \`tg\`, sender column shows \`tg\`
- No sender column overflow from bridge names
- Migration is idempotent — running twice doesn't duplicate sections

## Files

- \`server/routes.js\` — \`patchAgentchattrConfigForDiscordBridge()\` and \`patchAgentchattrConfigForTelegramBridge()\` — add rename logic
- Existing per-project AC configs at \`~/.quadwork/*/agentchattr/config.toml\`

## Branch

\`task/XXX-migrate-bridge-slugs-in-ac-config\`

## Related

- quadwork#439 (PR #451) — original slug rename (missed the AC config migration)
- quadwork#383 — Telegram bridge AC config patching (same pattern)

realproject7:
## Summary
- Adds startup-time migration that renames `[agents.discord-bridge]` → `[agents.dc]` and `[agents.telegram-bridge]` → `[agents.tg]` in all project AC configs
- Both bridge install handlers now restart AC after patching configs so the new slugs load immediately
- Exports `projectAgentchattrConfigPath` from routes.js for use in startup migration

Fixes #457

## Test plan
- [ ] Verify all 3 project AC configs have `[agents.dc]` and `[agents.tg]` after server startup
- [ ] Verify Discord bridge registers as `dc` in AC chat
- [ ] Verify Telegram bridge registers as `tg` in AC chat
- [ ] Verify migration is idempotent (running twice doesn't duplicate sections)
- [ ] Verify AC auto-restarts after config migration

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1957,10 +1957,8 @@ server.listen(PORT, "127.0.0.1", () => {
         // Restart AC so it loads the new agent slugs
         setTimeout(async () => {
           try {
-            const r = await fetch(`http://127.0.0.1:${PORT}/api/agentchattr/${encodeURIComponent(p.id)}`, {
+            const r = await fetch(`http://127.0.0.1:${PORT}/api/agentchattr/${encodeURIComponent(p.id)}/restart`, {
               method: "POST",
-              headers: { "Content-Type": "application/json" },
-              body: JSON.stringify({ action: "restart" }),
             });
             if (r.ok) console.log(`[bridge-migrate] ${p.id}: restarted AC`);
             else console.warn(`[bridge-migrate] ${p.id}: AC restart returned ${r.status}`);
--- a/server/index.js
+++ b/server/index.js
@@ -8,6 +8,11 @@ const pty = require("node-pty");
 const { spawn } = require("child_process");
 const { readConfig, resolveAgentCwd, resolveAgentCommand, resolveProjectChattr, resolveChattrSpawn, syncChattrToken, CONFIG_PATH } = require("./config");
 const routes = require("./routes");
+const {
+  patchAgentchattrConfigForDiscordBridge,
+  patchAgentchattrConfigForTelegramBridge,
+  projectAgentchattrConfigPath,
+} = routes;
 const { waitForAgentChattrReady, registerAgent, deregisterAgent, startHeartbeat, stopHeartbeat } = require("./agentchattr-registry");
 const { patchAgentchattrCss } = require("./install-agentchattr");
 const { startQueueWatcher, stopQueueWatcher } = require("./queue-watcher");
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The slug migration logic is in place, but the restart hook is wired to the wrong AgentChattr lifecycle endpoint. As written, migrated configs are not reloaded, so the new `dc`/`tg` slugs will not take effect automatically.

### Findings
- [high] Restart calls never hit the restart action, so migrated configs stay unloaded until a manual restart.
  - File: `server/index.js:1957`, `server/routes.js:2581`, `server/routes.js:3037`
  - Suggestion: call the existing per-project lifecycle route as `POST /api/agentchattr/<project>/restart` (or change the handler to read `req.body.action`). The current requests post to `/api/agentchattr/<project>` with `{ action: "restart" }`, but `handleAgentChattr()` only reads the action from the URL path (`server/index.js:108-202`). That means these calls resolve to the backward-compat single-segment route instead of `restart`, so the migrated AC process never reloads the patched config.

### Decision
Requesting changes because this misses a required acceptance criterion: after patching config, AC must restart so the new bridge slugs load immediately.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit fixes the lifecycle call shape so AgentChattr is restarted through the existing per-project restart endpoint after config migration.

### Findings
- None.

### Decision
Approving because the previously blocking restart regression is resolved and the PR now satisfies the required automatic reload behavior for migrated `dc`/`tg` slugs.

PR #463, opened 2026-04-12
[#462] Fix auto-trigger first-poll gap on already-complete batch

### LINKED ISSUE #462: [bug] Auto-trigger doesn't stop when batch is already complete on page load / refresh
## Problem

The auto-trigger's batch lifecycle detection requires a \`complete: false → true\` **transition** to fire the auto-stop. But if the page is refreshed, navigated away/back, or the component remounts after the batch has already completed, \`prevBatchRef\` resets to \`null\`. The first poll sees \`complete: true\`, sets \`prev = { complete: true, hasItems: true }\`, and no transition is ever detected — the trigger keeps running indefinitely into a completed batch, wasting tokens.

**Observed:** QuadWork Batch 17 shows ✅ COMPLETE but the Scheduled Trigger shows Running (auto) with 16m remaining. PlotLink Batch 1 correctly auto-stopped (likely because the page was open during the actual completion transition).

## Root cause

In \`ScheduledTriggerWidget.tsx\`, the first-poll handler at \`checkBatchLifecycle()\`:

\`\`\`js
if (!prev) {
  // First poll — if there's an active non-complete batch, auto-start
  if (hasItems && !data.complete && !triggerRef.current?.enabled) {
    // auto-start
  }
  return; // ← sets prev and returns without checking the "already complete" case
}
\`\`\`

The first-poll block only handles the \"active batch → auto-start\" case. It does NOT handle the **\"batch already complete + trigger already running → auto-stop\"** case. After the first poll returns, subsequent polls check for transitions but \`prev.complete === data.complete === true\` — no transition, no stop.

## Fix

Add a check in the first-poll block for the \"already complete while trigger is running\" case:

\`\`\`js
if (!prev) {
  // First poll — auto-start if active batch exists
  if (hasItems && !data.complete && !triggerRef.current?.enabled) {
    // auto-start (existing)
  }
  // First poll — auto-stop if batch is already complete and trigger is running
  if (hasItems && data.complete && triggerRef.current?.enabled) {
    await fetch(\`/api/triggers/\${projectId}/stop\`, { method: "POST" });
    setAutoTriggered(false);
    setAutoStatus("Batch complete — trigger paused. Waiting for next batch.");
    await load();
  }
  return;
}
\`\`\`

Same logic applies to the Keep Mac Awake auto-toggle (#441 / PR #454) — check if it has the same gap.

## Acceptance

- Batch completes while page is open → trigger auto-stops (existing, works)
- Batch completes while page is NOT open, user returns → trigger auto-stops within 30s of page load
- Page refresh during a completed batch with trigger running → trigger auto-stops within 30s
- Server restart during a completed batch with auto ON → trigger does NOT auto-start (batch is complete)
- Same fix applied to Keep Mac Awake auto-toggle if affected

## Files

- \`src/components/ScheduledTriggerWidget.tsx\` — \`checkBatchLifecycle()\` first-poll block
- \`src/components/ControlBar.tsx\` — Keep Mac Awake auto-toggle (check for same gap)

## Branch

\`task/XXX-auto-trigger-stop-on-already-complete\`

## Related

- quadwork#408 (PR #412) — original auto-trigger implementation
- quadwork#441 (PR #454) — Keep Mac Awake auto-toggle (same pattern, may have same bug)

realproject7:
## Summary
- Fix first-poll gap in `ScheduledTriggerWidget.tsx` `checkBatchLifecycle()`: auto-stop trigger when batch is already complete on page load/refresh
- Fix same gap in `ControlBar.tsx` Keep Mac Awake auto-toggle: auto-stop caffeinate when batch is already complete on remount
- Both fixes are in the `!prev` (first-poll) block — adds check for "already complete + still running → auto-stop"

Fixes #462

## Test plan
- [ ] Batch completes while page is open → trigger auto-stops (existing, regression check)
- [ ] Batch completes while page is NOT open, user returns → trigger auto-stops within 30s
- [ ] Page refresh during completed batch with trigger running → trigger auto-stops within 30s
- [ ] Server restart during completed batch with auto ON → trigger does NOT auto-start
- [ ] Same behavior verified for Keep Mac Awake auto-toggle

### DIFF
--- a/src/components/ControlBar.tsx
+++ b/src/components/ControlBar.tsx
@@ -405,6 +405,11 @@ function SystemSection({ projectId }: { projectId: string }) {
             autoStart();
             setAwakeAutoStatus("Batch active — auto-started caffeinate.");
           }
+          // #462: First poll — batch already complete but caffeinate still running → auto-stop
+          if (hasItems && data.complete && activeRef.current) {
+            autoStop();
+            setAwakeAutoStatus("Batch complete — awake paused.");
+          }
           return;
         }
 
--- a/src/components/ScheduledTriggerWidget.tsx
+++ b/src/components/ScheduledTriggerWidget.tsx
@@ -346,6 +346,13 @@ export default function ScheduledTriggerWidget({ projectId }: ScheduledTriggerWi
           });
           await load();
         }
+        // #462: First poll — batch already complete but trigger still running → auto-stop
+        if (hasItems && data.complete && triggerRef.current?.enabled) {
+          await fetch(`/api/triggers/${encodeURIComponent(projectId)}/stop`, { method: "POST" });
+          setAutoTriggered(false);
+          setAutoStatus("Batch complete — trigger paused. Waiting for next batch.");
+          await load();
+        }
         return;
       }
 


project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR fixes the reported first-poll lifecycle gap in both the scheduled trigger and Keep Mac Awake flows. The new logic is scoped to the initial `!prev` pass and correctly stops already-running automation when the batch is already complete on remount or refresh.

### Findings
- None.

### Decision
Approving because the change is focused, matches issue #462, and resolves the missing auto-stop path without altering the existing transition-based lifecycle behavior.

PR #464, opened 2026-04-12
[#461] Fix blank idle agent terminals after page navigation

### LINKED ISSUE #461: [bug] Idle agent terminals blank after page navigation despite #418 scrollback fix
## Problem

PR #434 (#418) added a 64KB scrollback ring buffer that replays on WebSocket reconnect. Active agents (e.g. HEAD mid-task) correctly show content after navigating away and back. But **idle agents** (DEV, RE1, RE2 waiting for assignments) show completely blank terminal panels — no prompt, no startup banner, nothing.

This was observed on the PlotLink project page where HEAD was actively working (content visible) while the other 3 agents showed blank panels with green \"online\" dots.

## Root cause

The scrollback buffer resets to \`Buffer.alloc(0)\` every time an agent's PTY session is created (\`server/index.js:550\`). After an AC restart → agent reset chain (#447 / PR #449), all agents get fresh sessions with empty buffers. Active agents immediately produce output that fills the buffer. Idle agents produce only their CLI startup banner (Claude Code version line, model info, prompt) — which is brief and may have been:

1. **Produced before any WS client connected** — the PTY outputs the startup banner, the buffer captures it, but the first browser WS connect may replay it. On subsequent navigations, the WS reconnects and replays successfully. This should work.

2. **Overwritten by the ring buffer trim** — unlikely at 64KB, startup banners are ~500 bytes.

3. **Lost during a WS handler cleanup race** — if multiple WS connections cycle (navigate away → back → away → back), the \`session.ws\` reassignment may not properly clean up old data handlers, causing output to be sent to a closed WS instead of the new one.

More likely: the scrollback replay IS working but the idle agents genuinely have near-empty buffers. The CLI startup banner was produced, replayed on first connect, and the xterm rendered it. But when the user navigated away, the xterm instance was destroyed. On return, a new xterm connects, replays the buffer (just the startup banner), but the xterm might not be rendering it due to a display issue (e.g. the terminal size hasn't been negotiated yet when the replay arrives).

## Investigation

1. **Check buffer sizes** — add a temporary log: \`console.log(agentId, session.scrollback.length)\` at the WS connect handler. If idle agents have >0 bytes, the buffer exists but isn't rendering. If 0 bytes, the buffer is genuinely empty (separate issue).

2. **Check WS replay timing** — the scrollback is sent immediately on WS open (\`index.js:1642\`). But the client may not have attached xterm's data handler yet at that point. Check if there's a race between \`ws.send(session.scrollback)\` and the client-side \`ws.onmessage\` registration.

3. **Check multiple WS cleanup** — when navigating away, the old WS closes. On return, a new WS opens. Verify the old \`dataHandler\` (PTY → WS) is properly disposed so output goes to the new WS, not the dead one.

## Possible fixes

1. **Delay the scrollback replay** — send it 100-200ms after WS open to ensure the client-side xterm + onmessage handler is ready.

2. **Send a synthetic status line for empty buffers** — if scrollback is empty, send a minimal status line like \`\\x1b[2m[agent online — waiting for input]\\x1b[0m\\r\\n\` so the terminal isn't completely blank.

3. **Re-request replay from client** — after the client's xterm is fully initialized, send a \`{type: "replay"}\` message to the server which then sends the scrollback. This eliminates the timing race.

## Acceptance

- Navigate away from project page and back → all 4 agent terminals show content (either real scrollback or a synthetic \"online\" indicator)
- Active agents show their full scrollback (existing behavior, works)
- Idle agents show at least their CLI startup banner (model info, prompt line)
- No blank terminal panels unless the agent session is genuinely dead

## Files

- \`server/index.js\` — WS terminal handler (\`/ws/terminal\`), scrollback replay timing
- \`src/components/TerminalPanel.tsx\` — WS connect + xterm initialization timing

## Branch

\`task/XXX-terminal-scrollback-idle-agents\`

## Related

- quadwork#418 (PR #434) — original scrollback fix (works for active agents, insufficient for idle)
- quadwork#447 (PR #449) — AC restart agent reset (resets scrollback buffers)

realproject7:
## Summary
- Add client-initiated `{"type":"replay"}` request in `TerminalPanel.tsx` sent after xterm + onmessage are fully initialized, eliminating the timing race where server-sent scrollback arrived before the client could process it
- Add `replay` message handler in `server/index.js` WS terminal handler that sends the scrollback buffer on request
- For genuinely empty buffers (idle agents with no output), send a synthetic `[agent online — waiting for input]` indicator so terminals are never blank
- Server still sends scrollback eagerly on connect as best-effort for the common case

Fixes #461

## Test plan
- [ ] Navigate away from project page and back → all 4 agent terminals show content
- [ ] Active agents show their full scrollback
- [ ] Idle agents show at least startup banner or "[agent online]" indicator
- [ ] No blank terminal panels unless the agent session is genuinely dead
- [ ] Page refresh during active session preserves scrollback

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1664,6 +1664,19 @@ wss.on("connection", async (ws, req) => {
         session.term.resize(parsed.cols, parsed.rows);
         return;
       }
+      // #461: client requests scrollback replay after xterm is fully
+      // initialized. This eliminates the timing race where the server
+      // sends scrollback before the client's onmessage handler is ready.
+      // If the buffer is empty (idle agent with no output yet), send a
+      // synthetic status line so the terminal isn't completely blank.
+      if (parsed.type === "replay") {
+        if (session.scrollback && session.scrollback.length > 0) {
+          ws.send(session.scrollback);
+        } else {
+          ws.send(`\x1b[2m[agent online — waiting for input]\x1b[0m\r\n`);
+        }
+        return;
+      }
     } catch {}
     session.term.write(str);
   });
--- a/server/index.js
+++ b/server/index.js
@@ -1641,11 +1641,9 @@ wss.on("connection", async (ws, req) => {
   // Attach WS to session
   session.ws = ws;
 
-  // #418: replay scrollback buffer so the terminal isn't blank on reconnect.
-  // xterm.js processes ANSI escapes from the buffer the same as live data.
-  if (session.scrollback && session.scrollback.length > 0) {
-    ws.send(session.scrollback);
-  }
+  // #418/#461: scrollback replay is now client-initiated via
+  // {"type":"replay"} to avoid the timing race where eager replay
+  // arrived before the client's onmessage handler was registered.
 
   // PTY → client
   const dataHandler = session.term.onData((data) => {
--- a/src/components/TerminalPanel.tsx
+++ b/src/components/TerminalPanel.tsx
@@ -199,10 +199,8 @@ export default function TerminalPanel({
           })
         );
         // #461: request scrollback replay after xterm + onmessage are
-        // ready. The server also sends scrollback immediately on WS
-        // connect, but that can arrive before onmessage is registered
-        // (race). This explicit request guarantees the replay arrives
-        // when the client can process it.
+        // ready. This eliminates the timing race where server-sent
+        // scrollback arrived before the client could process it.
         ws.send(JSON.stringify({ type: "replay" }));
       };
 
--- a/src/components/TerminalPanel.tsx
+++ b/src/components/TerminalPanel.tsx
@@ -198,6 +198,12 @@ export default function TerminalPanel({
             rows: term.rows,
           })
         );
+        // #461: request scrollback replay after xterm + onmessage are
+        // ready. The server also sends scrollback immediately on WS
+        // connect, but that can arrive before onmessage is registered
+        // (race). This explicit request guarantees the replay arrives
+        // when the client can process it.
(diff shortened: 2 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The replay handshake fixes the timing race for empty/idle terminals, but it introduces a reconnect regression for any session with existing scrollback.

### Findings
- [high] Scrollback is now sent twice on every reconnect for non-empty sessions.
  - File: `server/index.js:1644`, `server/index.js:1672`, `src/components/TerminalPanel.tsx:192`
  - Suggestion: choose a single replay path. Either remove the eager server-side replay on connect and rely on the client `{"type":"replay"}` request, or gate the explicit replay so it only happens when the initial eager path is absent/empty. As written, the server still sends `session.scrollback` immediately on WS connect and then sends it again when the client always posts `{"type":"replay"}` in `onopen`, so active terminals will duplicate their full buffer after navigation/refresh.

### Decision
Requesting changes because this causes a visible terminal regression on reconnect for active agents, which is part of the acceptance criteria for preserving full scrollback cleanly.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit removes the eager server-side scrollback replay, so reconnects now use a single client-initiated replay path after the terminal is fully initialized.

### Findings
- None.

### Decision
Approving because the duplicate-replay regression is resolved and the PR now cleanly addresses the blank idle terminal issue from #461.

PR #465, opened 2026-04-12
[#458] Fix Discord bridge AC→Discord message duplication

### LINKED ISSUE #458: [bug] Discord bridge duplicates messages — AC→Discord forwards same messages multiple times
## Problem

The Discord bridge forwards the same AC messages to Discord multiple times, producing visible duplication in the Discord channel:

- Head's batch queue message appeared **4 times**
- Operator's \"Start the batch 17 now\" message appeared **12+ times**

Only one message was sent from each source. The bridge is multiplying them on the AC→Discord forwarding path.

### H1 — Slug mismatch breaks echo prevention

The bridge config has \`bridge_sender = "dc"\` but AC registered the bridge as \`discord-bridge\` (because AC config still has \`[agents.discord-bridge]\` not \`[agents.dc]\` — see #457). When the bridge sends a message to AC with \`sender: "dc"\`, AC may store it under the registered name \`discord-bridge\` or the submitted name \`dc\`. If the stored sender doesn't match either \`ac["known_names"]\` or \`bridge_sender\`, echo prevention fails and the bridge re-forwards its own messages back to Discord in a feedback loop.

**Check:** query AC's message store directly and inspect the \`sender\` field on messages the bridge forwarded. Is it \`"dc"\`, \`"discord-bridge"\`, or something else?

### H2 — Pagination drain loop re-forwards

PR #429 (#425) added an inner \`while True\` loop that drains pages of 50 until a partial page is returned. If AC's \`/api/messages?since_id=N\` endpoint returns the same messages on repeated calls (e.g. \`since_id\` isn't being honored, or message IDs are non-monotonic after AC restart), the loop keeps forwarding the same batch and never breaks.

**Check:** add logging inside the drain loop to see how many iterations it runs per poll cycle. If > 1 for small message counts, the drain logic is re-fetching already-seen messages.

### H3 — Cursor reset after bridge crash/restart

If the bridge process restarted (crash, AC restart, manual stop/start), the in-memory cursor resets to whatever was persisted on disk. If the disk cursor was stale (behind the current AC message count), the bridge re-forwards all messages from the stale cursor to current on the next poll.

**Check:** compare the cursor file value to the latest AC message ID. If cursor < latest, this is the replay window.

### H4 — Multiple bridge instances

If orphan bridge processes from previous restarts are still running alongside the current one, each instance independently polls and forwards — producing N copies where N = number of running instances.

**Check:** \`ps aux | grep discord_bridge\` — should be exactly 1 process per project.

## Acceptance

- One AC message = one Discord message (no duplication)
- Echo prevention works regardless of slug mismatch (guard against both registered name and configured bridge_sender)
- Pagination drain loop is bounded and doesn't re-forward on same-page responses
- Bridge restart doesn't replay already-forwarded messages
- Add a dedup guard: track the last N forwarded message IDs in memory, skip any that were already sent to Discord in the current session

## Files

- \`bridges/discord/discord_bridge.py\` — AC poll loop, echo prevention, cursor management, pagination drain

## Branch

\`task/XXX-discord-bridge-dedup\`

## Related

- quadwork#457 — bridge slug migration (slug mismatch is a contributing factor)
- quadwork#425 (PR #429) — pagination drain loop (may be amplifying the duplication)
- quadwork#424 (PR #428) — echo prevention known_names tracking

realproject7:
## Summary
- Add in-memory dedup guard tracking recently forwarded message IDs (capped at 2000) to prevent duplicate Discord messages from stale cursors, drain-loop issues, or restart replays
- Bound the pagination drain loop at 20 iterations with stale batch detection (break if all IDs <= cursor)
- Harden echo prevention with "discord-bridge" and "discord_bridge" slug variants to survive AC config slug mismatches
- Simplify cursor advancement: always advance on every message regardless of skip/forward

Fixes #458

## Test plan
- [ ] One AC message = one Discord message (no duplication)
- [ ] Echo prevention works with both `dc` and `discord-bridge` sender names
- [ ] Bridge restart doesn't replay already-forwarded messages within the same session
- [ ] Drain loop terminates properly under normal conditions (< 50 messages)
- [ ] Drain loop breaks if AC returns stale responses

### DIFF
--- a/bridges/discord/discord_bridge.py
+++ b/bridges/discord/discord_bridge.py
@@ -277,41 +277,46 @@ async def poll_ac_to_discord(cfg, channel):
                     log.warning("AC returned stale batch (max_id=%d <= cursor=%d) — breaking drain", max_batch_id, _cursor["last_seen_id"])
                     break
 
+                # #458: build echo names once per batch (inputs don't
+                # change per-message).
+                echo_names = ac["known_names"] | {
+                    bridge_sender,
+                    "discord-bridge",
+                    "discord_bridge",
+                }
+
                 for msg in messages:
                     msg_id = msg.get("id", 0)
                     sender = msg.get("sender", "")
                     text = msg.get("text", "")
 
-                    # Always advance cursor and persist immediately so a
-                    # crash between messages doesn't replay from a stale
-                    # position (#458).
-                    if msg_id > _cursor["last_seen_id"]:
-                        _cursor["last_seen_id"] = msg_id
-                        save_cursor(cfg["cursor_file"])
+                    # Helper: advance cursor and persist. Called after
+                    # a message is fully handled (skipped or forwarded)
+                    # so a crash can't replay it (#458). NOT called
+                    # before Discord delivery to avoid silent message
+                    # loss on transient send failures.
+                    def commit_cursor():
+                        if msg_id > _cursor["last_seen_id"]:
+                            _cursor["last_seen_id"] = msg_id
+                            save_cursor(cfg["cursor_file"])
 
-                    # Echo prevention: skip our own messages (any name
-                    # the bridge has ever registered as this session,
-                    # plus the configured bridge_sender and common
-                    # slug variants to survive slug mismatches)
-                    # #458: also match "discord-bridge" (old slug) and
-                    # any dash/underscore variant of bridge_sender
-                    echo_names = ac["known_names"] | {
-                        bridge_sender,
-                        "discord-bridge",
-                        "discord_bridge",
-                    }
+                    # Echo prevention: skip our own messages
                     if sender in echo_names:
+                        commit_cursor()
                         continue
 
                     # Skip system auto-recovery messages
                     if sender == "system":
+                        commit_cursor()
                         continue
 
                     if not text:
+                        commit_cursor()
                         continue
 
                     # #458: dedup guard — skip already-forwarded messages
                     if msg_id in forwarded_ids:
+                        commit_cursor()
                         continue
 
                     # Forward to Discord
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The drain-loop and slug-mismatch defenses look reasonable, but the PR still leaves a restart replay window that the issue explicitly calls out.

### Findings
- [high] Restart/crash during a page can still replay already-forwarded messages because dedup state is in-memory only and the cursor is persisted only after the page finishes.
  - File: `bridges/discord/discord_bridge.py:225`, `bridges/discord/discord_bridge.py:321`, `bridges/discord/discord_bridge.py:331`
  - Suggestion: persist dedup/cursor progress at message granularity after a successful Discord send, or otherwise make restart recovery durable. Right now `forwarded_ids` is recreated on every process start, and `save_cursor()` still runs only after the whole page loop. If the bridge forwards message IDs 101-120, then crashes before `save_cursor()`, the next process starts with an empty `forwarded_ids` set plus the old cursor and re-sends that slice. That misses the acceptance requirement that restart should not replay duplicates.

### Decision
Requesting changes because the PR does not fully close the restart replay path described in #458.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new commit fixes the stale-cursor restart window, but it now acknowledges messages before they are successfully delivered to Discord.

### Findings
- [high] Cursor persistence now happens before the Discord send, so transient send failures will permanently drop AC messages.
  - File: `bridges/discord/discord_bridge.py:285-289`, `bridges/discord/discord_bridge.py:315-330`
  - Suggestion: only durably advance the cursor after a successful Discord send for forwardable messages, or persist a separate durable dedup/ack state that still allows retry on send failure. In the current flow, message `id=123` is written to `_cursor` and saved immediately, then `await channel.send(...)` may fail. On the next poll the bridge asks AC for `since_id=123`, so that unsent message is skipped forever. That trades duplication for silent data loss.

### Decision
Requesting changes because losing operator/agent messages on transient Discord errors is a correctness regression and does not satisfy the intended one-message-to-one-forward behavior.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commits close both of the blockers from review: restart replay is prevented by per-message cursor persistence, and forwarded messages now commit only after successful Discord delivery.

### Findings
- None.

### Decision
Approving because the dedup, drain-loop, echo-prevention, and cursor-handling changes now address the duplication paths in #458 without introducing replay or silent-drop regressions.

PR #485, opened 2026-04-12
[#478] Fix ghost slot accumulation on agent restart

### LINKED ISSUE #478: [bug] Agent slot ghosts accumulate on every restart — dev-1/dev-4/dev-9 instead of clean dev
## Problem

After multiple server restarts or AC restart cycles, agent registrations accumulate suffixed ghost slots. A new project's Head agent sees:

\`\`\`
dev-1, dev-4, dev-9, re1-1, re1-2, re2-1, re2-2
\`\`\`

instead of clean \`dev\`, \`re1\`, \`re2\`. Each restart creates a new slot because the deregister step fails (stale token from the pre-restart AC instance is rejected). The old slots persist as ghosts until AC's 60s timeout — but if restarts happen faster than 60s, ghosts accumulate indefinitely.

This causes:
- **Agent identity confusion** — Head assigns to \`@dev\` but the live agent is registered as \`dev-9\`, ignores the mention
- **Cluttered roster** — agents see 7+ online entries instead of 4
- **Chat spam** — disconnect/reconnect messages for every ghost

## Previous fixes (insufficient)

- **#388** (PR #392) — deregister on bridge stop. Only covers bridges, not agents.
- **#417** (PR #432) — Reset Agents full stop+respawn. Deregister still uses stale token → fails → suffix.
- **#447** (PR #449) — Chain agent reset into AC restart. Same stale-token deregister failure.

All three fixes attempt to deregister with the old token, which AC rejects because the token was issued by the previous AC instance. The fundamental problem: **you can't deregister a ghost slot without a valid token for that slot.**

### Option A (recommended) — AC-side: force-replace on localhost register

When a register request comes from \`127.0.0.1\` for a base that already has active slots, **immediately expire all existing slots for that base** and assign slot 1 to the new registration. Safe because localhost = same machine = the old slots are guaranteed to be ghosts from crashed/restarted processes.

This requires a change to the AgentChattr registry (\`~/.quadwork/<project>/agentchattr/registry.py\`) or a configuration option in \`config.toml\`.

### Option B — QuadWork-side: wait for ghost timeout before registering

After AC restart, wait 65s (ghost timeout + margin) before registering agents. Guarantees all old slots have expired. Downside: 65s delay on every restart before agents are usable.

### Option C — QuadWork-side: name-based deregister endpoint

Add an AC endpoint that accepts deregister by agent base name from localhost (no token needed). QuadWork calls this before registering. Requires an AC change but simpler than Option A.

## Acceptance

- Server restart → all 4 agents register as \`head\`, \`dev\`, \`re1\`, \`re2\` (slot 1, no suffix)
- No ghost slots visible in the AC roster after restart
- Head sees exactly 4 agents when checking the roster
- \`@dev\` mentions are picked up by the live agent immediately

## Files

- AgentChattr registry: \`~/.quadwork/<project>/agentchattr/registry.py\` — force-replace logic (Option A)
- Or: \`server/index.js\` — wait/retry logic (Option B), or new deregister endpoint (Option C)

## Branch

\`task/XXX-clean-agent-slots-on-restart\`

## Related

- quadwork#417 (PR #432) — first attempt (closed, insufficient)
- quadwork#447 (PR #449) — second attempt (insufficient)
- quadwork#388 (PR #392) — bridge deregister (different scope)

realproject7:
## Summary

- **Bug**: Agent slots accumulate ghost suffixes (`dev-1`, `dev-4`, `dev-9`) after restarts because deregister fails with stale tokens from pre-restart AC instances
- **Fix**: Implements Option A from the issue — force-replace on register:
  1. `registerAgent()` now accepts `{ force: true }` which tells AC to expire all existing slots for that base before creating a new one
  2. All register call sites (initial launch + 409 recovery) pass `force: true`
  3. Startup auto-patches deployed AC instances (`registry.py` + `app.py`) to support the `force` parameter, since we don't have push access to upstream agentchattr

Fixes #478

## Test plan

- [ ] Restart QuadWork server — verify `[ghost-fix]` log lines for patched AC instances
- [ ] Start all 4 agents — verify they register as clean `head`, `dev`, `re1`, `re2` (no suffixes)
- [ ] Restart agents multiple times in quick succession — verify no ghost accumulation
- [ ] Check AC roster (`chat_who`) — exactly 4 agents, no `-N` suffixes
- [ ] Verify `@dev` mentions are picked up by the live agent immediately

### DIFF
--- a/server/agentchattr-registry.js
+++ b/server/agentchattr-registry.js
@@ -43,7 +43,7 @@ async function waitForAgentChattrReady(serverPort, timeoutMs = 10000) {
  * Register an agent with AgentChattr. Returns {name, token, slot} on
  * success, null on failure (with registerAgent.lastError populated).
  */
-async function registerAgent(serverPort, base, label = null) {
+async function registerAgent(serverPort, base, label = null, { force = false } = {}) {
   registerAgent.lastError = null;
   try {
     const r = await fetchWithTimeout(
@@ -51,7 +51,7 @@ async function registerAgent(serverPort, base, label = null) {
       {
         method: "POST",
         headers: { "Content-Type": "application/json" },
-        body: JSON.stringify({ base, label }),
+        body: JSON.stringify({ base, label, force }),
       },
     );
     if (!r.ok) {
--- a/server/index.js
+++ b/server/index.js
@@ -369,7 +369,8 @@ async function buildAgentArgs(projectId, agentId) {
         await deregisterAgent(acServerPort, agentId, stalePersistedToken).catch(() => {});
         clearPersistedAgentToken(projectId, agentId);
       }
-      const registration = await registerAgent(acServerPort, agentId, agentCfg.display_name || null);
+      // #478: force-replace so AC expires any ghost slots for this base
+      const registration = await registerAgent(acServerPort, agentId, agentCfg.display_name || null, { force: true });
       if (!registration) {
         throw new Error(`Failed to register ${agentId}: ${registerAgent.lastError}`);
       }
@@ -394,7 +395,8 @@ async function buildAgentArgs(projectId, agentId) {
         await deregisterAgent(acServerPort, agentId, stalePersistedToken).catch(() => {});
         clearPersistedAgentToken(projectId, agentId);
       }
-      const registration = await registerAgent(acServerPort, agentId, agentCfg.display_name || null);
+      // #478: force-replace so AC expires any ghost slots for this base
+      const registration = await registerAgent(acServerPort, agentId, agentCfg.display_name || null, { force: true });
       if (!registration) {
         throw new Error(`Failed to register ${agentId}: ${registerAgent.lastError}`);
       }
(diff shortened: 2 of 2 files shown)


realproject7 (commented):
**RE2 verdict: APPROVED**

**Part 1 (registry.js + call sites):** Clean. `{ force }` param added correctly, all 3 register call sites updated. No issues.

**Part 2 (startup monkey-patching of AC Python):** The force-replace logic is correct — ghosts are moved to `_reserved` (reclaimable, not hard-deleted). Patches are idempotent and failures are caught gracefully.

**One note for awareness** (not blocking): The patches rely on exact string matching against AC source (`def register(self, base: str, label: str | None = None)`, exact comment/whitespace patterns). If upstream AC changes its source layout, the patches will silently no-op and ghosts will return. Consider logging a warning when the `replace()` produces no change — e.g., `if (reg === before) console.warn(...)` — so the failure is visible. But this is a hardening suggestion, not a blocker.

Overall the approach matches the issue's recommended Option A and is the right pragmatic call given no upstream push access.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The fix addresses the actual ghost-slot failure mode by making QuadWork’s canonical agent registrations force-replace stale family entries on loopback-only AgentChattr instances. The scope is limited to the project-agent startup/recovery paths, so normal multi-slot behavior remains opt-in rather than globally removed.

### Findings
- No blocking findings.
- Residual risk: GitHub reports no PR checks on this branch, so verification is manual.

### Decision
Approve because the current diff fixes issue #478’s restart/ghost accumulation path without changing general registry semantics for clients that do not pass `force: true`.

PR #489, opened 2026-04-13
[#487] Extend chat token resync to recover from 5xx/502

### LINKED ISSUE #487: [bug] Chat token sync doesn't trigger on 502 — only handles 401/403 (follow-up to #448)
## Problem

PR #450 (#448) added chat token re-sync on 401/403 responses. But when AC is temporarily unreachable (restart, crash recovery), the proxy returns **502** instead of 403. The 502 path doesn't trigger \`syncChattrToken()\`, so the operator sees \"Send failed: 502\" and the chat stays broken until a manual page refresh.

## Evidence

PlotLink project: loop guard paused agents at 14:22, system auto-resumed at 15:40. During the gap, the AC token became stale. Operator typed \`/continue\` → \"Send failed: 502\". The 502 comes from the proxy failing to connect to AC, not from AC rejecting the token.

## Fix

Extend the token sync trigger in \`server/routes.js\` GET \`/api/chat\` handler to also cover 502 and 5xx responses:

\`\`\`js
// Current: only syncs on 401/403
if (status === 401 || status === 403) {
  await syncChattrToken(projectId);
  // retry...
}

// Fix: also sync on 502/503/5xx
if (status === 401 || status === 403 || status >= 500) {
  await syncChattrToken(projectId);
  // retry...
}
\`\`\`

Same for the POST \`/api/chat\` handler.

## Acceptance

- AC temporarily down → chat send returns 502 → token re-synced → retry succeeds once AC is back
- No \"Send failed: 502\" visible to operator (auto-recovers within one retry)

## Files

- \`server/routes.js\` — GET and POST \`/api/chat\` error handling paths

## Branch

\`task/XXX-chat-token-sync-on-502\`

## Related

- quadwork#448 (PR #450) — original token sync fix (401/403 only)

realproject7:
## Summary
- **GET `/api/chat`**: extends the resync-and-retry condition from `401 || 403` to also include `>= 500`, covering 502 and other 5xx from temporary AC outages
- **POST `/api/chat`** (WebSocket path): previously only retried on ws close code 4003 (`EAGENTCHATTR_401`); now also attempts token resync on generic connection errors (ECONNREFUSED, ECONNRESET, timeout) so the operator auto-recovers once AC is back

## Test plan
- [x] Existing `routes.chatWsSend.test.js` passes (5 cases)
- [ ] Manual: stop AC, attempt send from dashboard (should get 502), restart AC → next send auto-recovers via resync
- [ ] Manual: verify 401/403 path still works as before (no regression)

Fixes #487

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -146,6 +146,20 @@ router.get("/api/chat", async (req, res) => {
     if (!r.ok) return res.status(r.status).json({ error: `AgentChattr returned ${r.status}` });
     res.json(await r.json());
   } catch (err) {
+    // #487: fetch threw (ECONNREFUSED, DNS failure, etc.) — AC is
+    // unreachable. Resync the token and retry once; AC may have
+    // restarted with a new session token by the time the retry fires.
+    if (projectId) {
+      try { await syncChattrToken(projectId); } catch {}
+      const { token: refreshed } = getChattrConfig(projectId);
+      if (refreshed && refreshed !== token) {
+        try {
+          const retry = await fetch(buildUrl(refreshed), { headers: chatAuthHeaders(refreshed) });
+          if (!retry.ok) return res.status(retry.status).json({ error: `AgentChattr returned ${retry.status}` });
+          return res.json(await retry.json());
+        } catch {}
+      }
+    }
     res.status(502).json({ error: "AgentChattr unreachable", detail: err.message });
   }
 });
--- a/server/routes.js
+++ b/server/routes.js
@@ -132,7 +132,9 @@ router.get("/api/chat", async (req, res) => {
     const r = await fetch(buildUrl(token), { headers: chatAuthHeaders(token) });
     // #448: on 401/403, re-sync the session token from AC and retry
     // once. The stored token may be stale after an AC restart.
-    if ((r.status === 401 || r.status === 403) && projectId) {
+    // #487: also retry on 5xx — a temporary AC outage (e.g. 502) may
+    // precede a restart that regenerates the session token.
+    if ((r.status === 401 || r.status === 403 || r.status >= 500) && projectId) {
       try { await syncChattrToken(projectId); } catch {}
       const { token: refreshed } = getChattrConfig(projectId);
       if (refreshed && refreshed !== token) {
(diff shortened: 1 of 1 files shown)


realproject7 (commented):
## RE2 Review — #489 (chat token resync on 502)

**Verdict: APPROVED**

### GET handler (line ~134)
Adding `|| r.status >= 500` to the retry condition is clean and correct. Matches the issue description exactly — 502 from a temporary AC outage now triggers token resync + one retry, same as 401/403. No concerns.

### POST handler (line ~984)
The `isConnError = !isAuthError && err` catch-all means *any* error (not just connection errors) now triggers a resync attempt. This is broader than the issue scope — it catches programming bugs, malformed payloads, etc. in addition to ECONNREFUSED/ECONNRESET/timeout.

That said, the impact is bounded: one extra `syncChattrToken` call + one retry attempt that will also fail for non-connection errors, then falls through to the existing 502 response. So it's harmless in practice, and arguably more resilient than enumerating specific error codes.

The status code fix (`isAuthError ? 401 : 502`) and the early-return guard on `isAuthError` are correct — connection errors that can't be recovered now properly fall through to the generic 502 path instead of returning a misleading 401.

### Tests
Existing 5 test cases pass. No new test cases for the 5xx/connection-error retry path, but the PR marks manual testing in the test plan, which is reasonable for this kind of network-error scenario.

### Acceptance criteria
✅ 5xx triggers token resync + retry on GET path
✅ Connection errors trigger token resync + retry on POST/ws path
✅ No regression on existing 401/403 handling

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The POST retry broadening is in the right area, but the GET `/api/chat` half of the fix still misses the actual 502 path described in #487. In this handler, an AC outage reaches the `catch` block as a thrown fetch error, so widening the `r.status` check to `>= 500` does not cover the reported failure mode.

### Findings
- [high] GET `/api/chat` still returns the visible 502 without any token re-sync when AC is temporarily unreachable.
  - File: `server/routes.js:148`
  - Suggestion: move the GET re-sync-and-retry logic into the fetch-error path as well, or otherwise handle thrown connection failures before returning `AgentChattr unreachable`. Right now only an upstream HTTP 5xx response retries; a refused/reset connection still falls straight through to the existing 502 response.

### Decision
Requesting changes because #487 explicitly requires the GET and POST `/api/chat` handlers to auto-recover from the observed 502/5xx outage case, and the GET handler still does not retry on the actual thrown-error path used by this server.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up addresses the missing GET failure mode I flagged. Both GET and POST `/api/chat` now attempt a single token re-sync and retry for the outage/restart paths described in #487.

### Findings
- No blocking findings on the updated diff.

### Decision
Approving. The GET handler now covers thrown connection failures in addition to upstream HTTP 5xx responses, which satisfies the previously missing half of the acceptance criteria.

PR #503, opened 2026-04-15
[#502] Fix idle agent -2 suffix accumulation from heartbeat timeout

### LINKED ISSUE #502: [bug] Idle agents accumulate -2 suffixes from heartbeat timeout cycles
## Problem

When agents are idle (PC on, no batch assigned), they cycle through heartbeat timeout → disconnect → reconnect → new registration with -2 suffix. Over hours, this produces:

\`\`\`
re1 disconnected (timeout)
re1-2 disconnected (timeout)
re1-2 disconnected (timeout)
re1-2 disconnected (timeout)
\`\`\`

The \`force: true\` fix from PR #485 (#478) handles restarts, but the **idle heartbeat timeout cycle** is a different path:
1. Agent's CLI session goes idle (no user prompt, no queue-watcher trigger)
2. The CLI's MCP connection to AC times out or the CLI session itself suspends
3. AC's 60s heartbeat timeout fires → marks agent as disconnected
4. Queue watcher's next tick wakes the CLI → CLI reconnects → re-registers
5. Old ghost slot from step 3 hasn't fully expired → new slot gets -2 suffix

## Root cause

The heartbeat from \`server/index.js\` sends periodic keepalive pings to AC. But if the agent CLI itself is idle (waiting for input), it may not be responding to AC's heartbeat checks. AC interprets silence as a crash and disconnects the agent.

### Option A — Increase AC heartbeat timeout for idle agents
Change AC's crash timeout from 60s to something much longer (e.g. 5 minutes) for agents that are in a known-idle state. QuadWork knows whether an agent is idle (no active batch, no queue-watcher triggers).

### Option B — QuadWork-side keepalive
QuadWork's server already runs a heartbeat loop. Ensure it fires reliably every 5s for ALL agents, even idle ones — the heartbeat should not be gated on CLI activity.

### Option C — Suppress disconnect/reconnect messages for known agents
If AC can't avoid the timeout, at least suppress the \`disconnected (timeout)\` → \`is online\` chat spam. AC could treat a quick disconnect → reconnect within 10s as a \"glitch\" and not broadcast either message.

## Acceptance

- Agent idle for 8+ hours (PC on, no batch) → stays registered as \`head\` / \`dev\` / \`re1\` / \`re2\` (no suffix)
- No \`disconnected (timeout)\` spam in the chat log for idle agents
- Actual crashes (agent CLI killed) still produce a disconnect message after a reasonable timeout

## Files

- \`server/index.js\` — heartbeat loop, agent keepalive
- \`server/agentchattr-registry.js\` — registration with force
- AC's crash timeout config (if configurable via \`config.toml\`)

## Branch

\`task/XXX-idle-agent-suffix-cycle\`

## Related

- quadwork#478 (PR #485) — force-replace on restart (different trigger path)
- The bridge noise ticket — the disconnect/reconnect spam is amplified by the bridge forwarding these messages

realproject7:
## Summary

- **force-replace now clears `_reserved`**: After crash-timeout deregisters an idle agent, its name moves to `_reserved` (grace period). The existing `force=true` logic only cleared `_instances`, so re-registration still found slot 1 blocked → `-2` suffix. Now `force=true` also clears `_reserved` entries for the base, ensuring slot 1 is always reclaimed.
- **Crash timeout increased from 15s to 120s**: The 15s timeout (with a misleading "60s" comment) was too aggressive — just 3 missed 5s heartbeats triggered a spurious disconnect. 120s tolerates brief OS sleep / process scheduling delays while still detecting actual crashes within 2 minutes.
- **Bootstrap patches updated**: Already-deployed AC instances get both fixes on next QW server start. Existing `#478` force-replace patches are upgraded in-place.

Also fixes the `~60s` comment in `agentchattr-registry.js` to match actual timeout.

## Test plan

- [ ] Start agents, leave idle for 8+ hours — verify no `-2` suffix or `disconnected (timeout)` spam
- [ ] Kill an agent CLI process — verify disconnect message appears within ~2 minutes
- [ ] Restart QW server — verify bootstrap patches apply (check `[ghost-fix]` and `[idle-fix]` console logs)
- [ ] With a previously patched AC instance, verify the `stale_reserved` upgrade path fires

Fixes #502

### DIFF
--- a/server/agentchattr-registry.js
+++ b/server/agentchattr-registry.js
@@ -101,7 +101,7 @@ async function deregisterAgent(serverPort, name, token) {
 /**
  * Start a per-agent heartbeat that POSTs /api/heartbeat/{name} every 5s
  * with bearer auth. AgentChattr considers an agent crashed and removes
- * it after ~60s without a heartbeat, so without this every registered
+ * it after ~120s without a heartbeat, so without this every registered
  * QuadWork agent silently disappears from the channel one minute after
  * registration.
  *
--- a/server/index.js
+++ b/server/index.js
@@ -2042,7 +2042,7 @@ server.listen(PORT, "127.0.0.1", () => {
   // #478 + #502: patch deployed AgentChattr instances to support force-replace
   // on register and fix idle-agent crash timeout.
   for (const p of (startupCfg.projects || [])) {
-    const acDir = p.agentchattr_dir || path.join(os.homedir(), ".quadwork", p.id, "agentchattr");
+    const acDir = resolveProjectChattr(p.id).dir;
     // Patch registry.py: add force parameter to register()
     const regPath = path.join(acDir, "registry.py");
     if (fs.existsSync(regPath)) {
--- a/server/index.js
+++ b/server/index.js
@@ -2039,8 +2039,8 @@ server.listen(PORT, "127.0.0.1", () => {
       }
     }
   }
-  // #478: patch deployed AgentChattr instances to support force-replace
-  // on register. Prevents ghost slot accumulation after restarts.
+  // #478 + #502: patch deployed AgentChattr instances to support force-replace
+  // on register and fix idle-agent crash timeout.
   for (const p of (startupCfg.projects || [])) {
     const acDir = p.agentchattr_dir || path.join(os.homedir(), ".quadwork", p.id, "agentchattr");
     // Patch registry.py: add force parameter to register()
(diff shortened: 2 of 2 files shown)


realproject7 (commented):
**RE2 verdict: APPROVED**

Both fixes are correct and well-scoped:

1. **`_reserved` clearing** — The root cause was that `force=true` moved ghosts to `_reserved` (grace period), which still blocked slot 1 on re-registration. New code skips the grace period entirely and also sweeps existing `_reserved` entries for the base. This is the right behavior for force-replace.

2. **Crash timeout 15s → 120s** — The old 15s timeout (3 missed heartbeats at 5s intervals) was too aggressive for idle agents. 120s tolerates idle/sleep without masking actual crashes. Comment in `agentchattr-registry.js` updated to match.

3. **Upgrade path** — The regex-based upgrade for already-patched AC instances correctly targets the old `self._reserved[name] = time.time()` pattern. Both fresh-install and upgrade scenarios are handled.

Acceptance criteria addressed:
- ✅ Idle agents keep their slot (longer timeout prevents spurious disconnect)
- ✅ No disconnect/reconnect spam (same mechanism)
- ✅ Actual crashes still detected within ~2 minutes

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The core fix direction looks right, but the startup patch targets the wrong AgentChattr directory for legacy/fallback installs. On those projects, the real running clone keeps the old 15s timeout and stale `_reserved` behavior, so the bug remains unfixed in production.

### Findings
- [high] Startup patching ignores the same legacy AgentChattr fallback used at runtime, so existing projects can keep running an unpatched clone.
  - File: `server/index.js:2045`
  - Suggestion: resolve the patch target through the same logic used to launch/connect to AgentChattr (`resolveProjectChattr` / legacy-global fallback) instead of hard-coding `p.agentchattr_dir || ~/.quadwork/<id>/agentchattr`.

### Decision
Requesting changes because this is supposed to backfill already-deployed AgentChattr instances, but on legacy/fallback installs the patch never touches the clone QuadWork actually uses.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up fixes the deployment-path bug from my previous review. Bootstrap patching now resolves the same AgentChattr directory QuadWork actually uses, so legacy/global-clone installs receive the `_reserved` cleanup and 120s timeout backfill.

### Findings
- None.

### Decision
Approve. The implementation now matches the runtime AgentChattr resolution path and addresses the idle `-2` suffix accumulation issue without the earlier rollout gap.

PR #504, opened 2026-04-15
[#501] Filter system/status noise from bridge forwarding

### LINKED ISSUE #501: [ux] Filter system/status messages from bridge forwarding — too noisy
## Problem

The Discord/Telegram bridge forwards **every** AC message including:
- \`head is online\` (repeated on every heartbeat cycle or reconnect)
- \`re1 disconnected (timeout)\`
- \`re1-2 disconnected (timeout)\`
- System status messages

These are AC housekeeping messages, not operator-relevant content. They flood the Discord/Telegram channel with noise, making it impossible to find actual agent work messages.

## Fix

Add a message filter in the bridge's AC→Discord/Telegram forwarding path that skips:

1. **\"X is online\" messages** — match pattern: \`/^.+ is online$/\`
2. **\"X disconnected (timeout)\" messages** — match pattern: \`/disconnected \\(timeout\\)/\`
3. **System messages** — sender = \`system\` (already partially filtered, verify completeness)
4. **Duplicate consecutive messages** — same sender + same text within 60s (e.g. \`head is online\` repeated 5 times)

### Implementation

\`\`\`python
SKIP_PATTERNS = [
    re.compile(r"^.+ is online$"),
    re.compile(r"disconnected \(timeout\)"),
    re.compile(r"^.+ disconnected$"),
    re.compile(r"auto-recovered"),
    re.compile(r"Resuming agent conversation"),
]

def should_forward(msg):
    text = msg.get("text", "")
    sender = msg.get("sender", "")
    if sender == "system":
        return False
    for pattern in SKIP_PATTERNS:
        if pattern.search(text):
            return False
    return True
\`\`\`

### Dedup guard
Track last forwarded (sender, text, timestamp). If same sender + same text within 60s, skip. Prevents the \`head is online\` × 8 pattern.

## Acceptance

- \`head is online\` messages NOT forwarded to Discord/Telegram
- \`re1 disconnected (timeout)\` NOT forwarded
- System auto-recovery messages NOT forwarded
- Actual agent work messages (PR reviews, assignments, merge notifications) still forwarded
- Duplicate consecutive messages suppressed

## Files

- \`bridges/discord/discord_bridge.py\` — add filter before \`channel.send()\`
- \`~/.quadwork/agentchattr-telegram/telegram_bridge.py\` — same filter

## Branch

\`task/XXX-bridge-filter-noise\`

realproject7:
## Summary

- **Discord bridge**: Added `_should_forward()` filter that skips join/leave messages, system sender, pattern-matched noise (online, disconnected, auto-recovered, Resuming agent conversation), and deduplicates identical messages within 60s
- **Telegram bridge**: Same filter added to source repo, pin bumped to `09bb557`
- Existing deployed Telegram bridges will pick up the filter on next re-install via the updated pin

## Test plan

- [ ] Verify `head is online` messages NOT forwarded to Discord/Telegram
- [ ] Verify `re1 disconnected (timeout)` NOT forwarded
- [ ] Verify system auto-recovery messages NOT forwarded
- [ ] Verify actual agent work messages (PR reviews, assignments, merge notifications) still forwarded
- [ ] Verify duplicate consecutive messages (same sender+text within 60s) suppressed
- [ ] Re-install Telegram bridge and verify new pin applies

Fixes #501

### DIFF
--- a/bridges/discord/discord_bridge.py
+++ b/bridges/discord/discord_bridge.py
@@ -18,6 +18,7 @@ import atexit
 import json
 import logging
 import os
+import re
 import signal
 import sys
 import threading
@@ -213,6 +214,60 @@ def start_heartbeat(url):
 
 
 # ---------------------------------------------------------------------------
+# #501: message filter — suppress AC housekeeping noise from bridge output
+# ---------------------------------------------------------------------------
+
+_NOISE_PATTERNS = [
+    re.compile(r"^.+ is online$"),
+    re.compile(r"disconnected \(timeout\)"),
+    re.compile(r"^.+ disconnected$"),
+    re.compile(r"auto-recovered"),
+    re.compile(r"Resuming agent conversation"),
+]
+
+# Dedup guard: (sender, text) → timestamp of last forward
+_last_forwarded: dict[tuple[str, str], float] = {}
+_DEDUP_WINDOW = 60  # seconds
+
+
+def _should_forward(msg: dict) -> bool:
+    """Return True if this AC message should be forwarded to Discord/Telegram."""
+    sender = msg.get("sender", "")
+    text = msg.get("text", "")
+    msg_type = msg.get("type", "chat")
+
+    # Skip join/leave system messages (online, disconnected, timeout)
+    if msg_type in ("join", "leave"):
+        return False
+
+    # Skip system sender entirely
+    if sender == "system":
+        return False
+
+    # Pattern-based filter for edge cases
+    for pat in _NOISE_PATTERNS:
+        if pat.search(text):
+            return False
+
+    # Dedup: suppress identical (sender, text) within window
+    key = (sender, text)
+    now = time.time()
+    last = _last_forwarded.get(key)
+    if last is not None and now - last < _DEDUP_WINDOW:
+        return False
+    _last_forwarded[key] = now
+
+    # Prune stale dedup entries periodically
+    if len(_last_forwarded) > 500:
+        cutoff = now - _DEDUP_WINDOW
+        stale = [k for k, t in _last_forwarded.items() if t < cutoff]
+        for k in stale:
+            del _last_forwarded[k]
+
+    return True
+
+
+# ---------------------------------------------------------------------------
 # AC → Discord polling
 # ---------------------------------------------------------------------------
 
@@ -305,12 +360,12 @@ async def poll_ac_to_discord(cfg, channel):
                         commit_cursor()
                         continue
 
-                    # Skip system auto-recovery messages
-                    if sender == "system":
+                    if not text:
                         commit_cursor()
                         continue
 
-                    if not text:
+                    # #501: skip system/status noise and dedup
+                    if not _should_forward(msg):
                         commit_cursor()
                         continue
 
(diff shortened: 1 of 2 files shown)


realproject7 (commented):
**RE2 verdict: APPROVED**

Clean implementation of the noise filter for bridge forwarding.

**What works well:**
- Triple-layer filtering (message type → sender → regex patterns) catches all the noise categories from the issue
- Dedup guard with 60s window and pruning at 500 entries is appropriate for a long-running bridge process
- The old inline `if sender == "system"` check is cleanly subsumed by `_should_forward()`
- Telegram bridge handled via pin bump to external repo — correct approach

**Acceptance criteria check:**
- ✅ `head is online` — caught by both `msg_type == "join"` and `_NOISE_PATTERNS[0]`
- ✅ `re1 disconnected (timeout)` — caught by both `msg_type == "leave"` and `_NOISE_PATTERNS[1]`
- ✅ System/auto-recovery messages — caught by `sender == "system"` and pattern match
- ✅ Actual work messages (PR reviews, assignments, merges) — no patterns match these
- ✅ Consecutive duplicates suppressed within 60s window

No issues found.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The filter direction is reasonable, but the new dedupe logic can now lose legitimate bridge messages after a transient Discord send failure.

### Findings
- [high] `_should_forward()` records `(sender, text)` as forwarded before `channel.send()` succeeds, so a transient send error causes the next poll to suppress the same still-uncommitted message and then advance the cursor, silently dropping it.
  - File: `bridges/discord/discord_bridge.py:248`
  - Suggestion: only update the dedupe cache after a successful Discord delivery, alongside the existing `forwarded_ids`/`commit_cursor()` success path.

### Decision
Requesting changes because this introduces silent message loss on transient Discord send failures, which is worse than the original bridge noise.

realproject7 (commented):
**RE2 update: agreeing with RE1's concern — revising to REQUEST CHANGES**

RE1 is correct. I missed this on my first pass: `_last_forwarded[key] = now` is recorded inside `_should_forward()` *before* `channel.send()` executes. If Discord delivery fails transiently, the polling loop retries the same AC message but `_should_forward()` suppresses it as a duplicate, and `commit_cursor()` advances past it. This silently drops a legitimate message.

The fix is straightforward: split dedup into check and commit phases — `_should_forward()` should only check patterns/type/sender, and the dedup timestamp should be recorded after successful Discord delivery (alongside `forwarded_ids` / cursor commit).

Changing my verdict to **REQUEST CHANGES** pending this fix.

realproject7 (commented):
**RE2 re-review verdict: APPROVED**

The dedup concern is resolved. `_should_forward()` is now a pure check (no side effects), and `_mark_forwarded(msg)` is called only after successful Discord delivery at line 392, alongside `forwarded_ids.add()` and `commit_cursor()`. If `channel.send()` fails, neither the cursor nor the dedup state advances — the message will be retried on the next poll.

Telegram pin updated to include the same fix. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up fixes the message-loss regression from my previous review. `_should_forward()` is now side-effect free, and dedupe state is committed only after successful bridge delivery.

### Findings
- None.

### Decision
Approve. The filter now suppresses bridge noise without risking silent drops on transient Discord send failures.

PR #517, opened 2026-04-18
[#516] Server-side auto-stop for trigger and caffeinate on batch completion

### LINKED ISSUE #516: [bug] Auto-trigger doesn't stop when batch completes while operator is on a different project page
## Problem

The auto-trigger's batch lifecycle detection (from #408 / PR #412) is **entirely client-side** — it runs in \`ScheduledTriggerWidget.tsx\` and only polls \`/api/batch-progress\` when the operator is viewing that project's page. The scheduled trigger itself runs **server-side** (\`server/index.js\` \`setInterval\`).

When the operator is viewing a different project's page (or the home page) and a batch completes, the auto-stop never fires because:
1. The \`ScheduledTriggerWidget\` component is unmounted (not on the project page)
2. No client-side code is polling \`/api/batch-progress\` for that project
3. The server-side trigger keeps firing \`setInterval\` indefinitely

### Observed behavior

QuadWork Batch 27 completed at 13:55. The operator was on a different project page. At 14:09 (14 minutes later), the scheduled trigger was still firing queue-check messages into the completed batch — wasting agent tokens on a batch with no remaining work.

## Root cause

The auto-start/stop logic lives in the React component (\`ScheduledTriggerWidget.tsx:314-390\`), which only runs when mounted. The server has no knowledge of the \`trigger_auto\` setting — it just runs the timer it was told to start.

### Option A (recommended) — server-side auto-stop

Add a batch-progress check to the server's trigger tick handler. When the trigger fires, before sending the message, check if the batch is complete:

\`\`\`js
// In the trigger tick callback (server/index.js):
if (project.trigger_auto) {
  const batchData = await getBatchProgress(projectId);
  if (batchData && batchData.complete) {
    stopTrigger(projectId);
    console.log(\`[auto-trigger] \${projectId}: batch complete, auto-stopped\`);
    return; // don't send the message
  }
}
\`\`\`

This runs on every trigger tick (every 15 min by default), independent of which page the operator is viewing.

### Option B — server-side background poller

Add a separate 30s \`setInterval\` on the server that checks batch-progress for all projects with \`trigger_auto: true\` and stops the trigger when complete. More responsive than Option A (checks every 30s instead of every trigger interval).

### Keep the client-side logic too

Don't remove the client-side auto-start/stop in \`ScheduledTriggerWidget\` — it provides faster response when the operator IS on the page. The server-side check is the safety net for when they're not.

## Also applies to Keep Mac Awake auto-toggle (#441)

The same issue affects the Keep Mac Awake auto-stop. If the operator isn't on the project page when the batch completes, caffeinate keeps running. The fix should cover both auto-triggers.

## Acceptance

- Batch completes while operator is on a DIFFERENT project page → trigger auto-stops within 30s
- Batch completes while operator is on the HOME page → same
- Batch completes while operator IS on the project page → auto-stops immediately (existing client-side behavior preserved)
- Keep Mac Awake auto-stop follows the same server-side check
- Log line: \`[auto-trigger] <project>: batch complete, auto-stopped\`

## Files

- \`server/index.js\` — trigger tick handler or new background poller
- \`server/routes.js\` — may need to expose \`getBatchProgress\` internally (it's currently HTTP-only)
- Keep \`src/components/ScheduledTriggerWidget.tsx\` as-is (client-side supplement)

## Branch

\`task/XXX-server-side-auto-trigger-stop\`

## Related

- quadwork#408 (PR #412) — original auto-trigger (client-side only)
- quadwork#462 (PR #463) — first-poll gap fix (also client-side)
- quadwork#441 (PR #454) — Keep Mac Awake auto-toggle (same issue)

realproject7:
## Summary
- Adds a batch-progress check to `sendTriggerMessage` — when `trigger_auto` is enabled, the server checks `/api/batch-progress` before each trigger tick
- If the batch is complete, auto-stops both the trigger and caffeinate server-side, independent of which page the operator is viewing
- Logs `[auto-trigger] <project>: batch complete, auto-stopped` on stop
- Client-side `ScheduledTriggerWidget` auto-stop preserved as a faster supplement when the page is active
- Batch-progress check failure is non-fatal — proceeds with the message if the check fails

Fixes #516

## Test plan
- [ ] Batch completes while operator is on a different project page → trigger auto-stops within one tick
- [ ] Batch completes while operator is on the home page → same
- [ ] Batch completes while operator IS on the project page → auto-stops immediately (existing client-side behavior)
- [ ] Caffeinate auto-stops when batch completes
- [ ] Log line appears: `[auto-trigger] <project>: batch complete, auto-stopped`
- [ ] Non-auto triggers (trigger_auto not set) are unaffected

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1318,11 +1318,13 @@ async function sendTriggerMessage(projectId) {
         if (bp && bp.complete) {
           console.log(`[auto-trigger] ${projectId}: batch complete, auto-stopped`);
           stopTrigger(projectId);
-          // Also stop caffeinate if running (#441 companion fix)
-          if (caffeinateProcess.process) {
+          // Also stop caffeinate if no other triggers remain running
+          // (#441 companion fix). caffeinateProcess is global (not
+          // project-scoped), so only kill it when all work is done.
+          if (caffeinateProcess.process && triggers.size === 0) {
             try { caffeinateProcess.process.kill("SIGTERM"); } catch {}
             caffeinateProcess = { process: null, pid: null, startedAt: null, duration: null };
-            console.log(`[auto-trigger] ${projectId}: caffeinate auto-stopped`);
+            console.log(`[auto-trigger] ${projectId}: caffeinate auto-stopped (no active triggers remain)`);
           }
           return;
         }
--- a/server/index.js
+++ b/server/index.js
@@ -1301,6 +1301,38 @@ ALL: Communicate via this chat by tagging agents. Your terminal is NOT visible.`
 async function sendTriggerMessage(projectId) {
   const cfg = readConfig();
   const project = cfg.projects && cfg.projects.find((p) => p.id === projectId);
+
+  // #516: server-side auto-stop — check batch progress before sending.
+  // When trigger_auto is enabled, skip the message and stop the trigger
+  // (plus caffeinate) if the batch is already complete. This covers the
+  // case where the operator is on a different page and the client-side
+  // ScheduledTriggerWidget is not mounted to detect completion.
+  if (project && project.trigger_auto) {
+    const qwPort = cfg.port || 8400;
+    try {
+      const bpRes = await fetch(
+        `http://127.0.0.1:${qwPort}/api/batch-progress?project=${encodeURIComponent(projectId)}`
+      );
+      if (bpRes.ok) {
+        const bp = await bpRes.json();
+        if (bp && bp.complete) {
+          console.log(`[auto-trigger] ${projectId}: batch complete, auto-stopped`);
+          stopTrigger(projectId);
+          // Also stop caffeinate if running (#441 companion fix)
+          if (caffeinateProcess.process) {
+            try { caffeinateProcess.process.kill("SIGTERM"); } catch {}
+            caffeinateProcess = { process: null, pid: null, startedAt: null, duration: null };
+            console.log(`[auto-trigger] ${projectId}: caffeinate auto-stopped`);
+          }
+          return;
+        }
+      }
+    } catch (err) {
+      // Non-fatal — if batch-progress fails, proceed with the message
+      console.error(`[auto-trigger] ${projectId}: batch-progress check failed:`, err.message);
+    }
+  }
+
   const message = (project && project.trigger_message) || DEFAULT_MESSAGE;
 
   // #401 / quadwork#277: route trigger sends through the local
--- a/server/index.js
+++ b/server/index.js
@@ -1755,6 +1755,44 @@ function syncTriggersFromConfig() {
   }
 }
 
+// #516: server-side batch-completion poller. Checks every 30s whether
+// any trigger_auto project's batch is complete, and auto-stops the
+// trigger (plus caffeinate when no triggers remain). This runs
+// independently of the trigger tick interval, so completion is
+// detected within 30s even if the operator is on a different page.
+
+const AUTO_STOP_POLL_INTERVAL_MS = 30_000;
+
+async function autoStopPollingTick() {
+  const cfg = readConfig();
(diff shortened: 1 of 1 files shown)


realproject7 (commented):
**RE2 review — APPROVE**

Clean server-side auto-stop. Key points checked:

- **Guard**: only runs when `trigger_auto` is set — non-auto triggers unaffected ✓
- **Self-call via loopback**: calling own `/api/batch-progress` is pragmatic — reuses existing batch logic without duplication. Async `fetch` won't block the event loop ✓
- **Caffeinate cleanup**: matches existing pattern (lines 124-127) exactly ✓
- **Non-fatal error handling**: if batch-progress check fails, logs and proceeds with normal trigger ✓
- **Early return**: stops trigger and returns before sending the message when batch is complete ✓

No concerns.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The server-side batch-completion check is the right direction, but the caffeinate auto-stop is not project-scoped. As written, any project whose batch completes will stop the one global caffeinate process, even if that sleep-prevention session was started for a different project.

### Findings
- [medium] `caffeinateProcess` is a single global process with no project ownership, but the new `sendTriggerMessage(projectId)` path kills it whenever that specific project's batch is complete. In a multi-project setup, project A completing its batch can shut off Keep Mac Awake that the operator started for project B, which is a behavioral regression and does not satisfy the issue's project-scoped intent.
  - File: `server/index.js:1321`
  - Suggestion: Only auto-stop caffeinate if you can prove it belongs to the same project (for example by storing a `projectId` on caffeinate start), or leave caffeinate untouched in this PR and fix project-scoped ownership first.

### Decision
Request changes. The trigger auto-stop is fine, but the global caffeinate shutdown needs project ownership to avoid stopping the wrong project's sleep-prevention session.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The caffeinate change avoids the cross-project kill I flagged earlier, but the main server-side auto-stop still only runs on the trigger's own interval. That does not meet issue #516's acceptance requirement to stop within 30 seconds when the operator is on a different page.

### Findings
- [medium] `sendTriggerMessage(projectId)` checks batch completion only when the scheduled trigger fires, and triggers are configured with `(interval || 30) * 60 * 1000`. In the default path that means the server can keep running for up to 30 minutes after a batch completes, which reproduces the core bug the ticket is trying to eliminate for off-page operators.
  - File: `server/index.js:1305`
  - Suggestion: Add a separate short-interval server poller for `trigger_auto` projects (as the issue's Option B suggests), or otherwise run the completion check on a cadence that actually satisfies the 30-second acceptance target.

### Decision
Request changes. The server-side safety net needs a faster check cadence; gating it on the normal trigger interval is still too slow.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commits resolve both review blockers. Batch completion is now checked by a dedicated 30s server-side poller for active `trigger_auto` projects, and the global caffeinate process is only auto-stopped when no triggers remain.

### Findings
- [info] Prior timing gap is resolved; off-page auto-stop no longer depends on the normal trigger interval.
  - File: `server/index.js:1758`
  - Suggestion: None.
- [info] Prior cross-project caffeinate issue is narrowed appropriately for the current global caffeinate design.
  - File: `server/index.js:1321`
  - Suggestion: None.
- [info] No checks were reported on the branch at review time.
  - File: `task/516-server-side-auto-trigger-stop`
  - Suggestion: None for this PR; this is a process gap rather than a code defect.

### Decision
Approve. The server-side safety net now meets issue #516's 30-second requirement and preserves the intended multi-project behavior.

PR #521, opened 2026-04-18
[#519] Toggle to hide system/status messages in chat panel

### LINKED ISSUE #519: [feat] Toggle to hide system/status messages in AC primary chat panel
## Problem

The AgentChattr primary chat panel shows all messages including system noise:
- \"head disconnected (timeout)\"
- \"re1 is online\"
- \"Resuming agent conversation...\"
- \"Loop guard: 30 agent-to-agent hops reached\"

These clutter the chat view when the operator is trying to follow the actual agent work. The bridges already filter these (#501), but the dashboard chat panel doesn't.

## Proposal

Add a small **toggle button** in the chat panel header that filters system/status messages from the displayed message list.

### UI

A compact filter toggle next to the chat header:

\`\`\`
AGENTCHATTR — PRIMARY CHAT (?)    [👤 Agents only]
\`\`\`

- **ON (\"Agents only\"):** hide messages matching system patterns. Only show messages from agents (head, dev, re1, re2) and the operator (user).
- **OFF (\"All messages\"):** show everything including system messages (default for transparency)

### Filter logic

Reuse the same patterns from the bridge noise filter (#501 / PR #504):

\`\`\`js
const SYSTEM_PATTERNS = [
  /^.+ is online$/,
  /disconnected \(timeout\)/,
  /^.+ disconnected$/,
  /auto-recovered/,
  /Resuming agent conversation/,
  /Loop guard:/,
];

const shouldShow = (msg) => {
  if (!filterOn) return true;
  if (msg.sender === "system") return false;
  if (msg.type === "join" || msg.type === "leave") return false;
  for (const pat of SYSTEM_PATTERNS) {
    if (pat.test(msg.text)) return false;
  }
  return true;
};
\`\`\`

Apply \`.filter(shouldShow)\` on the messages array before rendering.

### Persistence

Store the toggle state in \`localStorage\` (per-browser, not per-project). Default: OFF (show all messages).

## Acceptance

- Toggle button visible in the chat panel header
- Toggle ON: system/status messages hidden from the chat view
- Toggle OFF: all messages visible (current behavior)
- Persists in localStorage across page loads
- No messages deleted — just hidden from view. Toggling back to OFF reveals them.
- Agent work messages (PR reviews, assignments, merge notifications) always visible regardless of toggle state

## Files

- \`src/components/ChatPanel.tsx\` — \`ChatPanelAPI\` component: add toggle state + filter before render

## Branch

\`task/XXX-chat-system-message-filter\`

realproject7:
## Summary

- Adds a compact **filter toggle** button in the chat panel header: "Agents only" (filter ON) / "All messages" (filter OFF)
- When ON, hides system noise: join/leave events, disconnect timeouts, loop guard messages, auto-recovery notices
- Reuses the same filter patterns from the bridge noise filter (#501 / PR #504)
- Persisted in `localStorage` (per-browser, not per-project), default OFF
- No messages deleted — toggling back reveals all messages

## Files changed

- `src/components/ChatPanel.tsx` — `isSystemMessage()` filter function, toggle state with localStorage, filter applied before render

## Test plan

- [ ] Toggle button visible in chat panel header
- [ ] Toggle ON: system/status messages (join, leave, disconnect, loop guard) hidden
- [ ] Toggle OFF: all messages visible (current behavior)
- [ ] Agent work messages (PR reviews, assignments, merge notifications) always visible
- [ ] Toggle state persists across page reloads (localStorage)
- [ ] Empty state still shows correctly when all visible messages are filtered

Fixes #519

### DIFF
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -29,6 +29,26 @@ const SENDER_COLORS: Record<string, string> = {
 
 const AGENTS = ["head", "re1", "re2", "dev", "user"];
 
+// #519: system/status message filter patterns — reuses the same set
+// from the bridge noise filter (#501 / PR #504).
+const SYSTEM_PATTERNS = [
+  /^.+ is online$/,
+  /disconnected \(timeout\)/,
+  /^.+ disconnected$/,
+  /auto-recovered/,
+  /Resuming agent conversation/,
+  /Loop guard:/,
+];
+
+function isSystemMessage(msg: Message): boolean {
+  if (msg.sender === "system") return true;
+  if (msg.type === "join" || msg.type === "leave") return true;
+  for (const pat of SYSTEM_PATTERNS) {
+    if (pat.test(msg.text)) return true;
+  }
+  return false;
+}
+
 // #410 / quadwork#276: slash commands recognized by AgentChattr's
 // native UI (mirrors agentchattr/static/chat.js lines 1978-1989).
 // Typing `/` in the chat input opens an autocomplete menu of these.
@@ -115,6 +135,19 @@ export default function ChatPanel({ projectId }: ChatPanelProps) {
 /** API-driven fallback when iframe is blocked */
 function ChatPanelAPI({ projectId }: { projectId?: string }) {
   const channel = "general";
+  // #519: filter toggle — hide system/status messages. Persisted in
+  // localStorage (per-browser, not per-project). Default: OFF.
+  const [filterSystem, setFilterSystem] = useState(() => {
+    if (typeof window === "undefined") return false;
+    return localStorage.getItem("chatFilterSystem") === "1";
+  });
+  const toggleFilter = useCallback(() => {
+    setFilterSystem((prev) => {
+      const next = !prev;
+      localStorage.setItem("chatFilterSystem", next ? "1" : "0");
+      return next;
+    });
+  }, []);
   const [messages, setMessages] = useState<Message[]>([]);
   // #410: track whether the initial fetch has completed so we don't
   // flash the welcome/empty state while messages are still loading.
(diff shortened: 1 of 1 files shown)


realproject7 (commented):
**RE2 Review — APPROVE**

Small, focused change. Implementation is clean.

**What's good:**
- Reuses existing `SYSTEM_PATTERNS` from the bridge noise filter (#501) — no new pattern invention
- `isSystemMessage()` is thorough: checks `sender`, `type`, and text patterns
- localStorage persistence is simple and appropriate (browser-level, not project-level)
- Toggle button styling matches the Auto toggle pattern from other widgets
- `displayMessages` applied correctly to both the message list and the empty-state check

**Minor observation (non-blocking):**
- When filter is ON and all visible messages happen to be system messages, the empty state shows `ProjectChatEmptyState` (welcome chips + How to Work modal) — which is a first-time-user experience, not a "no matches" state. In practice this scenario is rare (system messages almost always accompany agent messages), so not worth adding a separate filtered-empty state.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The filter implementation itself is fine, but it only exists in `ChatPanelAPI`. The default chat path still renders the embedded AgentChattr iframe, so the new toggle is absent in the primary UI whenever the iframe loads successfully.

### Findings
- [high] The hide-system-messages toggle is not available in the normal iframe chat path, so issue #519 is not actually delivered for the primary chat panel.
  - File: `src/components/ChatPanel.tsx:69`
  - Suggestion: either force the primary panel onto the API-rendered path when this feature is needed, or add equivalent filtering/toggle support to the iframe-backed experience. As written, only the fallback UI gets the new control.

### Decision
Requesting changes because the acceptance says the toggle should be visible in the chat panel header, but this PR adds it only to the API fallback component and not to the main iframe path.

project7-interns (approved):
## Verdict: APPROVE

### Summary
I rechecked the actual PR branch. `ChatPanel` now renders `ChatPanelAPI` directly after #415, so the filter toggle added here is on the primary chat path, not just a fallback. The feature matches issue #519 as implemented.

### Findings
- No blocking issues found.

### Decision
Approving because the toggle is visible in the primary chat panel, hides only the intended system/status noise, preserves all messages in memory, and persists via `localStorage` as requested.

PR #552, opened 2026-04-20
[#542] Guard bridge auto-stop to fire only on transition

### LINKED ISSUE #542: [bug] Auto-stop poller repeatedly kills bridges on every 30s tick
## Problem

When a batch completes, the `autoStopPollingTick` (every 30s) stops Telegram and Discord bridges on **every poll cycle** as long as `bp.complete` is true. This causes:

1. Repeated `[auto-bridge] ... auto-stopped` log spam
2. If the operator manually restarts bridges after batch completion, they get killed again 30s later
3. Bridges stay down until a new batch is queued (which triggers auto-start)

## Root Cause

`server/index.js` — the auto-stop path at line 1876-1879 has no transition guard:

```js
if (bp && bp.complete) {
  // ...trigger stop (has guard via triggers.has())...
  if (hasBridgeAuto) {
    await autoStopBridges(project.id, project, qwPort);  // fires EVERY tick
  }
}
```

The auto-**start** path (line 1883-1886) correctly uses `_bridgeBatchPrev` to detect a state transition:

```js
const isNewBatch = !prev || prev.complete || !prev.hasItems;
if (isNewBatch) { await autoStartBridges(...); }
```

But the auto-**stop** path doesn't check `prev` — it fires unconditionally whenever `bp.complete` is true.

## Fix

Guard the auto-stop the same way auto-start is guarded — only stop on the **transition** from incomplete to complete:

```js
if (bp && bp.complete) {
  const wasComplete = prev && prev.complete;
  // ...existing trigger stop...
  if (hasBridgeAuto && !wasComplete) {
    await autoStopBridges(project.id, project, qwPort);
  }
}
```

Same fix needed in `sendTriggerMessage()` at line 1406 — that path also calls `autoStopBridges` but is less problematic since the trigger itself gets stopped first (so it won't re-fire).

## Acceptance Criteria

- [ ] Bridge auto-stop fires only once per batch completion transition, not on every poll tick
- [ ] Manually restarted bridges are not killed by the poller
- [ ] Auto-start on new batch still works correctly
- [ ] No repeated `auto-stopped` log lines for the same batch completion

realproject7:
## Summary
- Adds a transition guard (`!prev || !prev.complete`) to the bridge auto-stop path in `autoStopPollingTick`, so `autoStopBridges` fires only once when a batch transitions from incomplete → complete, not on every 30s poll tick
- The `sendTriggerMessage` path (line 1405) is already implicitly guarded because `stopTrigger()` prevents re-entry
- 9-test suite covering code-structure verification and guard-logic simulation

Fixes #542

## Test plan
- [x] 9/9 tests pass (`npx jest server/__tests__/bridge-auto-stop-guard.test.js`)
- [ ] Verify no repeated `[auto-bridge] ... auto-stopped` log lines after batch completion
- [ ] Verify manually restarted bridges survive subsequent poll ticks
- [ ] Verify auto-start on new batch still works

### DIFF
--- a/server/__tests__/bridge-auto-stop-guard.test.js
+++ b/server/__tests__/bridge-auto-stop-guard.test.js
@@ -54,6 +54,31 @@ describe("#542 bridge auto-stop transition guard (code analysis)", () => {
     expect(setIdx).toBeGreaterThan(-1);
     expect(setIdx).toBeLessThan(stopIdx);
   });
+
+  test("autoStopBridges in sendTriggerMessage is guarded by prev.complete check", () => {
+    const fnStart = src.indexOf("async function sendTriggerMessage(");
+    expect(fnStart).toBeGreaterThan(-1);
+
+    // Extract until the next top-level function
+    const fnBody = src.slice(fnStart, fnStart + 2000);
+
+    const stopIdx = fnBody.indexOf("autoStopBridges(");
+    expect(stopIdx).toBeGreaterThan(-1);
+
+    // The guard should appear before the call
+    const guardRegion = fnBody.slice(Math.max(0, stopIdx - 300), stopIdx);
+    expect(guardRegion).toMatch(/!prev\.complete/);
+  });
+
+  test("sendTriggerMessage updates _bridgeBatchPrev before the guard", () => {
+    const fnStart = src.indexOf("async function sendTriggerMessage(");
+    const fnBody = src.slice(fnStart, fnStart + 2000);
+
+    const setIdx = fnBody.indexOf("_bridgeBatchPrev.set(");
+    const stopIdx = fnBody.indexOf("autoStopBridges(");
+    expect(setIdx).toBeGreaterThan(-1);
+    expect(setIdx).toBeLessThan(stopIdx);
+  });
 });
 
 // ---------------------------------------------------------------------------
--- a/server/__tests__/bridge-auto-stop-guard.test.js
+++ b/server/__tests__/bridge-auto-stop-guard.test.js
@@ -1,0 +1,109 @@
+/**
+ * #542 — Verify bridge auto-stop fires only on the transition from
+ * incomplete → complete, not on every polling tick.
+ *
+ * Since autoStopPollingTick is tightly coupled to HTTP calls, we test
+ * the guard logic by reading server/index.js and verifying the code
+ * structure, then simulating the guard in isolation.
+ */
+
+const fs = require("fs");
+const path = require("path");
+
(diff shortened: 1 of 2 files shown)


realproject7 (commented):
**RE2 review — APPROVE**

**Production fix (server/index.js:1890)**
The `(!prev || !prev.complete)` guard correctly mirrors the existing auto-start transition pattern at line 1897. The state flow is sound: `prev` is read from `_bridgeBatchPrev` at line 1875, the map is updated at 1876 (before the guard), so subsequent ticks see `prev.complete = true` and skip the stop call.

Edge cases handled:
- First tick after server restart with already-complete batch: `!prev` fires once — acceptable one-shot behavior
- Manual bridge restart after completion: survives because `prev.complete` is already true on next tick
- New batch cycle: `prev` resets via the `_bridgeBatchPrev.set` with `complete: false` during the in-progress phase, so the next completion correctly triggers stop again

**`sendTriggerMessage` path (line 1405)**: Agree with the PR assessment — `stopTrigger()` at line 1395 prevents re-entry, and the early `return` at 1406 makes this inherently one-shot. No guard needed.

**Tests**: The code-structure assertions (reading source and pattern-matching) are inherently fragile to refactors, but the logic simulation tests at lines 66-109 properly verify all transition scenarios. Acceptable trade-off for a guard this simple.

No issues found.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The poller-side guard is correct, but the PR does not fully implement issue #542. A second bridge auto-stop path in `sendTriggerMessage()` still fires unconditionally on completed batches.

### Findings
- [high] `sendTriggerMessage()` still auto-stops bridges every time that path is hit on a completed batch.
  - File: `server/index.js:1405`
  - Suggestion: apply the same transition-safe guard there instead of relying on `stopTrigger()` side effects. The issue explicitly called out this path, and the current PR adds no protection or test coverage for it.

### Decision
Requesting changes because the PR only fixes the poller branch and leaves the related `sendTriggerMessage()` path outside the requested fix scope.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR now applies the bridge auto-stop transition guard in both completion paths: the 30s poller and `sendTriggerMessage()`. The added tests also cover the new `sendTriggerMessage()` guard and state-update order.

### Findings
- None.

### Decision
Approving because the fix now matches issue #542 scope: bridge auto-stop fires once per completion transition, manually restarted bridges are no longer re-killed by the repeated completion path, and the existing new-batch auto-start logic remains intact.

PR #566, opened 2026-04-27
[#565] Guard agent registration against AC startup race

### LINKED ISSUE #565: [bug] Server crashes on agent registration failure during first project setup
## Problem

On a fresh install (v1.12.0), adding the first project via the setup wizard causes the server to **crash** with an uncaught error:

```
Error: Failed to register head: register head: fetch failed
    at buildAgentArgs (server/index.js:401)
```

The entire Node process exits, killing all dashboard functionality. The user sees "Failed to fetch" across all panels, `[session closed: 1006]` on terminals, and must restart manually.

## Root Cause

**Race condition:** The setup wizard triggers agent spawn before AgentChattr has finished starting. AC needs time to install (git clone + venv + pip install), bind its port, and become ready. The agent registration fetch to AC fails because AC isn't listening yet, and **line 401 throws a fatal error** instead of handling the failure gracefully.

```js
// server/index.js:400-401
if (!registration) {
  throw new Error(`Failed to register ${agentId}: ${registerAgent.lastError}`);
}
```

This `throw` is inside `buildAgentArgs()` which is called from `spawnAgentPty()` which is called from the WebSocket handler — none of which catch this error. It becomes an unhandled rejection and crashes Node.

## Fix

1. **Don't crash on registration failure.** Replace the `throw` with a graceful fallback — log a warning and spawn the agent without MCP/AC registration. The agent can still work (just without chat integration until AC comes up).

2. **Add retry logic.** Before giving up on registration, retry a few times with backoff (e.g. 3 attempts, 2s apart) to wait for AC to finish starting.

3. **Ensure AC is ready before spawning agents.** The setup flow should wait for AC's port to be reachable before attempting agent registration. A simple TCP probe loop (like the existing AC health monitor) would work.

## Reproduction

1. Fresh machine with Node 20+, Python 3.10+, gh CLI
2. `npx quadwork`
3. Complete setup wizard (connect repo, pick models, create workspaces)
4. Click "Start" or let auto-start trigger → server crashes

## Acceptance Criteria

- [ ] Server does NOT crash when AC registration fails
- [ ] Agent spawn degrades gracefully (works without AC registration)
- [ ] Registration is retried with backoff before giving up
- [ ] First project setup on a fresh install completes without errors
- [ ] Existing behavior for already-configured projects is unchanged

realproject7:
## Summary

- **Server no longer crashes** when AgentChattr registration fails during first project setup
- Extended AC readiness probe from 10s → 30s to cover first-install (clone + venv + pip) scenario; early-returns without registration if AC never becomes reachable
- Added `registerAgentWithRetry` (3 attempts, exponential backoff at 2s/4s) so transient AC startup delays are tolerated
- Moved `buildAgentArgs` inside `spawnAgentPty`'s try-catch as a safety net against any unexpected errors
- Agents spawned without registration still work — they just lack chat integration until the AC health monitor recovers them

## Test plan

- [ ] Fresh install: `npx quadwork` → complete setup wizard → start agents → server stays up, agents spawn (without chat until AC is ready)
- [ ] Existing project: restart server → agents register normally with AC, chat works immediately
- [ ] Kill AC process mid-run → agents continue running, health monitor restarts AC, agents recover chat

Fixes #565

### DIFF
--- a/server/agentchattr-registry.js
+++ b/server/agentchattr-registry.js
@@ -162,9 +162,26 @@ function stopHeartbeat(handle) {
   if (handle) clearInterval(handle);
 }
 
+/**
+ * Register an agent with retries and exponential backoff.
+ * Returns {name, token, slot} on success, null after all attempts fail.
+ * Uses registerAgent internally so registerAgent.lastError is populated.
+ */
+async function registerAgentWithRetry(serverPort, base, label = null, { force = false, attempts = 3, delayMs = 2000 } = {}) {
+  for (let i = 0; i < attempts; i++) {
+    const result = await registerAgent(serverPort, base, label, { force });
+    if (result) return result;
+    if (i < attempts - 1) {
+      await new Promise((res) => setTimeout(res, delayMs * Math.pow(2, i)));
+    }
+  }
+  return null;
+}
+
 module.exports = {
   waitForAgentChattrReady,
   registerAgent,
+  registerAgentWithRetry,
   deregisterAgent,
   startHeartbeat,
   stopHeartbeat,
--- a/server/index.js
+++ b/server/index.js
@@ -13,7 +13,7 @@ const {
   patchAgentchattrConfigForTelegramBridge,
   projectAgentchattrConfigPath,
 } = routes;
-const { waitForAgentChattrReady, registerAgent, deregisterAgent, startHeartbeat, stopHeartbeat } = require("./agentchattr-registry");
+const { waitForAgentChattrReady, registerAgent, registerAgentWithRetry, deregisterAgent, startHeartbeat, stopHeartbeat } = require("./agentchattr-registry");
 const { patchAgentchattrCss } = require("./install-agentchattr");
 const { startQueueWatcher, stopQueueWatcher } = require("./queue-watcher");
 
@@ -356,7 +356,13 @@ async function buildAgentArgs(projectId, agentId) {
       // write that into the per-agent MCP config file.
       const chattrInfo = resolveProjectChattr(projectId);
       acServerPort = Number(new URL(chattrInfo.url).port) || 8300;
-      await waitForAgentChattrReady(acServerPort);
+      // #565: extend timeout to 30s — first setup may need AC to install
+      // (git clone + venv + pip install) before it can bind a port.
+      const acReady = await waitForAgentChattrReady(acServerPort, 30000);
+      if (!acReady) {
+        console.warn(`[#565] Agent ${agentId}: AC not reachable on port ${acServerPort} after 30s. Spawning without chat integration.`);
+        return { args, acRegistrationName: null, acServerPort: null, acRegistrationToken: null, acInjectMode: null, acMcpHttpPort: mcpHttpPort || null };
+      }
       // #242: best-effort deregister any stale registration of the
       // canonical name (left over by a crashed previous QuadWork
       // session) so the fresh register lands at slot 1 instead of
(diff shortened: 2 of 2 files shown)


realproject7 (commented):
**re2 review — approve**

Clean three-layer fix that matches #565 requirements:

1. **`registerAgentWithRetry`** — straightforward retry wrapper with exponential backoff. Uses `registerAgent` internally so `lastError` is correctly populated from the final attempt.

2. **`waitForAgentChattrReady(port, 30000)`** — the function already accepts `timeoutMs` (default 10s) and returns a boolean, so the 30s override and boolean check are correct. Both `mcp_config` and `proxy_flag` branches get the same treatment.

3. **`buildAgentArgs` moved inside try-catch** — prevents unhandled rejections from crashing the Node process. `built` is only referenced inside the try block, so no scope leak.

Early-return shape `{ args, acRegistrationName: null, ... }` matches the normal return at line 414 — all consumer fields present.

No concerns.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The crash is fixed, but the new fallback path leaves agents permanently detached from AgentChattr when AC becomes ready after the 30s probe window. That violates the intended graceful-degradation behavior described in the issue and PR summary.

### Findings
- [high] Fallback to "spawn without chat integration" never recovers once AC finishes starting after the initial 30s wait.
  - File: `server/index.js:364`
  - Suggestion: Preserve enough AC session metadata to retry registration later, or schedule a post-spawn re-registration/reset once AC becomes reachable, instead of returning with all AC fields nulled out.

The same issue exists in the parallel `proxy_flag` branch at `server/index.js:402`. After either early return, the spawned session has `acRegistrationName/acServerPort/acRegistrationToken` set to `null`, so the later heartbeat and queue-watcher setup guarded at `server/index.js:597` and `server/index.js:616` never start. That means an agent started during first-install AC warm-up will keep running without chat for the rest of its lifetime unless something external triggers a full agent reset.

### Decision
Requesting changes because this misses the acceptance criterion that first-setup agents should degrade gracefully while AC is still starting, then recover normally once AC is available.

realproject7 (commented):
**re2 re-review — approve**

RE1's finding was valid — my initial review missed the permanent chat loss. The fix addresses it well:

1. **Early-return now preserves `acServerPort` and `acInjectMode`** — both `mcp_config` and `proxy_flag` branches pass these through instead of nulling them, giving the deferred recovery the context it needs.

2. **Deferred recovery guard** (`!session.acRegistrationName && session.acServerPort && session.acInjectMode`) fires only for agents that spawned without registration. Correct.

3. **Reuses `recoverFrom409`** — already handles deregister-stale → re-register → update session fields. No duplicate logic.

4. **Safety checks**: `!session.term` bail-out prevents recovery on exited agents (term set to null in `onExit` at line 619). `recoverFrom409` also guards on `!session.acServerPort` internally.

5. **Heartbeat + queue watcher** started only if recovery succeeds and handles aren't already set — no double-start risk.

6. **Fire-and-forget with `.catch(() => {})`** — appropriate since the health monitor is a second safety net.

No concerns on the updated diff.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The deferred recovery closes the earlier gap in session state, but it still does not restore chat for agents that were originally spawned without MCP injection flags.

### Findings
- [high] Deferred registration cannot recover chat for already-running `flag` / `proxy_flag` agents because the original spawn omitted the MCP CLI args.
  - File: `server/index.js:364`
  - Suggestion: If you want post-spawn recovery for agents started before AC is ready, you need to restart the agent once registration succeeds, or always spawn with a stable MCP transport that can be activated later without changing the process argv.

When `buildAgentArgs()` returns early in the `flag` branch, it returns `args` without pushing `--mcp-config` (`server/index.js:364`). The same happens in the `proxy_flag` branch without pushing `-c mcp_servers.agentchattr.url=...` (`server/index.js:405`). The new deferred path at `server/index.js:640-680` later calls `recoverFrom409()`, which only writes the MCP config file or updates the proxy token in place (`server/index.js:514-519`). That is not enough for an already-running agent process that never received the original MCP flag or proxy URL on startup. So the server no longer crashes, but the chat integration still won't actually attach for those running agents after deferred registration succeeds.

### Decision
Requesting changes because the new recovery path still doesn't meet the intended graceful-recovery behavior for agents started before AgentChattr is ready.

realproject7 (commented):
**re2 re-review (v3) — approve**

Much cleaner approach. Since MCP CLI args (`--mcp-config` / `-c mcp_servers...`) can only be set at process launch, the restart strategy is the correct fix — no partial recovery needed.

Verified:
1. **No infinite recursion** — after `stopAgentSession` + `spawnAgentPty`, the respawned agent hits `waitForAgentChattrReady` which should succeed (we just confirmed AC is up). Registration succeeds → `acRegistrationName` is set → deferred restart guard doesn't trigger again.
2. **`stopAgentSession`** does full cleanup (kill term, close WS, stop heartbeat, deregister) before respawn — no leaked resources.
3. **Guard checks** are correct: `!current.term || current.state !== 'running'` prevents restart of manually stopped agents.
4. **Fire-and-forget with `.catch(() => {})`** — appropriate for best-effort background work.

The only theoretical edge case is AC going down between the deferred wait and the respawned `buildAgentArgs`, which would trigger another deferred restart cycle — but this is bounded by AC stability and the manual-stop guard. Not a practical concern.

No issues.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The v3 change fixes the remaining recovery gap cleanly. Agents that start before AgentChattr is ready now restart once AC becomes reachable, so they relaunch with the required MCP CLI flags instead of trying to retrofit chat onto an already-running process.

### Findings
- None.

### Decision
Approving because the server crash is prevented, registration is retried with backoff, and the delayed-AC path now recovers using a restart that correctly rebuilds argv with MCP integration.

PR #568, opened 2026-04-27
[#560] Handle sendFile NotFoundError with error callback

### LINKED ISSUE #560: [bug] sendFile NotFoundError spam in server logs
## Problem

The server console is flooded with `NotFoundError: Not Found` stack traces from `res.sendFile()` at `server/routes.js:1151` (the `/api/uploads/:project/:filename` route).

The `fs.existsSync` check at line 1150 passes, but `res.sendFile` still throws — likely a race condition or Express 5's `send` module being stricter about file resolution.

## Stack Trace

```
NotFoundError: Not Found
    at createHttpError (.../send/index.js:861:12)
    at SendStream.error (.../send/index.js:168:31)
    at SendStream.pipe (.../send/index.js:468:14)
    at sendfile (.../express/lib/response.js:1014:8)
    at ServerResponse.sendFile (.../express/lib/response.js:411:3)
    at .../quadwork/server/routes.js:1151:7
```

## Fix

Add an error callback to `sendFile` to handle missing files gracefully instead of letting the error propagate as an unhandled stack trace:

```js
res.sendFile(filePath, (err) => {
  if (err && !res.headersSent) {
    res.status(404).json({ error: "Not found" });
  }
});
```

## Acceptance Criteria

- [ ] No more `NotFoundError` stack traces in server console for missing upload files
- [ ] Missing files still return 404 JSON response to the client
- [ ] Valid uploads still serve correctly

realproject7:
## Summary

- Added error callback to `res.sendFile()` in the `/api/uploads/:project/:filename` route
- Race conditions between `fs.existsSync` and `sendFile` (or Express 5's stricter `send` module) now return a clean 404 JSON instead of spamming the server console with `NotFoundError` stack traces
- Existing `fs.existsSync` pre-check kept as a fast path; the callback handles the edge case

## Test plan

- [ ] Request a valid upload file → served correctly
- [ ] Request a missing upload file → clean 404 JSON response, no stack trace in server console
- [ ] Server console no longer shows `NotFoundError: Not Found` spam

Fixes #560

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -1154,8 +1154,11 @@ router.get("/api/uploads/:project/:filename", (req, res) => {
   // existsSync and sendFile, or stricter file resolution in Express 5's
   // send module) is handled gracefully instead of spamming the server log.
   res.sendFile(filePath, (err) => {
-    if (err && !res.headersSent) {
+    if (!err || res.headersSent) return;
+    if (err.status === 404 || err.code === "ENOENT") {
       res.status(404).json({ error: "Not found" });
+    } else {
+      res.status(500).json({ error: "Internal server error" });
     }
   });
 });
--- a/server/routes.js
+++ b/server/routes.js
@@ -1150,7 +1150,14 @@ router.get("/api/uploads/:project/:filename", (req, res) => {
   }
   const filePath = path.join(CONFIG_DIR, project, "uploads", filename);
   if (!fs.existsSync(filePath)) return res.status(404).json({ error: "Not found" });
-  res.sendFile(filePath);
+  // #560: pass error callback so Express/send NotFoundError (race between
+  // existsSync and sendFile, or stricter file resolution in Express 5's
+  // send module) is handled gracefully instead of spamming the server log.
+  res.sendFile(filePath, (err) => {
+    if (err && !res.headersSent) {
+      res.status(404).json({ error: "Not found" });
+    }
+  });
 });
 
 // ─── Projects (dashboard aggregation) ──────────────────────────────────────


realproject7 (commented):
**re2 review — approve**

Correct fix for the TOCTOU race between `existsSync` and `sendFile`. Error callback with `!res.headersSent` guard returns a clean 404 JSON. No concerns.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The fix is narrowly scoped in the right place, but the callback currently converts every `sendFile` error into a 404, which hides real server-side failures instead of only handling the `NotFoundError` race from the issue.

### Findings
- [medium] All `sendFile` failures are now reported as 404 `Not found`, including non-ENOENT errors.
  - File: `server/routes.js:1156`
  - Suggestion: Only translate the specific missing-file case (`err.status === 404`, `err.code === "ENOENT"`, or equivalent), and let unexpected errors propagate via `next(err)` or a 500 so real permission/I/O/path-resolution bugs are not silently mislabeled as missing files.

### Decision
Requesting changes because this would mask legitimate server errors, making debugging harder and changing the route's semantics beyond the `NotFoundError` spam bug described in #560.

realproject7 (commented):
**re2 re-review — approve**

Error handler now correctly narrows: `err.status === 404 || err.code === "ENOENT"` → 404, everything else → 500. RE1's concern addressed. No issues.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up narrows the callback correctly. Missing-file races now return a clean 404 without console spam, while non-404 `sendFile` failures are no longer mislabeled as missing files.

### Findings
- None.

### Decision
Approving because the route now handles the `NotFoundError` case from #560 without masking unrelated server-side errors.

PR #576, opened 2026-04-27
[#572] Auto-restart unregistered agents on AC recovery

### LINKED ISSUE #572: [bug] No retry/recovery for AC process itself — only agents retry connecting to AC
## Problem

If AgentChattr crashes at runtime (after a successful start), there's no automatic recovery path for agents that lose chat integration. The existing health monitor (`acHealthCheck` at `server/index.js:2099`) can detect AC is down and restart it, but agents that were spawned without AC registration (via #565 graceful degradation) are never automatically restarted to gain chat integration.

## Context

With #573 (separate install from launch), the wizard no longer spawns AC — only the server does. This simplifies the problem to runtime recovery only:

1. Server starts AC → AC is healthy → agents register successfully ✅
2. AC crashes mid-session → health monitor restarts AC → **but agents still lack chat integration** ❌
3. AC is slow to start on `quadwork start` → agents spawn without chat → #565 deferred restart waits 60s → **if AC takes longer, agents stay without chat** ❌

## Fix

1. **When AC recovers after a crash**, trigger restarts for any agents that are running without AC registration (check `session.acRegistrationName === null && session.acServerPort`)
2. **Connect to the #565 deferred restart logic** — if the deferred restart times out (60s), register a callback with the health monitor so when AC eventually comes up, the agent is restarted
3. **Health monitor should log AC recovery** so the operator can see it happened

## Acceptance Criteria

- [ ] If AC crashes and is restarted by health monitor, agents without chat integration are automatically restarted
- [ ] If #565 deferred restart times out, agent is still recovered when AC comes up later
- [ ] AC recovery is logged: `[health] AC recovered on port {port} — restarting {n} agents for chat integration`
- [ ] No manual intervention needed to recover from AC crashes

realproject7:
## Summary
- When health monitor detects AC recovered from a crash, scans `agentSessions` for agents running without AC registration (`acRegistrationName === null`) and stop+respawns them
- Connects to #565 deferred restart: if the 60s timeout expires, the health monitor picks up recovery when AC eventually comes up
- Adds timeout log message to #565 deferred restart so operators know the handoff happened

## Changes
- `server/index.js`: Added `restartUnregisteredAgents(projectId)` function — scans sessions, filters unregistered agents, stop+respawns each
- `server/index.js`: Health monitor recovery branch (alive + consecutiveFailures > 0) calls `restartUnregisteredAgents`
- `server/index.js`: #565 deferred restart logs when 60s timeout expires

## Test plan
- [ ] Start agents before AC is ready → agents spawn without chat → AC starts → health monitor triggers agent restart → agents gain chat integration
- [ ] AC crashes mid-session → health monitor restarts AC → unregistered agents are respawned
- [ ] #565 deferred restart timeout → log message appears → AC comes up later → health monitor restarts agents
- [ ] Agents already registered (acRegistrationName set) are NOT restarted on recovery
- [ ] Manually stopped agents are NOT restarted on recovery

Fixes #572

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -641,7 +641,12 @@ async function spawnAgentPty(project, agent) {
     if (!session.acRegistrationName && session.acServerPort && session.acInjectMode) {
       const deferredRestart = async () => {
         const ready = await waitForAgentChattrReady(session.acServerPort, 60000);
-        if (!ready) return;
+        if (!ready) {
+          // #572: log timeout so operators know the health monitor will
+          // handle recovery when AC eventually comes up.
+          console.log(`[#565] Agent ${agent}: AC not reachable after 60s — health monitor will restart agent when AC recovers.`);
+          return;
+        }
         // Guard: agent may have been stopped manually while we waited.
         const current = agentSessions.get(key);
         if (!current || !current.term || current.state !== "running") return;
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The intended recovery flow is correct, but the new restart helper passes the wrong project value into `spawnAgentPty()`, so the agent respawn step fails in the exact path this PR adds.

### Findings
- [high] AC recovery restart path passes a project object where `spawnAgentPty()` requires a project id string.
  - File: `server/index.js:2122`, `server/index.js:541`, `server/config.js:67`
  - Suggestion: call `spawnAgentPty(projectId, agentId)` or otherwise pass the project id string consistently through the helper.
  - Details: `restartUnregisteredAgents()` loads the full config entry into `project`, then does `await spawnAgentPty(project, agentId)`. But `spawnAgentPty()` builds `key = \`${project}/${agent}\`` and resolves cwd/command through `resolveAgentCwd(project, agent)` / `resolveAgentCommand(project, agent)`, both of which compare `p.id === projectId`. Passing the object makes those lookups fail, so agents stopped for recovery will not come back.

### Decision
Requesting changes because the new recovery path breaks on its own respawn call and therefore does not satisfy the runtime crash-recovery acceptance criteria.

realproject7 (commented):
**RE2 review — PR #576 (#572)**

Checked against acceptance criteria:

| Criterion | Status |
|-----------|--------|
| AC crash → health monitor restarts AC → agents without chat restarted | ✅ `restartUnregisteredAgents` fires inside `consecutiveFailures > 0` recovery path |
| #565 deferred restart timeout → still recovered later | ✅ Timeout now logs message; health monitor handles recovery when AC comes up |
| Recovery logged with port + agent count | ✅ `[health] AC recovered on port {port} — restarting {n} agent(s)` |
| No manual intervention needed | ✅ Fully automatic: health monitor → AC restart → agent respawn |

**Recovery trigger is correctly guarded** — `restartUnregisteredAgents` only fires when `consecutiveFailures > 0` (AC was previously detected as down), preventing spurious restarts on every health tick.

**Agent filter is correct** — `session.acRegistrationName` null + `state === "running"` + `acServerPort`/`acInjectMode` present correctly identifies agents that launched without chat integration.

**No race with #565 deferred restart** — if the health monitor stop+respawns an agent while its deferred restart timer is still pending, the deferred restart's guard (`!current.term || current.state !== "running"`) bails safely. If the deferred restart succeeds first, `acRegistrationName` gets set and `restartUnregisteredAgents` skips that agent via `if (session.acRegistrationName) continue`.

**Fire-and-forget is appropriate** — `restartUnregisteredAgents` is async with `.catch()`, so it doesn't block the health check loop for other projects.

Minor note: `samplePort` line could be simplified (the `toRestart.length > 0` check is always true since we return early when `length === 0`), but it's harmless.

No blocking issues.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up fixes the broken recovery call by passing the project id string into `spawnAgentPty()`. The runtime-only recovery flow now matches the ticket intent.

### Findings
- None.

### Decision
Approving because AC recovery now stop+respawns unregistered agents through the correct `spawnAgentPty(projectId, agentId)` path, while preserving the existing guards against manually stopped sessions.

PR #583, opened 2026-04-27
[#580] Add waitForAgentChattrReady after every AC spawn

### LINKED ISSUE #580: [#579-1] Add waitForAgentChattrReady after every AC spawn
## Context
Follow-up from #579 (Symptom 1 & 2). On fresh installs where AC takes several seconds to bind port 8300, the health monitor's first 30s tick sees port down → triggers auto-restart → causes ghost agent cascade. Existing installs are unaffected because AC starts in ~2s — but the fix is needed for slow-start environments.

## Problem
Both `bin/quadwork.js` (`cmdStart`) and `server/index.js` (`spawnChattr`) declare AC success the moment `spawn()` returns a PID. AC takes several seconds to actually bind 8300 (templates load, MCP servers start, FastAPI lifespan). The health monitor and dashboard history-snapshot fetch check before that's done → 502 + false-down → unnecessary auto-restart cascade.

## Fix
Add `await waitForAgentChattrReady(port, 30000)` after every AC spawn, before declaring success. The function already exists in `server/index.js` — reuse it.

### Call sites to patch

1. **`server/index.js` — `spawnChattr()` function** (around the `setProc({ process: child, state: "running" })` line)
   - After confirming `child.pid` is valid, before setting state to "running"
   - `await waitForAgentChattrReady(chattrPort, 30000)`
   - If not ready after 30s, set state to "error" instead of "running"

2. **`bin/quadwork.js` — `cmdStart()` AC spawn loop** (around the `ok("AgentChattr started for ...")` line)
   - After confirming `acProc.pid` is valid
   - Import or inline a simple port-check loop (bin/ doesn't have access to `waitForAgentChattrReady` from server/)
   - Can be a simple `fetch` poll: retry `http://127.0.0.1:${port}/api/health` every 2s for up to 30s

3. **`server/routes.js` — `/api/agentchattr/:id/restart` endpoint** — see #579-3 for this one

### Safety
When AC starts fast (existing installs, pinned commit), the wait resolves in 1-2s. No behavioral change. Only prevents false-down detection on slow starts.

## Acceptance criteria
- [ ] `spawnChattr()` waits for port before setting state to "running"
- [ ] `cmdStart()` waits for port before printing success
- [ ] Existing fast-start environments see no delay beyond AC's actual bind time
- [ ] `npm run build` passes

Fixes part of #579

realproject7:
## Summary
- `spawnChattr()` in `server/index.js` now `await`s `waitForAgentChattrReady(port, 30000)` before setting state to "running" — if AC doesn't bind within 30s, state is set to "error" instead
- `cmdStart()` in `bin/quadwork.js` polls `/api/health` every 2s (up to 30s) before printing success; warns on timeout
- On fast-start installs the wait resolves in 1-2s with no behavioral change; prevents false-down detection on slow starts that caused ghost agent cascades

## Test plan
- [ ] Fresh install with slow AC startup: verify no auto-restart cascade
- [ ] Existing fast-start install: verify startup completes in ~1-2s with no extra delay
- [ ] `npm run build` passes

Fixes #580

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -1564,7 +1564,30 @@ function migrateLoopGuardDefaults(config) {
   }
 }
 
-function cmdStart() {
+/**
+ * #580: Poll AC health endpoint until it responds 200, or timeout.
+ * Simple inline implementation for bin/ (no access to server/ modules).
+ */
+async function waitForAcHealth(baseUrl, timeoutMs = 30000) {
+  const http = require("http");
+  const deadline = Date.now() + timeoutMs;
+  const healthUrl = `${baseUrl}/api/health`;
+  while (Date.now() < deadline) {
+    const ok = await new Promise((resolve) => {
+      const req = http.get(healthUrl, (res) => {
+        res.resume();
+        resolve(res.statusCode >= 200 && res.statusCode < 400);
+      });
+      req.on("error", () => resolve(false));
+      req.setTimeout(2000, () => { req.destroy(); resolve(false); });
+    });
+    if (ok) return true;
+    await new Promise((r) => setTimeout(r, 2000));
+  }
+  return false;
+}
+
+async function cmdStart() {
   console.log("\n  QuadWork Start\n");
 
   const config = readConfig();
@@ -1638,7 +1661,14 @@ function cmdStart() {
     acProc.on("error", () => {});
     acProc.unref();
     if (acProc.pid) {
-      ok(`AgentChattr started for ${project.id} from ${projectAcDir} (PID: ${acProc.pid})`);
+      // #580: wait for AC to bind its port before declaring success.
+      const acUrl = project.agentchattr_url || "http://127.0.0.1:8300";
+      const acReady = await waitForAcHealth(acUrl, 30000);
+      if (acReady) {
+        ok(`AgentChattr started for ${project.id} from ${projectAcDir} (PID: ${acProc.pid})`);
+      } else {
+        warn(`AgentChattr spawned for ${project.id} (PID: ${acProc.pid}) but did not become ready within 30s`);
+      }
       log(`  Log: ${acLogPath}`);
       acPids.push(acProc.pid);
     }
--- a/server/index.js
+++ b/server/index.js
@@ -860,7 +860,7 @@ async function handleAgentChattr(req, res) {
     } catch {}
   }
 
-  function spawnChattr() {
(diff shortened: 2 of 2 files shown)


realproject7 (commented):
**RE2 APPROVED** ✓

Reviewed diff against #580 acceptance criteria:

1. ✅ `spawnChattr()` now awaits `waitForAgentChattrReady(chattrPort, 30000)` before setting state to "running" — sets "error" state on timeout
2. ✅ `cmdStart()` polls `/api/health` every 2s for up to 30s before printing success — prints warning on timeout
3. ✅ Fast-start environments unaffected (resolves as soon as AC binds)
4. ✅ Build passes per Dev

Notes:
- Correctly reuses existing `waitForAgentChattrReady` from `agentchattr-registry.js` in server/
- `bin/quadwork.js` correctly inlines its own HTTP-based health check (no access to server modules)
- All three `spawnChattr()` call sites (start, restart, update) properly awaited
- Timeout behavior is graceful: warns/sets error state but doesn't crash

No issues found.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The readiness wait is added in the right places, but the server start/restart paths still report success when AgentChattr never becomes ready within 30s.

### Findings
- [high] Start/restart can return success even after readiness timeout
  - File: `server/index.js:919`
  - Suggestion: when `waitForAgentChattrReady(chattrPort, 30000)` returns false, treat the spawn as failed for the caller as well. Right now `spawnChattr()` sets `state: "error"` but still returns `child`, so the `start`/`restart` handlers continue to send `res.json({ ok: true, state: "running", pid })` at `server/index.js:1007` and `server/index.js:1080`. Return `null` or throw on timeout so the HTTP response matches the actual error state.

### Decision
Requesting changes because the main failure case this ticket is trying to harden still surfaces as a successful start/restart to API callers, which can leave the dashboard and automation believing AC is up when it already timed out.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The readiness wait is now applied in the intended spawn paths, and the timeout case correctly propagates as a failed start/restart instead of a false success.

### Findings
- None.

### Decision
Approving because the server-side timeout path now returns `null` from `spawnChattr()`, which makes the `start` and `restart` handlers report the actual error state. That closes the regression from the prior revision and matches issue #580's acceptance criteria.

PR #584, opened 2026-04-27
[#581] Dedupe auto-reset triggers — at most once per AC restart

### LINKED ISSUE #581: [#579-2] Dedupe auto-reset triggers — at most once per AC restart
## Context
Follow-up from #579 (Symptom 1 & 2). When the health monitor restarts AC, `auto-reset 4 agent(s)` fires three times for one logical restart event. Each reset re-registers all four agents, pushing AC's slot suffix counter up (head-2, head-3, ..., head-13) and creating ghost identities that pollute the chat.

## Problem
Multiple code paths trigger agent resets after an AC restart:
- Health monitor recovery in `server/index.js` calls `restartUnregisteredAgents()` (#572)
- The AC restart endpoint chains a reset (#447)
- Bridge migration (`[bridge-migrate]`) and ghost-fix patches may also trigger resets

These fire independently for the same logical event → triple reset → suffix proliferation.

## Fix
Add a per-project debounce/guard so that `auto-reset` can only fire **once per AC restart event**.

### Implementation approach
In `server/index.js`, add a timestamp guard to the health monitor's recovery branch:

```js
// In _acHealth state, add:
lastResetTimestamp: new Map()  // projectId → Date.now()

// In the recovery branch (where restartUnregisteredAgents is called):
const lastReset = _acHealth.lastResetTimestamp.get(project.id) || 0;
const now = Date.now();
if (now - lastReset < 60000) {
  // Already reset agents within the last 60s — skip
  return;
}
_acHealth.lastResetTimestamp.set(project.id, now);
restartUnregisteredAgents(project.id);
```

Also check the `/api/agentchattr/:id/restart` endpoint in `server/routes.js` — if it chains its own agent reset, it should set the same timestamp so the health monitor doesn't double-fire.

### Safety
On fast-start environments where AC recovers before the first health tick, `restartUnregisteredAgents` finds 0 agents to restart (all already registered). The guard is a no-op. No behavioral change.

## Acceptance criteria
- [ ] `auto-reset` fires at most once per AC restart event
- [ ] Health monitor and restart endpoint share the debounce state
- [ ] No ghost suffixed identities created during normal AC restart cycle
- [ ] `npm run build` passes

Fixes part of #579

realproject7:
## Summary
- Add `lastResetTimestamp` map to `_acHealth` — per-project 60s debounce
- Restart endpoint sets the timestamp before its setTimeout reset fires
- Health monitor recovery branch checks the timestamp before calling `restartUnregisteredAgents()`
- Prevents triple-reset → ghost suffix proliferation on a single logical AC restart

## Test plan
- [ ] Trigger AC restart: verify `auto-reset N agent(s)` logs only once (not 3×)
- [ ] Fast-start environment: verify no behavioral change (reset still fires once)
- [ ] `npm run build` passes

Fixes #581

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1080,8 +1080,9 @@ async function handleAgentChattr(req, res) {
       }
       res.json({ ok: true, state: "running", pid: child.pid });
       // #447: auto-reset all agents after AC restart so they get
-      // fresh MCP tokens. #581: set debounce timestamp only on success
-      // so the health monitor can still recover if this reset fails.
+      // fresh MCP tokens. #581: mark reset as scheduled immediately
+      // so the health monitor skips its own reset while ours is in-flight.
+      _acHealth.resetState.set(projectId, { status: "scheduled", timestamp: Date.now() });
       setTimeout(async () => {
         try {
           const resetResp = await fetch(`http://127.0.0.1:${PORT}/api/agents/${encodeURIComponent(projectId)}/reset`, {
@@ -1089,12 +1090,14 @@ async function handleAgentChattr(req, res) {
           });
           if (resetResp.ok) {
             const resetData = await resetResp.json();
-            _acHealth.lastResetTimestamp.set(projectId, Date.now());
+            _acHealth.resetState.set(projectId, { status: "succeeded", timestamp: Date.now() });
             console.log(`[agentchattr] ${projectId} auto-reset ${resetData.restarted} agent(s) after AC restart`);
           } else {
+            _acHealth.resetState.set(projectId, { status: "failed", timestamp: Date.now() });
             console.warn(`[agentchattr] ${projectId} agent reset after AC restart returned ${resetResp.status}`);
           }
         } catch (err) {
+          _acHealth.resetState.set(projectId, { status: "failed", timestamp: Date.now() });
           console.warn(`[agentchattr] ${projectId} agent reset after AC restart failed: ${err.message || err}`);
         }
       }, 2000);
@@ -2134,8 +2137,9 @@ const _acHealth = {
   // Per-project: { lastRestart: timestamp, consecutiveFailures: number }
   state: new Map(),
   intervalHandle: null,
-  // #581: per-project debounce — auto-reset fires at most once per 60s
-  lastResetTimestamp: new Map(),
+  // #581: per-project reset state — prevents duplicate resets per restart event.
+  // Values: { status: "scheduled"|"succeeded"|"failed", timestamp: number }
+  resetState: new Map(),
 };
 
 function isPortAlive(port) {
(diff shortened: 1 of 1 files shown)


realproject7 (commented):
**RE2 APPROVED** ✓

Reviewed against #581 acceptance criteria:

1. ✅ Auto-reset fires at most once per 60s per project — health monitor checks `lastResetTimestamp` before calling `restartUnregisteredAgents()`
2. ✅ Health monitor and restart endpoint share debounce state via `_acHealth.lastResetTimestamp` Map
3. ✅ Restart endpoint sets timestamp *before* its own chained reset, preventing the health monitor from double-firing on recovery detection
4. ✅ Build passes per Dev

Clean, minimal implementation. No issues found.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The shared debounce wiring is close, but it now suppresses the health-monitor recovery fallback before the restart endpoint's own reset has actually succeeded.

### Findings
- [high] Debounce can drop the only recovery reset on transient restart-endpoint reset failure
  - File: `server/index.js:1085`
  - Suggestion: don't mark the debounce as consumed before the `/api/agents/:project/reset` path has actually succeeded, or track an explicit restart-event/in-flight state instead of a blind timestamp. Right now the restart endpoint sets `_acHealth.lastResetTimestamp` immediately, then schedules the reset 2s later. On the first healthy tick after recovery, the health monitor sees the fresh timestamp and skips `restartUnregisteredAgents()` at `server/index.js:2176`. If the scheduled reset then fails or returns non-OK, `health.consecutiveFailures` is reset to `0` and that recovery branch will not run again on later healthy ticks, so agents can remain unregistered with no fallback reset at all.

### Decision
Requesting changes because this turns the previous recovery fallback into a potential lost-reset case, which can leave agents without restored chat integration after an AC restart.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Moving the timestamp to the successful reset branch fixes the lost-fallback problem, but it reopens a race where the health monitor can still double-fire before that delayed reset records the debounce.

### Findings
- [high] Recovery tick can still double-fire before the delayed reset marks the debounce
  - File: `server/index.js:1085`
  - Suggestion: track the restart event as soon as the restart endpoint commits to scheduling its reset, but keep a separate success/failure state so the health monitor knows whether to skip, wait, or recover. Right now the restart endpoint only updates `_acHealth.lastResetTimestamp` after the delayed reset succeeds, 2s later. If AC comes back and the next 30s health tick lands in that window, the recovery branch at `server/index.js:2176` still sees no recent reset and runs `restartUnregisteredAgents()` anyway. Then the delayed `/api/agents/:project/reset` fires too, so one logical restart can still produce two resets.

### Decision
Requesting changes because the PR still does not guarantee "at most once per AC restart event"; it trades the previous lost-fallback bug for a remaining duplicate-reset race.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The three-state marker fixes the duplicate-reset race, but it still loses the health-monitor fallback if recovery is observed while the endpoint reset is merely scheduled and that scheduled reset later fails.

### Findings
- [high] Scheduled-state skip still clears the only recovery opportunity before reset success is known
  - File: `server/index.js:2173`
  - Suggestion: don't consume the health-monitor recovery transition just because reset state is `scheduled`. You need either a follow-up retry path after `scheduled -> failed`, or separate recovery bookkeeping so the monitor can attempt fallback once the in-flight reset definitively fails. Right now, if AC is healthy on the first tick after restart, the recovery branch sees `status === "scheduled"` and skips `restartUnregisteredAgents()`, then immediately sets `health.consecutiveFailures = 0`. If the delayed endpoint reset fails a moment later and flips to `failed`, there is no later `health.consecutiveFailures > 0` recovery tick to trigger the fallback, so agents can still remain unregistered with no reset.

### Decision
Requesting changes because the PR still doesn't guarantee one successful recovery reset per restart event; this version can still produce zero resets after a scheduled reset failure.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The three-state reset marker now handles both sides of the race correctly: in-flight endpoint resets suppress duplicate recovery resets, and failed in-flight resets preserve the health-monitor retry path on the next healthy tick.

### Findings
- None.

### Decision
Approving because the updated `scheduled` handling keeps `consecutiveFailures` latched until the in-flight reset resolves, which preserves the fallback path after failure while still preventing duplicate resets for the same restart event.

PR #585, opened 2026-04-27
[#582] Harden restart endpoint — SIGKILL fallback + logging

### LINKED ISSUE #582: [#579-3] Restart endpoint must re-spawn AC after killing it
## Context
Follow-up from #579 (Symptom 3). Clicking "Restart server and agents" in the dashboard kills AC but never re-spawns it. Agents then hit the 30s/60s timeout and fall back to no-chat mode. `lsof -iTCP:8300` shows nothing listening.

## Problem
The `/api/agentchattr/:id/restart` endpoint in `server/routes.js` calls `killProcessOnPort` (or similar) to stop AC, but the spawn-new step either:
- Doesn't execute (code path missing)
- Executes but the new AC dies silently
- Doesn't wait for port to be free before spawning

The AC log ends with `Finished server process` shutdown messages and no new startup banner, confirming the spawn never happened.

## Fix
Audit and fix the restart handler in `server/routes.js`:

1. **Trace the handler** — find the `/api/agentchattr/:id/restart` (or `/api/agentchattr/:id` POST with action=restart) route
2. **Verify it calls `spawnChattr`** (or equivalent) after killing the old process
3. **Add `waitForPortFree`** before spawning if not already present — the old process may take a moment to release port 8300
4. **Add `waitForAgentChattrReady`** after spawn (from ticket #579-1) so the restart endpoint doesn't return success before AC is actually bound
5. **Log the spawn attempt** so failures are visible in server output

### Safety
On environments where restart already works, the added waits resolve quickly. No behavioral change — just ensures the spawn step actually executes and confirms AC is alive before returning.

## Acceptance criteria
- [ ] Restart endpoint kills old AC, waits for port free, spawns new AC, waits for port bound
- [ ] After clicking Restart, `lsof -iTCP:8300` shows AC listening
- [ ] Agents re-register successfully after restart
- [ ] Server log shows spawn confirmation after restart
- [ ] `npm run build` passes

Fixes part of #579

realproject7:
## Summary
- Restart handler now logs each step (kill, port free, spawn, ready) for diagnosability
- Extended port-free wait to 5s after SIGTERM, with SIGKILL fallback if still occupied
- Returns 500 with clear error if port cannot be freed even after SIGKILL
- `killProcessOnPort` accepts optional signal parameter for the escalation
- Combined with #580's `waitForAgentChattrReady`, the restart path now guarantees AC is bound before returning success

## Test plan
- [ ] Click Restart in dashboard: verify AC re-spawns and `lsof -iTCP:8300` shows it listening
- [ ] Check server logs: should show kill → port free → spawn → ready sequence
- [ ] Simulate stuck AC (ignore SIGTERM): verify SIGKILL fallback kicks in
- [ ] `npm run build` passes

Fixes #582

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1049,8 +1049,10 @@ async function handleAgentChattr(req, res) {
       killProcessOnPort(chattrPort, "SIGKILL");
       portFree = await waitForPortFree(chattrPort, 3000);
       if (!portFree) {
-        console.error(`[agentchattr] ${projectId} port ${chattrPort} still occupied after SIGKILL — cannot restart`);
-        return res.status(500).json({ ok: false, state: "error", error: `Port ${chattrPort} still occupied — cannot restart` });
+        const portErr = `Port ${chattrPort} still occupied — cannot restart`;
+        console.error(`[agentchattr] ${projectId} ${portErr}`);
+        setProc({ process: null, state: "error", error: portErr });
+        return res.status(500).json({ ok: false, state: "error", error: portErr });
       }
     }
     console.log(`[agentchattr] ${projectId} restart: port ${chattrPort} is free, spawning AC`);
--- a/server/index.js
+++ b/server/index.js
@@ -929,7 +929,7 @@ async function handleAgentChattr(req, res) {
   // #386: Kill any process listening on the AC port. Handles orphaned
   // processes that survive QuadWork restarts (detached + unref'd spawns
   // lose their tracked reference when the Node process recycles).
-  function killProcessOnPort(port) {
+  function killProcessOnPort(port, signal = "SIGTERM") {
     try {
       const pids = execFileSync("lsof", ["-ti", `TCP:${port}`, "-sTCP:LISTEN"], {
         encoding: "utf-8",
@@ -940,7 +940,7 @@ async function handleAgentChattr(req, res) {
       for (const line of pids.split("\n")) {
         const pid = parseInt(line, 10);
         if (pid > 0) {
-          try { process.kill(pid, "SIGTERM"); } catch {}
+          try { process.kill(pid, signal); } catch {}
         }
       }
     } catch {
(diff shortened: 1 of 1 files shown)


realproject7 (commented):
**RE2 APPROVED** ✓

Reviewed against #582 acceptance criteria:

1. ✅ Restart endpoint: kills old AC → waits 5s for port free → SIGKILL fallback → waits 3s more → returns 500 if still occupied → spawns new AC via `await spawnChattr()` (which includes `waitForAgentChattrReady` from #580)
2. ✅ `lsof` will show AC listening after restart — `spawnChattr()` only returns non-null after readiness confirmed
3. ✅ Agents re-register: existing `syncChattrToken` + auto-reset chain preserved
4. ✅ Step-by-step logging at kill, port-free, spawn, and failure points
5. ✅ Build passes per Dev

Notes:
- `killProcessOnPort` now accepts a signal parameter — clean extension, no change to existing SIGTERM callers
- SIGKILL escalation is the right fallback for stuck AC processes that ignore SIGTERM
- Returning 500 when port can't be freed is correct — spawning on an occupied port was the original bug

No issues found.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The restart flow now does the kill -> wait -> respawn sequence the ticket asked for, but the new hard-failure path leaves the internal AC state inconsistent.

### Findings
- [high] Occupied-port failure returns 500 but leaves AC tracked as intentionally stopped
  - File: `server/index.js:1043`
  - Suggestion: when restart gives up because the port is still occupied after SIGKILL, call `setProc({ process: null, state: "error", error: ... })` before returning the 500. Right now the branch sets `state: "stopped"` at `server/index.js:1043`, then if the port still cannot be freed it returns a 500 at `server/index.js:1052` without changing that state. That means the dashboard will show AC as stopped rather than errored, and the health monitor will skip it (`proc.state === "stopped"`), so the failure is effectively hidden instead of surfacing as a persistent restart error.

### Decision
Requesting changes because the new restart hard-failure path should surface an actual error state, not look like an intentional stop.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The restart handler now does the required kill -> wait -> respawn -> ready-confirmed flow, and the occupied-port hard-failure path correctly persists an error state instead of looking like an intentional stop.

### Findings
- None.

### Decision
Approving because the updated restart branch now surfaces restart failure consistently to both the HTTP caller and QuadWork's internal AC state, while the success path waits for the respawned AC instance to be ready before returning.

PR #589, opened 2026-04-28
[#579] Harden AC startup race: detect running AC, grace period, reset dedup

### LINKED ISSUE #579: Fresh 1.13.1 install: AC startup race → triple auto-reset → ghost agent identities (head-N, re1-N) + Restart button leaves AC dead
# Fresh 1.13.1 install: AC startup race → triple auto-reset cascade → ghost agent identities + Restart button leaves AC dead

## Context
Builds on #577 (--config flag fix, shipped in 1.13.1). The argparse error is gone — AC now starts and binds 8300. But on a clean install (`rm -rf ~/.quadwork ~/.npm/_npx/*/*/quadwork`, then `npx quadwork@latest start` → create project via dashboard), three layered issues surface that compound into a broken chat experience.

## Symptom 1 — startup race triggers triple auto-reset
Server log on first project creation:
```
QuadWork server listening on http://127.0.0.1:8400
[health] AC health monitor started (30s interval)
Cloning into '/Users/<user>/Projects/<project>'...
Preparing worktree (checking out 'worktree-head')
Preparing worktree (checking out 'worktree-re1')
Preparing worktree (checking out 'worktree-re2')
Preparing worktree (checking out 'worktree-dev')
[snapshot] <project> history fetch returned 502; skipping snapshot
[snapshot] <project> history fetch returned 502; skipping snapshot
[health] AC for <project> on port 8300 is down (failure 1/3) — auto-restarting
[snapshot] <project> history fetch returned 502; skipping snapshot
[health] AC for <project> auto-restarted (PID: <n>)
[agentchattr] <project> auto-reset 4 agent(s) after AC restart
[agentchattr] <project> auto-reset 4 agent(s) after AC restart
[agentchattr] <project> auto-reset 4 agent(s) after AC restart
[health] AC for <project> recovered (port 8300 alive)
```

What happens (timeline):
1. Worktrees clone successfully.
2. Dashboard fetches AC history → 502 (AC not bound yet).
3. Health monitor's first 30s tick sees port 8300 down → auto-restarts AC.
4. **`auto-reset 4 agent(s)` fires three times in a row.** Each reset re-registers all four agents via `/api/register` against AC.
5. AC eventually recovers.

The triple-fire of `auto-reset` is the trigger for Symptom 2.

## Symptom 2 — ghost agent identity proliferation (most user-visible)
After the cascade above, AC's registry has accumulated a long list of suffixed identities. Evidence from one session's `agentchattr.log`:

- **597 total `409 Conflict` heartbeat responses** across the log
- Distinct ghost identities observed: `head-2 … head-13`, `dev-1 … dev-10`, `re1-1 … re1-13`, `re2-2 … re2-10`
- Per-identity failed-heartbeat counts: 13–18 for each ghost slot

This becomes user-visible in the AGENTCHATTR primary chat: a single user prompt (`@head ping dev, re1, and re2 to check if they're online`) produced ~25 reply messages, most from suffixed ghost agents searching for mentions addressed to *their own* suffix and finding none. Excerpt (paraphrased, names preserved):

> head-13: I checked #general for @head-2 mentions and found none. The only current message is addressed to @head, so I am not taking action on it as @head-2.

> head-7: I attempted to send as head-3, but AgentChattr reports that session as stale and will not let me reclaim it because head-3 is already claimed.

> re1-12: @user I re-read #general for @re1-3. The only @re1-3 occurrence is the prior status note saying no direct actionable mention was found.

The agents are confused about their own identity, repeatedly searching for mentions that don't exist, and answering as ghost slots. The user-prompt-to-noise ratio is roughly 1:25.

### Likely mechanism
- AC's slot system suffixes a registration when the canonical name is already claimed (`head-2` if `head` is taken).
- Quadwork's auto-restart path tries to deregister the old slot and re-register with `force: true` to claim slot 1 (per code comments around `server/index.js:380-388, 417-419` and `#478`).
- During the triple-fire cascade, the deregister-then-register sequence isn't reliably claiming slot 1 — every retry pushes the suffix counter up by one.
- The agents keep heartbeating as their previously-acquired ghost slot, so old suffixed identities stay alive in AC's registry until their 120s crash timeout (per `[idle-fix] increased crash timeout to 120s (#502)`).

## Symptom 3 — Dashboard "Restart server and agents" leaves AC dead
After Symptom 2 manifested, clicking the **Restart** button (server controls) produces:
```
[agentchattr] <project> auto-reset 4 agent(s) after AC restart
[agentchattr] <project> auto-reset 4 agent(s) after AC restart
[agentchattr] <project> auto-reset 4 agent(s) after AC restart
[health] AC for <project> recovered (port 8300 alive)
[#565] Agent head: AC not reachable on port 8300 after 30s. Spawning without chat integration.
[#565] Agent re1:  AC not reachable on port 8300 after 30s. Spawning without chat integration.
[#565] Agent head: AC not reachable after 60s — health monitor will restart agent when AC recovers.
[#565] Agent re2:  AC not reachable on port 8300 after 30s. Spawning without chat integration.
[#565] Agent re1:  AC not reachable after 60s — health monitor will restart agent when AC recovers.
[#565] Agent dev:  AC not reachable on port 8300 after 30s. Spawning without chat integration.
[#565] Agent re2:  AC not reachable after 60s — health monitor will restart agent when AC recovers.
[#565] Agent dev:  AC not reachable after 60s — health monitor will restart agent when AC recovers.
```

End state observed via `lsof -iTCP:8300` and `ps aux | grep run.py`:
- Nothing listening on 8300
- No `run.py` Python process
- AC log ends with: `INFO: Shutting down / Waiting for application shutdown / Application shutdown complete / Finished server process [<pid>]`

So the Restart action **kills AC successfully but never re-spawns it**. Agents are reset and try to register against a dead port → 30s timeout → fall back to no-chat → 60s deferred wait also fails.

## Suspected root causes

These three symptoms compound from issues already noted as out-of-scope follow-ups in #577 — but together they make the chat unusable on a fresh install, so they're worth promoting to first-class bugs:

1. **No `waitForAgentChattrReady` after spawn (RC2 in #577).** Both `bin/quadwork.js` (around the `ok("AgentChattr started")` line) and `server/index.js`'s `spawnChattr` declare success the moment `spawn()` returns a PID. AC takes several seconds to actually bind 8300 (templates load, MCP servers start, FastAPI lifespan). The dashboard's history-snapshot fetch and the health monitor both check before that's done → 502 + false-down → auto-restart triggered unnecessarily. **A simple `await waitForAgentChattrReady(8300, 30000)` after spawn would prevent the entire cascade.**

2. **Auto-reset triggered from multiple paths.** `[agentchattr] auto-reset 4 agent(s)` firing three times for one logical AC restart suggests duplicate triggers — possibly the health-monitor restart endpoint's chained reset (`#447`) plus the startup migrations (`[bridge-migrate] restarted AC`, `[ghost-fix] patched registry.py`) all firing resets independently. Worth deduping.

3. **Restart endpoint's spawn-new step is failing or skipped.** The kill half works (`Finished server process` logged); the spawn-new half doesn't appear to run, OR it runs but the new AC dies silently. Since #574-576 added stdio capture, an AC startup failure should show in `agentchattr.log` — but the log just ends with shutdown messages, no new startup banner. So the spawn step likely never executed. Worth tracing the `/api/agentchattr/<id>/restart` handler in `server/routes.js` and verifying it actually calls `spawnChattr` after `killProcessOnPort`.

4. **Pin checkout silently failing (RC5 in #577, still not addressed).** AC clone HEAD is at `main` (currently `0440f5d`), not the pinned commit `3e71d42`. The `pinned` branch doesn't exist locally on this install. The console.warn at `bin/quadwork.js:272-273` is buried mid-spinner during install. Now that --config is fixed, HEAD works, so this isn't breaking — but it's worth surfacing post-spinner so users know they're not on the pinned version.

5. **Legacy `~/.quadwork/agentchattr/` clone still created** alongside the per-project clone (RC6 in #577). Cosmetic cruft, but confusing during diagnosis.

## Environment
- macOS 12.x (Darwin 21.6.0), Intel x86_64
- Node 24.15.0, Python 3.14.4 (python.org installer), gh 2.91.0, codex 0.125.0, claude 2.1.119
- Quadwork 1.13.1 (npx cache `~/.npm/_npx/<hash>/`)
- AC clone HEAD: `0440f5d` (`chore: bump version to 0.4.0`); pin `3e71d42` not checked out
- Fresh install: `~/.quadwork`, all `~/.npm/_npx/*/node_modules/quadwork`, and project worktrees all removed before `npx quadwork@latest start`

## Reproduction
1. `rm -rf ~/.quadwork ~/.npm/_npx/*/node_modules/quadwork && rm -rf <project>-{head,dev,re1,re2}` worktrees
2. `npx quadwork@latest start`
3. Dashboard → create project, configure agents, save
4. Watch server log → observe triple `auto-reset` and `agentchattr.log` showing two startup banners interleaved
5. In chat, send a single `@head ping dev, re1, re2` message → observe ghost-suffixed agents replying
6. Click **Restart** in dashboard server controls → observe AC log ends at `Finished server process` with no new startup banner; `lsof -iTCP:8300` empty; `[#565]` flood resumes

## Suggested fix order (smallest blast radius first)
1. Add `waitForAgentChattrReady` after every `spawn()` of AC (wizard init, server `spawnChattr`, restart endpoint). This single change probably eliminates Symptoms 1 and 2 by preventing the false-down detection that triggers the auto-reset cascade.
2. Audit the `/api/agentchattr/<id>/restart` endpoint to confirm it actually invokes `spawnChattr` after `killProcessOnPort` and `waitForPortFree`. (Symptom 3.)
3. Dedupe `auto-reset` triggers — at most one should fire per AC restart event.
4. Carry the secondary recommendations from #577 (loud pin failure, legacy clone cleanup).

realproject7:
## Summary
Fixes #579 — comprehensive startup hardening for the AC race condition that caused triple auto-reset cascades, ghost agent identities, and restart failures on fresh installs.

Builds on the acute fixes in #580 (waitForReady), #581 (reset dedup), and #582 (restart port-free) with five additional guards:

- **Detect already-running AC on server boot** — probes each project's AC port and populates `chattrProcesses`, so the health monitor tracks cmdStart-spawned ACs and the dashboard doesn't redundantly restart them
- **60s health monitor startup grace period** — skips health checks during the initial startup window so cmdStart spawns and startup migrations (bridge-migrate, ghost-fix) can settle
- **Dedupe auto-reset across rapid restart calls** — skips the reset if one was already scheduled or succeeded within the last 30s, preventing triple auto-reset when multiple restart sources fire in succession
- **Surface pin checkout warning after spawn** — clearly warns in cmdStart if AC clone drifted from the pinned commit, instead of burying the message mid-spinner during install
- **Skip legacy global clone on fresh installs** — no longer creates `~/.quadwork/agentchattr/` on fresh installs with no projects; AC is installed per-project by the web setup wizard

## Test plan
- [ ] Fresh install: verify no `~/.quadwork/agentchattr/` global clone created during `npx quadwork init`
- [ ] Fresh install: verify AC is installed per-project when creating first project via dashboard
- [ ] Existing install: verify server boot detects already-running AC (`[startup]` log line)
- [ ] Verify health monitor skips checks during first 60s (`60s startup grace` in log)
- [ ] Trigger multiple rapid AC restarts (bridge-migrate + dashboard) and verify only one auto-reset fires
- [ ] Verify pin drift warning appears in cmdStart output when AC clone is not on pinned commit
- [ ] `npm run build` passes

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -556,11 +556,18 @@ async function checkPrereqs(rl) {
   }
 
   // ── 3. AgentChattr (clone + venv — needs Python and git) ──
+  // #579: skip global install on fresh installs with no projects. Each
+  // project gets its own per-project clone via the web setup wizard, so
+  // a global ~/.quadwork/agentchattr/ clone is redundant cruft.
+  // Only check for an existing install to report status.
   const acDir = findAgentChattr();
+  const config = readConfig();
   if (acDir) {
     ok(`AgentChattr (${acDir})`);
     agentChattrFound = true;
-  } else if (hasPython) {
+  } else if (hasPython && config.projects && config.projects.length > 0) {
+    // Existing projects but no global install — install globally as
+    // fallback for legacy projects that don't have per-project clones.
     console.log("");
     warn("AgentChattr lets your AI agents communicate with each other.");
     log("It will be cloned and set up in a virtualenv.");
@@ -583,6 +590,11 @@ async function checkPrereqs(rl) {
       log(`  → Install later: git clone ${AGENTCHATTR_REPO} ${DEFAULT_AGENTCHATTR_DIR}`);
       allOk = false;
     }
+  } else if (hasPython) {
+    // Fresh install with no projects — AC will be installed per-project
+    // when the user creates their first project via the dashboard.
+    ok("AgentChattr will be installed per-project when you create a project.");
+    agentChattrFound = true;
   } else {
     warn("AgentChattr requires Python — install Python first, then re-run init.");
     allOk = false;
@@ -1671,6 +1683,17 @@ async function cmdStart() {
       }
       log(`  Log: ${acLogPath}`);
       acPids.push(acProc.pid);
+    }
+    // #579: verify pin checkout after spawn — surface loudly if AC clone
+    // drifted from the pinned commit, so operators know they're not on
+    // the tested version. The install-time warn() is often buried mid-
+    // spinner; this post-spawn check is always visible.
+    const headSha = (run("git", ["-C", projectAcDir, "rev-parse", "HEAD"]) || "").trim();
+    if (headSha && headSha !== AGENTCHATTR_PIN) {
+      warn(`AgentChattr for ${project.id} is NOT on the pinned commit.`);
+      log(`  Current: ${headSha.slice(0, 12)}`);
+      log(`  Pinned:  ${AGENTCHATTR_PIN.slice(0, 12)}`);
+      log(`  Run: git -C ${projectAcDir} checkout -B pinned ${AGENTCHATTR_PIN}`);
     }
   }
 
--- a/server/index.js
+++ b/server/index.js
@@ -1094,8 +1094,19 @@ async function handleAgentChattr(req, res) {
       // #447: auto-reset all agents after AC restart so they get
       // fresh MCP tokens. #581: mark reset as scheduled immediately
       // so the health monitor skips its own reset while ours is in-flight.
+      // #579: also skip if a reset already succeeded within the last 30s.
+      // Multiple restart sources (bridge-migrate, health monitor, dashboard)
+      // can fire in rapid succession — only the first should trigger a reset.
+      const existingReset = _acHealth.resetState.get(projectId);
+      const resetRecentlyDone = existingReset &&
+        (existingReset.status === "succeeded" || existingReset.status === "scheduled") &&
+        Date.now() - existingReset.timestamp < 30_000;
+      if (resetRecentlyDone) {
+        console.log(`[agentchattr] ${projectId} skipping auto-reset — one already ${existingReset.status} ${Math.round((Date.now() - existingReset.timestamp) / 1000)}s ago`);
+      } else {
       _acHealth.resetState.set(projectId, { status: "scheduled", timestamp: Date.now() });
-      setTimeout(async () => {
+      }
+      if (!resetRecentlyDone) setTimeout(async () => {
         try {
           const resetResp = await fetch(`http://127.0.0.1:${PORT}/api/agents/${encodeURIComponent(projectId)}/reset`, {
             method: "POST",
@@ -2152,6 +2163,10 @@ const _acHealth = {
   // #581: per-project reset state — prevents duplicate resets per restart event.
   // Values: { status: "scheduled"|"succeeded"|"failed", timestamp: number }
   resetState: new Map(),
+  // #579: timestamp when the health monitor started. The first 60s are a
+  // grace period: skip checks so cmdStart spawns and startup migrations
+  // (bridge-migrate, ghost-fix) can settle before the monitor acts.
+  startedAt: 0,
 };
 
 function isPortAlive(port) {
@@ -2166,6 +2181,10 @@ function isPortAlive(port) {
 }
 
 async function acHealthCheck() {
+  // #579: skip checks during the startup grace period so cmdStart spawns
+  // and startup migrations (bridge-migrate, ghost-fix) can settle.
+  if (_acHealth.startedAt && Date.now() - _acHealth.startedAt < 60_000) return;
+
   const cfg = readConfig();
   for (const project of (cfg.projects || [])) {
     const proc = chattrProcesses.get(project.id);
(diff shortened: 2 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new startup grace window is anchored to server boot, not to the moment a project's AgentChattr instance starts. That leaves the original race unfixed whenever the operator takes more than 60 seconds to finish setup before creating the first project.

### Findings
- [high] Health-check grace period expires before late project creation
  - File: `server/index.js:2183`
  - Suggestion: Track grace per project / per AC start or restart event, not as a single global `startedAt` timestamp set at server boot.
- [high] Global timer is initialized too early for the reported repro path
  - File: `server/index.js:2285`
  - Suggestion: Reset the grace window when QuadWork starts or restarts a given project's AC, otherwise a fresh-install user who spends >60s in the dashboard still hits the false-down auto-restart path.

### Decision
Requesting changes because the main race in issue #579 can still reproduce under normal fresh-install usage.

realproject7 (commented):
## RE2 Review — #589 (AC startup race hardening)

**Verdict: APPROVED**

Reviewed all five guards against the issue's root causes and the existing server code paths. The changes are sound and well-targeted.

### What works well

1. **60s startup grace period** (`_acHealth.startedAt` + early return in `acHealthCheck`): Cleanly prevents the false-down detection that triggers the entire cascade. Simpler than the issue's suggested `waitForAgentChattrReady` and doesn't block.

2. **30s reset dedupe** (`resetRecentlyDone` check): The triple-fire from concurrent restart sources (bridge-migrate, health monitor, dashboard) is now gated — only the first reset within 30s fires. This directly addresses Symptom 2 (ghost identity proliferation from repeated register/deregister churn).

3. **Boot-time AC detection** (port probing in `server.listen` callback): Correctly populates `chattrProcesses` with `{ process: null, state: "running" }` for cmdStart-spawned ACs. Verified that all kill paths (`stop`, `restart`, `shutdown`) guard on `proc.process` before `.kill()` and fall back to `killProcessOnPort`, so the `null` sentinel is safe.

4. **Pin checkout warning post-spawn** and **legacy clone skip**: Both clean quality-of-life improvements. The conditional `config.projects.length > 0` gate on global install is a good heuristic.

### Notes (non-blocking)

- **RC3 (Symptom 3 — restart leaves AC dead)**: The restart handler at `index.js:986` does call `spawnChattr()` after kill, so the code path itself is correct. The issue's Symptom 3 was likely caused by the cascade interfering with the spawn (triple-reset flooding the freshly-started AC). The dedupe guard should prevent that, but worth monitoring in the next fresh-install test.

- **Dedupe structure** (cosmetic): The `if/else` setting `resetState` followed by a separate `if (!resetRecentlyDone) setTimeout(...)` is slightly tangled. A single guard clause would read cleaner, but functionally correct as-is.

- **`server.listen` callback is now `async`**: Fine for Node/Express — the callback isn't awaited by the caller, and the `await isPortAlive()` inside is correctly used.

LGTM — the five guards form good defense-in-depth against the startup race.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit fixes the blocking issue from the prior review by moving the 60s health-monitor grace from a global server-boot timer to a per-project `runningSince` timestamp. That matches the reported fresh-install repro path and keeps late project creation from falling back into the false-down restart cascade.

### Findings
- None.

### Decision
Approving because the prior regression concern is resolved and the rest of the hardening in this PR remains sound.

PR #595, opened 2026-04-28
[#593] Harden AC pin checkout with retry + hard failure

### LINKED ISSUE #593: Harden AC pin checkout — retry with explicit fetch, fail on persistent failure
## Context
QuadWork pins AgentChattr to commit `3e71d42` via `git checkout -B pinned <sha>` after cloning. When this checkout fails (timeout, network issue, commit not fetched), AC stays on `main` HEAD which has breaking API changes. This has caused cascading bugs on every fresh install where the pin fails (#577, #579, #586, #588).

The pin checkout works perfectly when tested manually — the commit is reachable in the repo. The failures are transient (timeouts, git fetch race) but the current error handling silently continues with HEAD, leading to hours of debugging.

## Current behavior (`bin/quadwork.js:271-281`)
```js
const pinResult = run("git", ["-C", dir, "checkout", "-B", "pinned", AGENTCHATTR_PIN], { timeout: 30000 });
if (pinResult === null) {
  warn(pinWarnMsg);  // buried in output, easy to miss
  // continues with HEAD — all downstream bugs follow
}
```

## Proposed fix
Make the pin checkout robust with retry + explicit fetch, and fail loudly if it can't be resolved:

```js
// Attempt 1: direct checkout (works if commit is in the clone)
let pinResult = run("git", ["-C", dir, "checkout", "-B", "pinned", AGENTCHATTR_PIN], { timeout: 30000 });

if (pinResult === null) {
  // Attempt 2: explicit fetch of the exact commit, then retry checkout
  log("Pin checkout failed — fetching commit explicitly...");
  run("git", ["-C", dir, "fetch", "origin", AGENTCHATTR_PIN], { timeout: 60000 });
  pinResult = run("git", ["-C", dir, "checkout", "-B", "pinned", AGENTCHATTR_PIN], { timeout: 30000 });
}

if (pinResult === null) {
  // Both attempts failed — abort install instead of continuing with HEAD
  return setError(
    `Could not check out AgentChattr pin ${AGENTCHATTR_PIN.slice(0, 12)}. ` +
    `This means AC would run an untested version with known incompatibilities. ` +
    `Check your network connection and retry, or manually run:\n` +
    `  git -C ${dir} fetch origin ${AGENTCHATTR_PIN}\n` +
    `  git -C ${dir} checkout -B pinned ${AGENTCHATTR_PIN}`
  );
}
```

## Key changes
1. **Retry with explicit fetch** — `git fetch origin <sha>` ensures the commit is downloaded even if the initial clone didn't include it
2. **Fail the install** on persistent failure — no more silent continuation with HEAD
3. **Increase fetch timeout** to 60s — some VPS/network environments are slow
4. **Clear error message** with manual fix instructions

## Safety
- On fast networks (this Mac), attempt 1 succeeds instantly — no change
- On slow networks, attempt 2 adds one fetch + retry — ~30s extra at most
- If both fail, install stops cleanly instead of producing a broken AC installation
- Existing installs are unaffected (pin checkout only runs on fresh clone)

## Acceptance criteria
- [ ] Pin checkout retries with explicit fetch on first failure
- [ ] Install aborts with clear error if both attempts fail
- [ ] Fresh install on fast network sees no delay
- [ ] Fresh install on slow network succeeds after retry
- [ ] `npm run build` passes

Related: #577 (original pin failure discovery), #579, #588

realproject7:
## Summary
- Pin checkout now retries with explicit `git fetch origin <sha>` when the initial attempt fails
- Install aborts with clear error + manual fix instructions if both attempts fail — no more silent continuation on HEAD AC
- Fetch timeout increased to 60s for slow VPS/network environments

## Changes
- **`bin/quadwork.js`**: Replaced single-attempt pin checkout + silent fallback with retry logic and hard failure

## Safety
- Fast networks: attempt 1 succeeds instantly, no behavior change
- Slow networks: adds one fetch + retry (~30s extra at most)
- Persistent failure: install stops cleanly instead of producing a broken AC installation
- Existing installs unaffected (pin checkout only runs on fresh clone)
- `npm run build` passes

## Test plan
- [ ] Fresh install on fast network succeeds without delay
- [ ] Simulated first-attempt failure recovers via explicit fetch
- [ ] Both attempts failing produces clear error with manual fix commands
- [ ] Existing AC installs are unaffected (code path not reached)

Fixes #593

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -268,16 +268,21 @@ function _installAgentChattrLocked(dir, setError) {
     // On failure (commit unreachable / force-pushed away), fall
     // back to the default branch with a loud warning instead of
     // hard-failing the install.
-    const pinResult = run("git", ["-C", dir, "checkout", "-B", "pinned", AGENTCHATTR_PIN], { timeout: 30000 });
+    // #593: Retry with explicit fetch on failure, abort install on persistent failure.
+    let pinResult = run("git", ["-C", dir, "checkout", "-B", "pinned", AGENTCHATTR_PIN], { timeout: 30000 });
     if (pinResult === null) {
-      const pinWarnMsg = `Could not check out AgentChattr pin ${AGENTCHATTR_PIN} at ${dir}; falling back to default branch. AC's argparse may reject unknown flags. Update AGENTCHATTR_PIN in bin/quadwork.js for reproducible installs.`;
-      warn(pinWarnMsg);
-      // Persist to install log so post-mortem diagnosis is possible
-      // without re-running the install.
-      try {
-        const logDir = path.join(path.dirname(dir));
-        fs.appendFileSync(path.join(logDir, "install.log"), `[${new Date().toISOString()}] PIN-CHECKOUT-FAIL: ${pinWarnMsg}\n`);
-      } catch {}
+      log("Pin checkout failed — fetching commit explicitly...");
+      run("git", ["-C", dir, "fetch", "origin", AGENTCHATTR_PIN], { timeout: 60000 });
+      pinResult = run("git", ["-C", dir, "checkout", "-B", "pinned", AGENTCHATTR_PIN], { timeout: 30000 });
+    }
+    if (pinResult === null) {
+      return setError(
+        `Could not check out AgentChattr pin ${AGENTCHATTR_PIN.slice(0, 12)} at ${dir}. ` +
+        `AC would run an untested HEAD version with known incompatibilities. ` +
+        `Check your network connection and retry, or manually run:\n` +
+        `  git -C ${dir} fetch origin ${AGENTCHATTR_PIN}\n` +
+        `  git -C ${dir} checkout -B pinned ${AGENTCHATTR_PIN}`
+      );
     }
   } else {
     // #366: existing clone from a pre-fix install. If the clone


realproject7 (commented):
## RE2 Review — APPROVED

**Scope check:** PR matches #593 acceptance criteria — retry with explicit fetch, hard failure on persistent checkout errors, no silent HEAD fallback.

**Logic verified:**

1. **Attempt 1:** `git checkout -B pinned <sha>` with 30s timeout — unchanged from before ✅
2. **Attempt 2 (on failure):** `git fetch origin <sha>` with 60s timeout, then retry checkout — covers the case where the commit wasn't included in the initial clone ✅
3. **Hard failure:** `setError()` aborts install with truncated SHA, explanation, and manual fix commands — no more silent continuation on HEAD ✅

**Correctness:**
- `run()` returns `null` on any failure (line 80-81), so both null checks are sound
- The `else` branch (existing clone migration, line 282+) is correctly untouched — retry logic only applies to fresh clones
- `setError()` returns early, preventing any further execution in the broken state
- Error message includes the exact manual fix commands, which is good for VPS debugging

**Safety:**
- Fast path: attempt 1 succeeds → zero overhead
- Slow path: one extra fetch + retry → ~90s max before abort
- Removed the install.log write — acceptable since the error message is now actionable and the install stops

LGTM.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The retry/fail-fast behavior is the right direction, but the new implementation does not actually invoke `git` correctly in this codebase.

### Findings
- [high] The new checkout/fetch calls use `run("git", [..args..], { timeout })`, but `run()` only accepts a single shell-command string plus one options object. In `bin/quadwork.js:54`, `run(cmd, opts = {})` forwards `cmd` directly into `execSync(cmd, ...)`. That means the new calls in the pin-checkout block pass the argument array as the `opts` object, so the intended `git checkout` / `git fetch` commands are never constructed. The retry path therefore cannot work as written.
  - File: `bin/quadwork.js:54`
  - Suggestion: keep using the existing string-based `run(...)` API here, e.g. `run(`git -C "${dir}" checkout -B pinned ${AGENTCHATTR_PIN} 2>&1`, { timeout: 30000 })` and the analogous fetch command, or refactor `run()` itself to support argv-style invocation before using it.

### Decision
Requesting changes because the new hardening path is currently broken at the call site and would not reliably pin AgentChattr.

project7-interns (approved):
## Verdict: APPROVE

### Summary
Re-reviewed against the actual PR head ref and the change matches issue #593: pin checkout now retries with an explicit fetch and aborts cleanly on persistent failure instead of silently continuing on HEAD.

### Findings
- None.

### Decision
Approved. My prior change request was based on a stale local workspace copy. The PR branch itself already has `bin/quadwork.js:77` using `execFileSync` via `run(cmd, args = [], opts = {})`, so the new argv-style `git` calls are valid on the reviewed ref.

PR #623, opened 2026-04-29
Add nginx auth cookie caching to VPS guide

### LINKED ISSUE #622: Add nginx auth cookie caching to VPS guide to reduce sign-in popups
## Problem
On mobile (especially Safari), HTTP basic auth reprompts every 2-3 minutes when switching projects or when WebSocket connections reconnect. This is because Safari drops the Authorization header across new connections more aggressively than desktop browsers.

## Fix
Add an nginx cookie-based auth cache configuration to `docs/install-vps.md` in the HTTP Basic Auth section. This caches successful authentication in a cookie for 24h, so nginx skips the basic auth challenge on subsequent requests.

No QuadWork code changes needed — nginx-only configuration.

## Configuration to add

Update the nginx server block example in `docs/install-vps.md`:

```nginx
# Cache basic auth in a cookie so mobile browsers don't reprompt
map $cookie_qw_auth $auth_ok {
    "authenticated" "off";
    default         "QuadWork";
}

server {
    listen 443 ssl;
    server_name app.example.com;

    auth_basic $auth_ok;
    auth_basic_user_file /etc/nginx/.htpasswd;

    # Set cookie after successful auth (24h expiry)
    add_header Set-Cookie "qw_auth=authenticated; Path=/; Max-Age=86400; HttpOnly; Secure" always;

    location / {
        proxy_pass http://127.0.0.1:3000;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_read_timeout 86400;
    }
}
```

Note: the `map` block goes outside the `server` block (at the `http` level in nginx config).

## Where to add in docs
In `docs/install-vps.md`, replace the existing nginx + basic auth example with this updated version. Add a note explaining:
- First visit: browser prompts for credentials (normal basic auth)
- Cookie is set for 24h after successful login
- Subsequent requests use the cookie — no reprompt
- To force re-login: clear the `qw_auth` cookie or wait 24h

## Acceptance criteria
- [ ] VPS install guide updated with cookie-caching nginx config
- [ ] `map` block placement explained (http level, not server level)
- [ ] Note about cookie expiry and forced re-login

realproject7:
## Summary
- Adds cookie-based auth cache configuration to `docs/install-vps.md` Step 10
- Uses nginx `map` block to check `qw_auth` cookie and skip basic auth when present
- Documents `map` block placement (http level, not server level) and cookie expiry/forced re-login

Fixes #622

## Test plan
- [ ] Verify `map` block placement explanation is clear (http level vs server level)
- [ ] Confirm cookie expiry behavior documented (24h, clear cookie to force re-login)
- [ ] Check nginx config syntax is valid (`nginx -t` passes with this config)

### DIFF
--- a/docs/install-vps.md
+++ b/docs/install-vps.md
@@ -279,17 +279,25 @@ sudo htpasswd -cb /etc/nginx/.htpasswd admin 'YOUR_GENERATED_PASSWORD'
 
 Mobile browsers (especially Safari) drop the `Authorization` header aggressively on new connections and WebSocket reconnects, causing repeated sign-in popups every few minutes. The fix: cache successful auth in a cookie so nginx skips the challenge on subsequent requests.
 
-Add a `map` block **outside** the `server` block (at the `http` level — typically at the top of your site config file or in `/etc/nginx/conf.d/auth-cache.conf`):
+Generate a unique secret for your deployment (this becomes the cookie value):
+
+```bash
+openssl rand -hex 16
+# Example output: a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6
+# Save this — you'll use it in the nginx config below
+```
+
+Add a `map` block **outside** the `server` block (at the `http` level — typically at the top of your site config file or in `/etc/nginx/conf.d/auth-cache.conf`). Replace `YOUR_AUTH_SECRET` with the value generated above:
 
 ```nginx
 # Cache basic auth in a cookie so mobile browsers don't reprompt
 map $cookie_qw_auth $auth_ok {
-    "authenticated" "off";
-    default         "QuadWork";
+    "YOUR_AUTH_SECRET" "off";
+    default            "QuadWork";
 }
 ```
 
-Then update the `server` block (inside `listen 443 ssl`) to use `$auth_ok` instead of a static string, and set the cookie on every response:
+Then update the `server` block (inside `listen 443 ssl`) to use `$auth_ok` instead of a static string, and set the cookie on every response. Replace `YOUR_AUTH_SECRET` with the same value:
 
 ```nginx
 auth_basic $auth_ok;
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The doc update matches the issue scope, but the nginx snippet as written creates an auth bypass because it marks the client as authenticated before successful basic auth has been established.

### Findings
- [high] The example sets `Set-Cookie: qw_auth=authenticated` with `add_header ... always`, which also applies to the initial `401 Unauthorized` challenge response. A client can therefore receive the auth-bypass cookie before presenting valid credentials, and the next request will hit `auth_basic off` through the `map` block. That turns the documented config into an authentication bypass, not just a cache.
  - File: `docs/install-vps.md:291`
  - Suggestion: Do not set the bypass cookie unconditionally on all responses. Either document a safe, verified pattern that only issues the cookie after successful authentication, or remove this config from the guide until that condition is enforceable.

### Decision
Requesting changes because the current snippet weakens access control and should not be published as a copy-pasteable setup.

realproject7:
## RE2 Review — REQUEST CHANGES

### [high] Static cookie value bypasses auth for anyone who reads this guide

The cookie check is:
```nginx
map $cookie_qw_auth $auth_ok {
    "authenticated" "off";
    ...
}
```

The value `"authenticated"` is a hardcoded, publicly-documented string. Anyone who reads this guide (public repo) can set `qw_auth=authenticated` in their browser DevTools and bypass basic auth entirely. `HttpOnly` prevents JS access but doesn't prevent manual cookie injection.

**Fix:** Instruct the user to generate a random secret and use that as the map value:

```bash
# Generate once, save somewhere safe
openssl rand -hex 16
# e.g. produces: a3f8b2c1d4e5f60718293a4b5c6d7e8f
```

```nginx
map $cookie_qw_auth $auth_ok {
    "a3f8b2c1d4e5f60718293a4b5c6d7e8f" "off";
    default "QuadWork";
}
```

```nginx
add_header Set-Cookie "qw_auth=a3f8b2c1d4e5f60718293a4b5c6d7e8f; Path=/; Max-Age=86400; HttpOnly; Secure" always;
```

This keeps the same mechanism but the bypass value is unique per deployment.

### [low] `always` flag sets cookie on 401 responses

`add_header ... always` adds the header to ALL responses including 401 (failed auth). While most browsers won't persist cookies from cancelled auth dialogs, it's cleaner to omit `always` here so the cookie is only set on successful (2xx/3xx) responses.

---

Everything else looks good — map block placement explanation, cookie flow description, and forced re-login instructions are clear.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The `401`-response problem is fixed, but the documented nginx config still allows trivial auth bypass because the cookie value is a known static string.

### Findings
- [high] The guide tells every deployment to trust the literal cookie value `qw_auth=authenticated`. Because that value is now documented and constant, any user can set that cookie manually in browser devtools and immediately satisfy the `map`, which disables basic auth without presenting credentials.
  - File: `docs/install-vps.md:284`
  - Suggestion: Use a deployment-specific secret value instead of a public constant, and document generating it explicitly (for example with `openssl rand -hex 16`) so the cookie cannot be forged trivially.

### Decision
Requesting changes because the example still weakens authentication and should not be published in a copy-paste guide until the bypass is removed.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up revisions resolve the auth-bypass issues in the original draft. The guide now documents a deployment-specific secret for the cookie value and correctly avoids setting the cookie on `401` responses.

### Findings
- None.

### Decision
Approving because the VPS guide now meets issue #622 without documenting a trivial authentication bypass.

realproject7:
## RE2 Re-review — APPROVED

Both issues resolved:
- [high] Cookie value is now a per-deployment random secret (generated via `openssl rand -hex 16`) instead of a static string. Auth bypass no longer possible from reading the guide.
- [low] `always` removed from `add_header` — cookie only set on successful responses.

Guide is clear: secret generation step, placeholder substitution in both map and Set-Cookie, security note explaining why it's safe. LGTM.

PR #628, opened 2026-04-29
Fix bridge-migrate: only restart AC on slug rename

### LINKED ISSUE #616: bridge-migrate restarts AC unnecessarily after script-only patches
## Bug
The `bridge-migrate` startup migration restarts AC even when only the bridge Python script was patched (not config.toml). This kills the Discord/Telegram bridge registration, causing bridges to stop after every QuadWork restart.

## Server log
```
[bridge-refresh] refreshed Discord bridge script from package
[bridge-migrate] patched stale bridge_sender in discord_bridge.py
[bridge-migrate] plotlink: restarted AC
```

The script refresh and `bridge_sender` patch are file-level changes to the bridge Python code — AC doesn't need to restart for those. Only config.toml slug changes (`[agents.discord-bridge]` → `[agents.dc]`) require an AC restart.

## Impact
On VPS with pm2, every QuadWork restart (including pm2 auto-restart, server reboot) triggers this migration → AC restarts → bridge loses registration → bridge stops. Operator has to manually restart the bridge each time.

## Root cause
In `server/index.js`, the bridge-migrate block restarts AC whenever ANY migration step ran, not just when config.toml changed. The config.toml change is already tracked via `dc.changed || tg.changed`, but the AC restart fires regardless.

## Fix
Only restart AC when config.toml was actually modified. The bridge script patches and `bridge_sender` fixes should NOT trigger an AC restart.

Check the bridge-migrate section in `server/index.js` (around line 2330-2356). The AC restart (`setTimeout → fetch restart`) should be gated behind the existing `dc.changed || tg.changed` condition — which it likely already is for config.toml, but other patches (script refresh, bridge_sender) may be triggering separate restarts.

## Safety
- This Mac: migration already runs at every boot. With the fix, it'll just skip the unnecessary restart. No behavior change.
- New installs: no config.toml slug migration needed (fresh configs have correct slugs). AC restart skipped. No behavior change.
- VPS: bridges survive QuadWork restarts. Improvement.

## Acceptance criteria
- [ ] AC only restarts when config.toml slug migration actually changed the file
- [ ] Bridge script patches do NOT trigger AC restart
- [ ] Bridges stay running across QuadWork restarts
- [ ] `npm run build` passes

realproject7:
## Summary
- Bridge-migrate startup migration was restarting AC whenever config.toml changed, including fresh `[agents.dc]`/`[agents.tg]` block appends
- This killed bridge registration on every QuadWork restart (pm2, reboot)
- Now tracks whether an actual slug RENAME happened (`discord-bridge`→`dc`, `telegram-bridge`→`tg`) and only restarts AC in that case
- Fresh block appends (for projects that never had the old slug) write config but skip the restart — AC picks up new blocks on its next natural start

Fixes #616

## Test plan
- [ ] Project with old slugs: config renamed + AC restarted (existing behavior preserved)
- [ ] Project without bridge: block appended but NO AC restart
- [ ] Bridges survive QuadWork restart (no unnecessary AC restart)
- [ ] `npm run build` passes ✅

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -2330,29 +2330,39 @@ server.listen(PORT, "127.0.0.1", async () => {
   // #457: migrate bridge slugs in AC configs on startup.
   // Renames [agents.discord-bridge] → [agents.dc] and
   // [agents.telegram-bridge] → [agents.tg] so bridges register
-  // under the short slug. Restarts AC for projects whose config changed.
+  // under the short slug. Restarts AC ONLY for slug renames (not
+  // fresh block appends) — #616: script-only patches should not
+  // trigger AC restarts which kill bridge registration.
   for (const p of (startupCfg.projects || [])) {
     const acPath = projectAgentchattrConfigPath(p.id);
     if (!fs.existsSync(acPath)) continue;
     try {
       const before = fs.readFileSync(acPath, "utf-8");
+      // Track whether an actual slug RENAME happened (old → new).
+      // Fresh block appends don't need an AC restart — AC picks them
+      // up on its next natural start.
+      const hadOldDc = /^\[agents\.discord-bridge\]\s*$/m.test(before);
+      const hadOldTg = /^\[agents\.telegram-bridge\]\s*$/m.test(before);
       const dc = patchAgentchattrConfigForDiscordBridge(before);
       const tg = patchAgentchattrConfigForTelegramBridge(dc.text);
       if (dc.changed || tg.changed) {
         fs.writeFileSync(acPath, tg.text);
         console.log(`[bridge-migrate] ${p.id}: migrated AC config slugs`);
-        // Restart AC so it loads the new agent slugs
-        setTimeout(async () => {
-          try {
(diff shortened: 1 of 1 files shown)


realproject7:
## RE2 Review — APPROVED

Correct fix for the restart loop:
- Captures whether old slugs (`discord-bridge`, `telegram-bridge`) existed BEFORE patching
- Config is still written regardless (fresh blocks still appended)
- AC restart only fires when `hadOldDc || hadOldTg` — meaning an actual rename happened
- Regex patterns correctly match TOML section headers with multiline flag

This breaks the cycle: boot → append fresh block → no restart → AC picks up naturally. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The startup migration now distinguishes actual slug renames from fresh bridge block appends and only restarts AC for the rename case. That matches issue #616 and removes the unnecessary restart cycle that was breaking bridge registration after QuadWork restarts.

### Findings
- None.

### Decision
Approving because the patch is focused, preserves the intended rename behavior, and eliminates the restart path that was harming bridge stability.

Issue #629, opened 2026-04-30
[idle-fix] runtime patch never reaches AC's running Python: _CRASH_TIMEOUT stays at 15s in memory while on-disk is 120, agents deregistered while alive


realproject7:
# `[idle-fix]` runtime patch never reaches AC's running Python process — `_CRASH_TIMEOUT` stays at 15s, agents get deregistered while alive

## TL;DR
Quadwork's `[idle-fix] (#502)` startup migration edits `_CRASH_TIMEOUT = 15` → `120` in AC's `app.py` *on disk*, but AC's Python process imports the module once at startup and caches the value in memory. The on-disk patch and the in-memory value diverge, and AC keeps using 15s. Combined with macOS sleep/App Nap stretching Node's 5-second heartbeat interval past 15s, agents get deregistered while their underlying CLI processes are still alive — surfacing as recurring "agents appear dead" symptoms, ghost-identity cascades, and the cross-cutting evidence in #577 / #579 / #598 / #606.

The contradiction is observable:
- `app.py` on disk: `_CRASH_TIMEOUT = 120` ✅ (patch applied to file)
- AC log: `Crash timeout: deregistering X (no heartbeat for 15s)` ❌ (running process still using 15s)

## Smoking gun

**The log message format is** (`agentchattr/app.py:369`):
```python
log.info(f"Crash timeout: deregistering {name} (no heartbeat for {_CRASH_TIMEOUT}s)")
```

So the literal "15s" or "120s" in any crash-timeout log line directly reflects what value `_CRASH_TIMEOUT` had **in memory** when that line was logged.

**Across three projects' AC logs on a single machine, 14 of 14 crash-timeout entries say `15s`. None say `120s`:**

```
$ grep 'no heartbeat for' ~/.quadwork/*/agentchattr.log | sed 's/.*\(no heartbeat for [0-9]*s\).*/\1/' | sort | uniq -c
  14 no heartbeat for 15s
```

Meanwhile, the source file shows the patch did write to disk:
```
$ grep -nE '_CRASH_TIMEOUT' ~/.quadwork/*/agentchattr/app.py
…/app.py:363:                _CRASH_TIMEOUT = 120
…/app.py:368:                    if last_seen > 0 and now - last_seen > _CRASH_TIMEOUT:
…/app.py:369:                        log.info(f"... no heartbeat for {_CRASH_TIMEOUT}s)")
```

[...]

PR #636, opened 2026-04-30
[#629] Move crash timeout patch to clone time and restart AC for existing installs

realproject7:
## Summary
- Moved `_CRASH_TIMEOUT` patch from `server/index.js` startup migration into `server/install-agentchattr.js` (`patchCrashTimeout()`) so it runs at clone time — before AC's first `python run.py` import. Fresh installs now get `_CRASH_TIMEOUT = 120` in memory from the start.
- For existing installs where the startup migration patches `app.py` on a running AC instance, the migration now flags the project and triggers an AC restart so the running Python process picks up the new value.
- Added focused tests for `patchCrashTimeout()` covering: patch application, idempotency, missing `app.py`, and null/undefined dir.

## Test plan
- [ ] Fresh install: verify `_CRASH_TIMEOUT = 120` in `app.py` before AC starts, confirm AC log shows `120s` (not `15s`) in any crash-timeout entries
- [ ] Existing install with `_CRASH_TIMEOUT = 15` still on disk: verify startup migration patches and restarts AC, confirm log shows restart message
- [ ] Existing install already patched (`_CRASH_TIMEOUT = 120`): verify no restart triggered, no-op behavior
- [ ] Run `node server/install-agentchattr.patchCrashTimeout.test.js` — all assertions pass
- [ ] Leave machine idle overnight, check `agentchattr.log` for crash-timeout entries showing `120s`

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -320,9 +320,31 @@ function _installAgentChattrLocked(dir, setError) {
       if (pipResult === null) return setError(`pip install -r ${reqFile} failed`);
     }
   }
+  // #629: patch crash timeout before AC's first import
+  _patchCrashTimeout(dir);
   return dir;
 }
 installAgentChattr.lastError = null;
+
+function _patchCrashTimeout(dir) {
+  if (!dir) return;
+  const appPath = path.join(dir, "app.py");
+  if (!fs.existsSync(appPath)) return;
+  try {
+    let app = fs.readFileSync(appPath, "utf-8");
+    if (app.includes("_CRASH_TIMEOUT = 15")) {
+      app = app.replace("_CRASH_TIMEOUT = 15", "_CRASH_TIMEOUT = 120");
+      app = app.replace(
+        "# Crash timeout: if a wrapper hasn't heartbeated for 60s,\n",
+        "# Crash timeout: if a wrapper hasn't heartbeated for 120s,\n",
+      );
+      fs.writeFileSync(appPath, app);
+      log("[idle-fix] patched crash timeout to 120s at clone time (#629)");
+    }
+  } catch (err) {
+    try { console.warn(`[idle-fix] failed to patch crash timeout: ${err.message}`); } catch {}
+  }
+}
 
 /**
  * Get spawn args for launching AgentChattr from its cloned directory.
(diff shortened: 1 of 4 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR patches the server helper path, but it does not cover the main CLI install path used by fresh project setup, and the existing-install restart path skips already-running AC processes discovered at server startup. That leaves the exact 15s-in-memory failure unresolved for common flows.

### Findings
- [high] Fresh CLI installs still start AC from an unpatched `app.py`.
  - File: `bin/quadwork.js:323`
  - Details: `bin/quadwork.js` has its own `installAgentChattr()` implementation used by `cmdInit`/project setup/migration (`installAgentChattr(perProjectDir)`), and `_installAgentChattrLocked()` returns without applying `patchCrashTimeout()`. The issue acceptance requires fresh installs to get `_CRASH_TIMEOUT = 120` before AC first starts, but the CLI path still clones/checks out AgentChattr and can then spawn it with `_CRASH_TIMEOUT = 15`.
  - Suggestion: share the server installer helper or duplicate the crash-timeout patch in the CLI installer immediately after clone/checkout and before any AC spawn.

- [high] Existing running AC processes tracked on startup are never restarted after the patch.
  - File: `server/index.js:2567`
  - Details: startup discovery records already-running AC as `{ process: null, state: "running" }` because the server does not own the child. The new restart loop then does `if (!proc || !proc.process) continue`, so it skips exactly those existing processes. Their on-disk `app.py` is patched, but the running Python process keeps the old imported 15s value.
  - Suggestion: trigger the restart endpoint for every project marked `_acRestartNeeded`; the endpoint already kills orphaned processes on the port, so it can handle `process: null` cases.

- [medium] Updating AgentChattr can revert the patch and immediately restart into 15s.
  - File: `server/index.js:1171`
  - Details: the update flow runs `git pull`, reapplies only the CSS patch, and then restarts AC if it was running. If `git pull` rewrites `app.py` back to `_CRASH_TIMEOUT = 15`, the restarted process imports the unpatched value until a later QuadWork startup migration runs.
  - Suggestion: call `patchCrashTimeout(acDir)` after `git pull` and before any update-triggered restart.

### Decision
Requesting changes because the PR does not satisfy the fresh-install acceptance criterion and still leaves existing running AC processes using the stale in-memory timeout in a common startup path.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The rework addresses the RE1 blockers: the CLI install path now patches `_CRASH_TIMEOUT` before AC can first import it, the existing-install startup restart path handles already-running AC instances tracked with `process: null`, and the update path reapplies the patch after `git pull` before restart.

### Findings
- None blocking.

### Decision
Approved. No CI checks are reported on this PR branch, but the reviewed changes now satisfy the issue acceptance criteria for fresh installs, existing running installs, and post-update restarts.

Issue #631, opened 2026-04-30
[#630-1] Butler agent backend — PTY session + API endpoints


realproject7:
## Summary
Spawn a Butler agent as a Claude Code PTY session (project-agnostic, runs in `~/docs/`). Butler is a direct PTY session like project agents but without AgentChattr routing — no heartbeat, no registration, no AC dependency.

### `POST /api/butler/start`
- Spawns a Claude Code PTY in `~/docs/` working directory
- Passes `--dangerously-skip-permissions` if auto-approve configured
- Loads the Butler AGENTS.md seed (from Sub-4, #630-4) as context
- Returns `{ ok, pid }`
- No-op if already running

### `POST /api/butler/stop`
- Kills the Butler PTY process
- Returns `{ ok }`

### `GET /api/butler/status`
- Returns `{ running, pid }`

### WebSocket: `/ws/butler`
- Streams PTY I/O to the frontend (reuse existing xterm.js WebSocket pattern from agent terminals)
- Bidirectional: operator types → PTY stdin, PTY stdout → browser

## Implementation
- Add `butlerSession` to `server/index.js` (similar to `agentSessions` but single instance, not per-project)
- Reuse `pty.spawn()` pattern from `spawnAgentPty`
- Butler PTY runs the configured CLI command (default: `claude`)
- Working directory: `~/docs/` (create if missing)
- Auto-start on server boot if `config.butler.enabled` is true

## What Butler is NOT
- NOT an AgentChattr agent — no AC registration, no heartbeat, no slot management
- NOT project-specific — operates across all projects
- NOT spawned per-project — single global instance

## Acceptance criteria
- [ ] `POST /api/butler/start` spawns Claude Code PTY in `~/docs/`
- [ ] `POST /api/butler/stop` kills the PTY
- [ ] `GET /api/butler/status` returns running state
- [ ] WebSocket `/ws/butler` streams PTY I/O
- [ ] Auto-start on boot if configured
- [ ] `npm run build` passes

realproject7:
Fixed by merged PR #638.

PR #638, opened 2026-04-30
[#631] Butler agent backend — PTY session + API endpoints

realproject7:
## Summary
- Added Butler agent backend as a single global PTY session running Claude Code in `~/docs/`
- No AgentChattr integration — Butler operates independently across all projects
- Three API endpoints: `POST /api/butler/start`, `POST /api/butler/stop`, `GET /api/butler/status`
- WebSocket `/ws/butler` for bidirectional PTY I/O with 64KB scrollback buffer
- Auto-seeds `~/docs/AGENTS.md` from `templates/seeds/butler.AGENTS.md` on first start
- Auto-start on boot if `config.butler.enabled` is true
- Clean shutdown integrated into `shutdownChattrProcesses()`

## Test plan
- [ ] `POST /api/butler/start` spawns Claude Code PTY in `~/docs/`
- [ ] `POST /api/butler/stop` kills the PTY
- [ ] `GET /api/butler/status` returns `{ running: true/false, pid }` correctly
- [ ] WebSocket `/ws/butler` streams PTY I/O bidirectionally
- [ ] Scrollback replay works on reconnect
- [ ] Auto-start works when `config.butler.enabled` is true
- [ ] `~/docs/AGENTS.md` seeded from template on first start (not overwritten if exists)
- [ ] `npm run build` passes

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1394,26 +1394,26 @@ app.post("/api/agents/:project/:agent/write", (req, res) => {
 function spawnButlerPty() {
   if (butlerSession.term) return { ok: true, pid: butlerSession.term.pid };
 
-  const docsDir = path.join(os.homedir(), "docs");
-  if (!fs.existsSync(docsDir)) {
-    fs.mkdirSync(docsDir, { recursive: true, mode: 0o700 });
-  }
-
-  const cfg = readConfig();
-  const butlerCfg = cfg.butler || {};
-  const command = butlerCfg.command || "claude";
-  const args = [];
-  if (butlerCfg.auto_approve) args.push("--dangerously-skip-permissions");
-
-  const seedPath = path.join(__dirname, "..", "templates", "seeds", "butler.AGENTS.md");
-  if (fs.existsSync(seedPath)) {
-    const agentsPath = path.join(docsDir, "AGENTS.md");
-    if (!fs.existsSync(agentsPath)) {
-      fs.copyFileSync(seedPath, agentsPath);
-    }
-  }
-
   try {
+    const docsDir = path.join(os.homedir(), "docs");
+    if (!fs.existsSync(docsDir)) {
+      fs.mkdirSync(docsDir, { recursive: true, mode: 0o700 });
+    }
+
+    const cfg = readConfig();
+    const butlerCfg = cfg.butler || {};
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The Butler PTY/WebSocket shape mostly matches the requested backend feature, but the start path has unhandled filesystem failures before its own error handling, and the PR has no passing build signal despite the issue requiring one.

### Findings
- [high] Butler start can throw before returning a controlled `{ ok: false, error }` result.
  - File: `server/index.js:1398`
  - Details: `spawnButlerPty()` does `fs.mkdirSync(docsDir)`, `readConfig()`, and `fs.copyFileSync(seedPath, agentsPath)` before entering the `try` block that returns `{ ok: false, error }`. That means permission errors in `~/docs`, config read failures, or seed-copy failures escape the Butler lifecycle helper. From `/api/butler/start` this bypasses the endpoint's JSON response contract; from `/ws/butler` it can escape the connection handler; from boot auto-start it can reject/throw during `server.listen` startup instead of logging `[butler] auto-start failed`.
  - Suggestion: move all setup work in `spawnButlerPty()` into the existing `try` block, set `butlerSession` to `state: "error"` on failure, and ensure callers receive/log the `{ ok: false, error }` result.

- [medium] Build acceptance criterion is not verified on the PR.
  - File: `server/index.js:1392`
  - Details: Issue #631 requires `npm run build` to pass, but GitHub reports no checks for this branch and the PR body leaves the build test unchecked. This is a backend server change, so we need a build signal before merge.
  - Suggestion: run `npm run build` and update the PR/re-review note with the result, or attach CI/check output if available.

### Decision
Requesting changes because the start/auto-start path should fail safely, and the required build verification is missing.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review fix addresses the blocking robustness issue: Butler setup work now runs inside the `spawnButlerPty()` try/catch, so mkdir/config/seed-copy failures return the controlled `{ ok: false, error }` path instead of escaping the API/WebSocket/auto-start callers.

### Findings
- None blocking.

### Decision
Approved. GitHub still reports no CI checks for this branch, but Dev reported that `npm run build` now passes in the re-review note.

Issue #652, opened 2026-05-04
Scrollback replay sends Buffer (binary frame) — xterm silently drops Blob


realproject7:
## Bug
Butler terminal (and potentially agent terminals with accumulated scrollback) renders blank because scrollback replay is sent as a binary WebSocket frame. The browser receives it as a Blob, and xterm.js silently drops Blob input — it only accepts `string | Uint8Array`.

Confirmed independently by two separate agents with browser DevTools evidence.

## Root cause
`server/scrub-secrets.js:46-49`:
```js
function scrubScrollback(buf) {
  if (!buf || buf.length === 0) return buf;
  return Buffer.from(scrubSecrets(buf.toString("utf-8")), "utf-8");
}
```

Returns a **Buffer**. Node's `ws` library sends Buffer as binary frames. Browser receives binary frames as Blob (default `binaryType`). `term.write(blob)` is a silent no-op in xterm.js v6.

## Why agent terminals work but Butler doesn't
- Agent terminals: scrollback is usually empty on connect → hits the `else` branch → sends string `"[agent online — waiting for input]"` → text frame → renders fine
- Live PTY output: goes through `scrubSecrets()` which returns a **string** → text frame → renders fine
- Butler: long-lived process, scrollback has accumulated bytes → hits the `if` branch → Buffer → binary frame → Blob → dropped

## Evidence
- WebSocket frame type: Binary (not Text)
- `e.data` in browser: `[object Blob]`, `isBlob: true`
- Typing in the input box also produces no output (same binary frame path for live data when scrubSecrets wraps in Buffer)
- Container dimensions are fine (802x360), xterm canvas/DOM renderer exists, cursor visible at 0,0

## Fix
One-line change in `server/scrub-secrets.js:48` — return string instead of Buffer:

```js
// Current:
return Buffer.from(scrubSecrets(buf.toString("utf-8")), "utf-8");

// Fix:
return scrubSecrets(buf.toString("utf-8"));
```

This makes `scrubScrollback()` return a string, which `ws.send()` sends as a text frame, which the browser receives as a string, which `term.write()` renders correctly.

## Note
This also affects agent terminals if they ever have accumulated scrollback on reconnect (e.g., after page refresh while agents are actively producing output). It's just rare in practice because agent terminals are usually empty on first connect.

## Acceptance criteria
- [ ] `scrubScrollback()` returns string, not Buffer
- [ ] Butler terminal renders scrollback on home screen
[...]

PR #653, opened 2026-05-04
[#652] Fix scrollback replay binary frame dropping

realproject7:
## Summary
- `scrubScrollback()` in `server/scrub-secrets.js` was returning a `Buffer`, causing `ws.send()` to emit a binary WebSocket frame
- Browser receives binary frames as `Blob`, and xterm.js silently drops `Blob` input
- One-line fix: return the string from `scrubSecrets()` directly instead of wrapping in `Buffer.from()`
- Fixes both Butler terminal and agent terminal scrollback replay on reconnect

## Test plan
- [ ] Enable Butler → navigate away → navigate back to Home → scrollback renders (not blank)
- [ ] Agent terminal with active output → refresh page → scrollback replays correctly
- [ ] `npm run build` passes

### DIFF
--- a/server/__tests__/scrub-secrets.test.js
+++ b/server/__tests__/scrub-secrets.test.js
@@ -172,31 +172,32 @@ test("handles multi-line with mixed secrets and normal output", () => {
   assert(!out.includes("eyJhbGci"));
 });
 
-// --- 5. scrubScrollback (Buffer handling) ---
+// --- 5. scrubScrollback (string output for WebSocket text frames) ---
 
-test("scrubScrollback returns Buffer", () => {
+test("scrubScrollback returns string, not Buffer", () => {
   const buf = Buffer.from("SOME_SECRET_KEY=abc123\nnormal line");
   const out = scrubScrollback(buf);
-  assert(Buffer.isBuffer(out), "should return a Buffer");
-  assert(!out.toString().includes("abc123"));
+  assert.strictEqual(typeof out, "string", "should return a string for text WebSocket frames");
+  assert(!out.includes("abc123"));
 });
 
 test("scrubScrollback handles empty buffer", () => {
   const buf = Buffer.alloc(0);
   const out = scrubScrollback(buf);
-  assert(Buffer.isBuffer(out));
-  assert.strictEqual(out.length, 0);
+  assert.strictEqual(typeof out, "string");
+  assert.strictEqual(out, "");
 });
 
 test("scrubScrollback handles null", () => {
-  assert.strictEqual(scrubScrollback(null), null);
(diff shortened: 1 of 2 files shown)


realproject7 (commented):
**RE2 review: ✅ Approve (posted as comment — same GitHub account)**

Clean one-line fix with correct root cause:

- `scrubScrollback()` was wrapping the scrubbed string back in `Buffer.from()`, causing `ws.send(buffer)` to emit a binary WebSocket frame.
- Browser receives binary frames as `Blob`, which xterm.js silently drops (it expects string/text frames).
- Fix: return the string directly from `scrubSecrets()` — `ws.send(string)` emits a text frame that xterm.js handles correctly.
- Affects both Butler and agent terminal scrollback replay on reconnect.

No issues found.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The production one-line fix is directionally correct, but the existing focused regression tests were not updated and still assert the old Buffer behavior. That leaves the acceptance criterion around agent terminal replay coverage unmet and makes the checked-in test suite stale.

### Findings
- [medium] Existing `scrubScrollback` tests still expect a Buffer return value.
  - File: `server/__tests__/scrub-secrets.test.js:177`
  - Suggestion: Update the scrubScrollback tests to assert a string return for non-empty scrollback, preserve redaction, and cover the replay contract that `ws.send(scrubScrollback(...))` receives text rather than a Buffer. The empty-buffer expectation may need to become `""` as well if the function contract is now string output.

### Decision
Requesting changes because issue #652 explicitly calls for agent terminal replay to work or have a focused regression test, and the existing focused tests currently encode the opposite behavior.

realproject7 (commented):

RE1's finding addressed:

- **Consistent string contract**: `scrubScrollback` now returns `""` for null/empty input instead of passing through the original value. Function always returns a string.
- **Tests updated**: All 4 `scrubScrollback` tests assert `typeof out === "string"`. `null` input expects `""`. Buffer assertions removed. Redaction and content preservation still verified.

No issues found.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The rework now makes `scrubScrollback()` consistently return strings for replay, including empty/null input, and updates the focused scrub-secrets tests to assert the new text-frame contract instead of the old Buffer behavior.

### Findings
- None.

### Decision
Approving. The production fix covers both Butler and agent replay call sites because both send `scrubScrollback(...)` directly through `ws.send()`, and the updated tests now guard against regressing to binary frames. GitHub reports no CI checks; Dev reported all 28 tests and `npm run build` pass.

Issue #654, opened 2026-05-04
Butler stuck at trust prompt — pre-trust and --dangerously-skip-permissions don't bypass it


realproject7:
## Bug
Butler terminal shows Claude Code's trust prompt ("Do you trust this folder?") even though:
- `--dangerously-skip-permissions` is passed (doesn't skip the trust dialog — it only skips tool permission prompts)
- `claude -p "echo ok"` pre-trust runs before spawn (doesn't persist for full PTY sessions)

The ButlerChat input box has `disableStdin: true` on xterm, so the user can't click the terminal to select option 1. They must type `1` in the message input and click Send as a workaround.

## Root cause
Claude Code has two separate security gates:
1. **Directory trust** — "Do you trust this folder?" (interactive, per-directory, first-time)
2. **Permission bypass** — `--dangerously-skip-permissions` (skips tool approval prompts)

The `--dangerously-skip-permissions` flag does NOT skip the trust prompt. And `claude -p "echo ok"` trusts the directory for `-p` (print) mode but that trust may not carry over to full interactive PTY sessions.

## Fix
Auto-answer the trust prompt by sending `1\r` to the PTY shortly after spawn if the trust prompt is detected.

In `spawnButlerPty()` in `server/index.js`, after `pty.spawn()`:

```js
// Auto-answer trust prompt if it appears within the first few seconds
let trustHandled = false;
const trustListener = term.onData((data) => {
  if (trustHandled) return;
  if (data.includes("trust") || data.includes("Yes,") || data.includes("1.")) {
    setTimeout(() => {
      if (!trustHandled) {
        term.write("1\r");
        trustHandled = true;
      }
    }, 500);
  }
});
// Clean up listener after 10s (trust prompt only appears at startup)
setTimeout(() => {
  trustListener.dispose();
  trustHandled = true;
}, 10000);
```

Alternative simpler approach — just always send `1\r` after a 2s delay on first spawn:

```js
setTimeout(() => {
  if (butlerSession.term === term) {
    term.write("1\r");
  }
}, 2000);
```

This is safe because:
- If trust prompt is showing: selects "Yes" and continues
- If trust prompt was already answered: `1\r` is sent to Claude Code's input prompt, which just ignores a bare `1`

## Acceptance criteria
- [ ] Butler auto-answers trust prompt without user intervention
- [ ] Butler session starts and shows Claude Code prompt after trust
- [ ] No side effects if trust was already granted (sending `1\r` to active session is harmless)
[...]

PR #655, opened 2026-05-04
[#654] Auto-answer Butler trust prompt

realproject7:
## Summary
- Added trust prompt auto-answer in `spawnButlerPty()`: listens for trust-related output ("trust", "Yes,", "1.") in the first 10s after spawn
- When detected, sends `1\r` after 500ms delay to select "Yes, continue"
- Listener self-cleans after 10s — no side effects if trust was already granted
- Safe if no trust prompt: the listener simply expires without sending anything

## Test plan
- [ ] Fresh Butler start with untrusted directory → trust prompt auto-answered, session starts
- [ ] Butler start with already-trusted directory → no side effects, session starts normally
- [ ] Trust listener cleans up after 10s (no lingering data handlers)
- [ ] `npm run build` passes

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1473,20 +1473,22 @@ function spawnButlerPty() {
       }
     });
 
-    // Auto-answer trust prompt if it appears within the first 10s
-    let trustHandled = false;
-    const trustListener = term.onData((data) => {
-      if (trustHandled) return;
-      if (data.includes("trust") || data.includes("Yes,") || data.includes("1.")) {
-        setTimeout(() => {
-          if (!trustHandled && butlerSession.term === term) {
-            term.write("1\r");
-            trustHandled = true;
-          }
-        }, 500);
-      }
-    });
-    setTimeout(() => { trustListener.dispose(); trustHandled = true; }, 10000);
+    // Auto-answer Claude's trust prompt if it appears within the first 10s
+    if (command === "claude") {
+      let trustHandled = false;
+      const trustListener = term.onData((data) => {
+        if (trustHandled) return;
+        if (data.includes("trust") || data.includes("Yes,") || data.includes("1.")) {
+          setTimeout(() => {
+            if (!trustHandled && butlerSession.term === term) {
+              term.write("1\r");
+              trustHandled = true;
+            }
+          }, 500);
+        }
+      });
+      setTimeout(() => { trustListener.dispose(); trustHandled = true; }, 10000);
+    }
 
     term.onExit(({ exitCode }) => {
       if (butlerSession.term === term) {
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The Claude trust-prompt approach is reasonable, but the listener is currently installed for every Butler CLI. Since Butler can now be configured as Claude, Codex, or Gemini, this can inject `1\r` into non-Claude sessions based on unrelated startup output.

### Findings
- [high] Trust auto-answer listener is not scoped to Claude.
  - File: `server/index.js:1478`
  - Suggestion: Only install the trust prompt listener when the Butler command resolves to Claude, matching the existing pre-trust block. Codex/Gemini do not have this Claude Code directory trust prompt and should not receive auto-written input if their first 10s of output contains `trust`, `Yes,`, or `1.`.

### Decision
Requesting changes because the fix can affect non-Claude Butler sessions introduced by the recent Butler CLI selection work.

realproject7 (commented):
**RE2 review: ✅ Approve with note (posted as comment — same GitHub account)**

The approach is correct — detect trust prompt, auto-answer within a time window, self-clean:

- **Self-cleanup**: 10s timeout disposes listener and sets `trustHandled = true` ✓
- **Session guard**: `butlerSession.term === term` prevents writing to a stale terminal ✓
- **Multiple-match safety**: `trustHandled` flag inside setTimeout prevents duplicate sends ✓
- **No side effects if already trusted**: listener simply expires ✓

**[medium] Loose pattern matching**: The patterns `"trust"`, `"Yes,"`, `"1."` could match normal Claude output within the first 10 seconds (e.g., numbered lists, conversational text mentioning trust). This would cause an unintended `1\r` input. Consider tightening to more specific patterns like `"Do you trust"` or `"Yes, continue"` to reduce false positive risk. Not blocking since the 10s window limits blast radius, but worth hardening in a follow-up.

No blocking issues.

realproject7 (commented):

Fix correctly scopes the trust prompt listener to `command === "claude"` only. Codex and Gemini backends are no longer affected. All previous correctness points still hold.

No blocking issues.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review fix resolves the blocker: the trust prompt auto-answer listener is now installed only for Claude Butler sessions, matching the Claude-specific trust prompt and pre-trust path.

### Findings
- None.

### Decision
Approving. The implementation now satisfies issue #654 without affecting Codex/Gemini Butler sessions. Non-blocking note: the match strings are still broad (`trust`, `Yes,`, `1.`), so tightening them to prompt-specific text would be a useful follow-up. GitHub reports no CI checks; Dev reported local `npm run build` passes.

Issue #673, opened 2026-05-05
Settings: sticky save bar + Butler restart hint on CLI/model change


realproject7:
### 1. Save button hard to notice
The Save button is only at the very top and bottom of the page. When users change a setting mid-page (like Butler CLI), they don't see the Save button without scrolling. Easy to miss changes that need saving.

**Fix:** Add a sticky save bar that appears at the bottom of the viewport when there are unsaved changes:

```
┌─────────────────────────────────────────────────┐
│  Unsaved changes                        [Save]  │
└─────────────────────────────────────────────────┘
```

- Only visible when config has been modified (dirty state)
- Fixed to bottom of viewport (`fixed bottom-0`)
- Disappears after save
- Subtle styling: border-top, bg-bg-surface, accent Save button

### 2. Butler needs restart after CLI/model change — no indication
Changing Butler's CLI or model saves to config but the running Butler process keeps the old settings. The user has no way to know they need to restart Butler.

**Fix:** Show an inline hint below the Butler section when CLI or model changed while Butler is running:

```
⚠ Butler is running with previous settings. Disable and re-enable to apply changes.
```

- Only shows when saved config differs from what Butler was started with
- Yellow/warning text, small (`text-[10px]`)
- Disappears after Butler is restarted

## Acceptance criteria
- [ ] Sticky save bar appears at bottom when settings are dirty
- [ ] Sticky bar disappears after save
- [ ] Butler section shows restart hint when CLI/model changed while running
- [ ] Existing top/bottom Save buttons can remain (sticky bar is supplemental)
- [ ] npm run build passes

realproject7:
Merged via PR #675.

PR #675, opened 2026-05-06
[#673] Settings sticky save bar + Butler restart hint

realproject7:
## Summary
- Adds a fixed sticky save bar at the bottom of the viewport that appears when settings have unsaved changes and disappears after save
- Adds an inline warning hint below the Butler section when CLI or model has been changed while Butler is running
- Existing top/bottom Save buttons remain untouched

## Test plan
- [ ] Open Settings, change any field → sticky bar appears at bottom with "Unsaved changes" + Save button
- [ ] Click Save on sticky bar → bar disappears
- [ ] Start Butler, change Butler CLI or model → yellow warning appears: "Butler is running with previous settings..."
- [ ] Disable and re-enable Butler → warning disappears
- [ ] `npm run build` passes

Reviewers: @re1 @re2

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1555,6 +1555,8 @@ function spawnButlerPty() {
       state: "running",
       error: null,
       scrollback: Buffer.alloc(0),
+      command,
+      model: butlerCfg.model || "",
     };
 
     const SCROLLBACK_SIZE = 64 * 1024;
@@ -1637,9 +1639,12 @@ app.post("/api/butler/stop", (_req, res) => {
 });
 
 app.get("/api/butler/status", (_req, res) => {
+  const running = butlerSession.state === "running" && !!butlerSession.term;
   res.json({
-    running: butlerSession.state === "running" && !!butlerSession.term,
+    running,
     pid: butlerSession.term ? butlerSession.term.pid : null,
+    command: running ? butlerSession.command : undefined,
+    model: running ? butlerSession.model : undefined,
   });
 });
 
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The sticky save bar is scoped appropriately, but the Butler restart hint does not reliably represent the settings the running Butler process actually started with. After a Settings reload while Butler is already running, the PR infers the start snapshot from the current saved config, which can hide a real CLI/model mismatch.

### Findings
- [high] Restart hint is lost after reloading Settings while Butler is already running
  - File: `src/components/SettingsPage.tsx:398`
  - Details: In `refreshButlerStatus`, when `/api/butler/status` reports `running` and `butlerStartConfig.current` is empty, the code sets `butlerStartConfig` from `savedConfigRef.current`. That is the current saved config, not necessarily the settings the already-running Butler process was launched with. Repro: start Butler with CLI/model A, change CLI/model to B and save, then reload Settings. The running process is still on A, but this code snapshots B as the start config, so `butlerConfigChanged` becomes false and the warning disappears even though a restart is still required.
  - Suggestion: Track the Butler process start command/model at the time it is launched in a place that survives Settings remounts, or have `/api/butler/status` return the command/model for the currently running process and compare saved config against that. Do not infer an already-running process's start settings from the current saved config.

### Decision
Requesting changes because the restart hint fails a normal persistence/remount case and no CI checks are configured for this PR.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The requested blocker is fixed: `/api/butler/status` now reports the command/model for the active Butler session, and Settings compares against that server-side running-process state instead of reconstructing it from the current saved config. The sticky save bar and restart hint remain scoped to issue #673.

### Findings
- None blocking.

### Decision
Approved. The branch has no GitHub checks configured, so I could not verify CI from GitHub, but the updated diff satisfies the issue acceptance criteria and resolves the prior reload/persistence concern.

PR #701, opened 2026-05-15
[#698] Fix GitHub secondary rate limit with multi-project polling

### LINKED ISSUE #698: GitHub API rate limit exceeded with multiple projects and clients
## Summary

GitHub's GraphQL secondary rate limit is triggered when multiple clients (e.g., local browser + remote via Tailscale) poll the GitHub panel with multiple projects configured.

## Root Cause

Each connected client fires 4 GitHub API calls per project every 30 seconds:
- `/api/github/issues`
- `/api/github/prs`
- `/api/github/closed-issues`
- `/api/github/merged-prs`

Plus `/api/github/rate-limit` every 60s.

With 5 projects configured, that's **20 `gh` CLI calls per 30s per client** (~40/min). The server-side cache (`GH_ENDPOINT_CACHE_TTL = 30_000`) helps when multiple clients poll within the same window, but simultaneous cache expirations across endpoints cause burst traffic that triggers GitHub's secondary/abuse rate limit — even when the hourly quota (5,000) is not exhausted.

## Observed Behavior

- Error: `GraphQL: API rate limit already exceeded for user ID <id>`
- `gh api rate_limit` shows ~4,000/5,000 remaining — hourly quota is fine
- The error is a **secondary rate limit** (burst detection), not the primary hourly quota

## Suggested Fix

- Increase `GH_ENDPOINT_CACHE_TTL` from 30s to 60s+ (or make it configurable)
- Increase frontend polling interval from 30s to 60s (`setInterval(fetchData, 30000)` in `GitHubPanel.tsx`)
- Consider staggering cache expiration per endpoint to avoid burst refreshes

## Environment

- 5 projects configured
- 2 clients connected (local + remote via Tailscale)
- macOS, `gh` CLI shelling out to GitHub GraphQL API

realproject7:
## Summary
- Increases `GH_ENDPOINT_CACHE_TTL` from 30s to 60s — halves server-side API call rate
- Increases `GitHubPanel.tsx` frontend polling from 30s to 60s to match server cache
- Adds deterministic per-endpoint jitter (0–15s) to stagger cache expirations, preventing burst refreshes that trigger GitHub's secondary/abuse rate limit

With 5 projects and 2 clients, this reduces API calls from ~40/min to ~20/min per client. The jitter ensures endpoints don't all expire simultaneously.

## Test plan
- [ ] Verify GitHubPanel still refreshes data (now at 60s intervals)
- [ ] Verify cache TTL is 60s base + jitter per endpoint
- [ ] Verify jitter is deterministic per cache key (no randomness)
- [ ] Verify adaptive TTL still kicks in for low/critical rate states
- [ ] `npm run build` passes

Fixes #698

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -75,10 +75,13 @@ function adaptiveTTL(baseTTL) {
 // Wraps a synchronous execFileSync gh call with an in-memory cache that
 // serves stale data when rate-limited instead of hammering the API.
 const _ghEndpointCache = new Map(); // key → { ts, data }
-const GH_ENDPOINT_CACHE_TTL = 30_000; // 30s base TTL
+const GH_ENDPOINT_CACHE_TTL = 60_000; // #698: 60s base TTL (was 30s)
 
 function cachedGhEndpoint(cacheKey, ghArgs, res, { transform } = {}) {
-  const ttl = adaptiveTTL(GH_ENDPOINT_CACHE_TTL);
+  // #698: add per-endpoint jitter (0–15s) to stagger cache expirations and
+  // avoid burst traffic that triggers GitHub's secondary rate limit.
+  const jitter = (cacheKey.length * 7919) % 15_000; // deterministic per key
+  const ttl = adaptiveTTL(GH_ENDPOINT_CACHE_TTL) + jitter;
   const cached = _ghEndpointCache.get(cacheKey);
   if (cached && Date.now() - cached.ts < ttl) {
     return res.json(cached.stale ? { ...cached.data, _stale: true } : cached.data);
--- a/server/routes.js
+++ b/server/routes.js
@@ -77,11 +77,11 @@ function adaptiveTTL(baseTTL) {
 const _ghEndpointCache = new Map(); // key → { ts, data }
 const GH_ENDPOINT_CACHE_TTL = 60_000; // #698: 60s base TTL (was 30s)
 
+// #698: track in-flight background refreshes to avoid duplicate fetches
+const _ghRefreshing = new Set();
+
 function cachedGhEndpoint(cacheKey, ghArgs, res, { transform } = {}) {
-  // #698: add per-endpoint jitter (0–15s) to stagger cache expirations and
-  // avoid burst traffic that triggers GitHub's secondary rate limit.
-  const jitter = (cacheKey.length * 7919) % 15_000; // deterministic per key
-  const ttl = adaptiveTTL(GH_ENDPOINT_CACHE_TTL) + jitter;
+  const ttl = adaptiveTTL(GH_ENDPOINT_CACHE_TTL);
   const cached = _ghEndpointCache.get(cacheKey);
   if (cached && Date.now() - cached.ts < ttl) {
     return res.json(cached.stale ? { ...cached.data, _stale: true } : cached.data);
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR reduces the polling frequency, but the added cache-expiration jitter does not actually stagger refreshes with the new 60s frontend polling cadence. The burst behavior that #698 is trying to mitigate can still happen, just less frequently.

### Findings
- [medium] The 0-15s TTL jitter is skipped by the 60s polling cadence, so endpoints still refresh in a burst.
  - File: `server/routes.js:83`
  - `GitHubPanel` now polls every 60s, while `cachedGhEndpoint()` uses a TTL of `60_000 + jitter(0..15_000)`. If all four endpoint caches are populated on the same poll, then at t=60s they are still fresh because of the positive jitter, and the frontend will not ask again until t=120s. At that point every endpoint's 60-75s TTL has expired, so `/issues`, `/prs`, `/closed-issues`, and `/merged-prs` all miss together and call `gh` in one burst. That preserves the secondary-rate-limit failure mode every other poll. Please make the staggering effective under the actual request cadence, for example by storing a per-key next refresh deadline with a wider spread than the poll interval, using per-endpoint scheduled refresh/backoff, or otherwise ensuring cache misses do not align on the same frontend poll.

### Decision
Requesting changes because the PR's burst-mitigation mechanism is ineffective with the paired 60s client interval, leaving the core #698 secondary-rate-limit scenario only partially addressed.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The follow-up improves response behavior by serving stale data immediately, but it does not resolve the burst of GitHub API calls that triggers #698. Expired endpoint requests still launch background `gh` calls at the same time.

### Findings
- [medium] Stale-while-revalidate moves the burst to the background instead of preventing it.
  - File: `server/routes.js:94`
  - When the 60s poll hits stale caches, each endpoint with cached data immediately enters the new background refresh branch and starts `_execFileAsync("gh", ghArgs, ...)`. `_ghRefreshing` prevents duplicate refreshes for the same cache key, but it does not stagger different keys. With 5 projects, the first stale poll can still start `/issues`, `/prs`, `/closed-issues`, and `/merged-prs` refreshes for every project at once, i.e. the same ~20 `gh` calls in a burst that #698 identifies as the secondary-rate-limit trigger. Please add actual cross-key throttling/staggering, such as a small refresh queue with limited concurrency, per-key scheduled refresh deadlines, or another mechanism that bounds simultaneous `gh` refreshes across endpoint keys.

### Decision
Requesting changes because the core secondary-rate-limit problem is burst API execution, not only request latency; the current fix still allows aligned background bursts.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The latest revision resolves the burst-refresh issue by combining the 60s poll/cache interval with stale-while-revalidate and a concurrency-limited background refresh queue. This bounds simultaneous GitHub CLI calls across endpoint keys instead of letting all expired endpoints refresh together.

### Findings
- No blocking findings.

### Decision
Approved. `_ghRefreshQueue` and `_ghRefreshing` prevent duplicate queued refreshes per key, cap concurrent background `gh` calls at 2, and keep stale responses non-blocking while refreshes run. `gh pr checks` reports no checks for this branch, so there was no CI result to verify.

Issue #730, opened 2026-05-20
[#713] PTY injection dispatcher — deliver chat messages to agents via terminal stdin


realproject7:
## Problem
MCP server notifications (`notifications/message`) are not processed by Claude Code or Codex CLIs. Agents in file-chat mode never receive messages — they sit idle at "agent online — waiting for input" regardless of @mentions or scheduled triggers.

Confirmed empirically: messages are correctly appended to JSONL, MCP shim callback fires, but the CLI ignores the notification.

## Fix: PTY stdin injection as the primary delivery mechanism

When a message with @mentions is appended to JSONL, QuadWork writes the message content directly into the mentioned agent's PTY stdin. The agent treats it as a user prompt and responds.

### Delivery path
```
appendMessage() → extract mentions → for each mentioned agent:
  → check idle (not mid-turn)
  → check loop guard (not paused)
  → coalesce if burst
  → safeWrite(session.term, formattedMessage)
```

### Injection format
```
[chat @dev]: @re1 @re2 PR #187 ready for review — issue #53.
```

NOT a meta-instruction like "go check chat." The actual message content so the agent treats it as a direct prompt and responds.

### Guardrails

**1. Idle gating**
Don't inject while the agent is mid-response (producing PTY output). Track `session.lastOutputAt` — if output was produced within the last 5 seconds, queue the injection and deliver when idle. The JSONL is the durable queue — no separate buffer needed.

**2. Coalesce bursts**
If 5 messages mention `@head` within 1 second, inject ONE combined prompt:
```
[chat] 5 new messages mentioning @head. Latest from @dev: PR #187 ready for review.
```
Use a 1-second coalescing window per agent.

**3. Loop guard suppression**
When loop guard is paused for a project, do NOT inject into any agent's PTY for that project. Injections resume only after the user sends a message or `/continue`.

**4. Only inject for @mentions**
Don't inject for every chat message. Only when the agent is specifically @mentioned or the message is a broadcast (`@head @dev @re1 @re2`). System messages are never injected.

**5. Per-CLI format verification**
Test that the injection format works for both Claude Code and Codex. Both accept text on stdin as a user prompt. Verify Enter/newline handling.

### Implementation in `server/index.js` or `server/file-chat.js`

Add to the `appendMessage` callback path:

```js
function dispatchToAgentPTY(projectId, msg) {
  if (msg.type === "system") return;
  if (isLoopGuardPaused(projectId)) return;
[...]

PR #733, opened 2026-05-20
[#730] PTY injection dispatcher for file-chat agents

realproject7:
## Summary
- Adds `server/pty-dispatcher.js` — delivers chat messages to agents via terminal stdin when @mentioned in file-chat mode
- Idle gating: queues injection when agent is mid-output (last output <5s ago), drains via `term.onData` + high-water mark `since_id`
- Burst coalescing: multiple mentions within 1s window produce a single combined injection
- Loop guard suppression: no injection when loop guard is paused
- Self-mention filtering: agents don't receive their own messages
- Wired into `POST /api/chat` via callback pattern (routes.js → index.js) so scheduled triggers also go through this path
- Cleanup on session stop/exit to prevent timer leaks

## Test plan
- [x] 14 unit tests in `server/pty-dispatcher.test.js` covering:
  - System messages not injected
  - Loop guard paused blocks injection
  - No injection without mentions
  - Idle agent receives formatted injection
  - Busy agent queues pending wake
  - 5 rapid mentions coalesce into one injection
  - Agent doesn't inject its own message
  - Stopped session receives no injection
  - `cleanupSession` clears all timers/state
- [x] Existing tests pass (`file-chat.system-messages.test.js`)
- [ ] Manual: send `@head ping` from dashboard → Head terminal shows message → Head responds
- [ ] Manual: verify with both Claude Code and Codex CLIs

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -17,6 +17,7 @@ const {
 const { waitForAgentChattrReady, registerAgent, registerAgentWithRetry, deregisterAgent, startHeartbeat, stopHeartbeat } = require("./agentchattr-registry");
 const { patchAgentchattrCss, patchCrashTimeout } = require("./install-agentchattr");
 const { startQueueWatcher, stopQueueWatcher } = require("./queue-watcher");
+const { dispatchToAgentPTY, cleanupSession: cleanupPtyDispatcher } = require("./pty-dispatcher");
 
 const net = require("net");
 const config = readConfig();
@@ -40,6 +41,14 @@ app.use(express.json({ limit: "10mb" }));
 
 // --- Mount migrated API routes (from Next.js) ---
 app.use(routes);
+
+// #730: wire PTY injection dispatcher into the chat route
+routes.setPtyDispatchCallback((projectId, msg) => {
+  dispatchToAgentPTY(projectId, msg, agentSessions, {
+    isLoopGuardPaused: fileChat.isLoopGuardPaused,
+    safeWrite,
+  });
+});
 
 const server = http.createServer(app);
 
@@ -761,6 +770,7 @@ async function spawnAgentPty(project, agent, opts = {}) {
     term.onExit(({ exitCode }) => {
       const current = agentSessions.get(key);
       if (current && current.term === term) {
+        cleanupPtyDispatcher(key);
         current.state = "stopped";
         current.error = exitCode ? `exit:${exitCode}` : null;
         current.term = null;
@@ -812,6 +822,7 @@ async function stopAgentSession(key) {
   if (session.projectId && session.agentId && !session._suppressLifecycleMsg) {
     emitSystemMessage(session.projectId, `${session.agentId} left`);
   }
+  cleanupPtyDispatcher(key);
   if (session.term) {
     try { session.term.kill(); } catch {}
     session.term = null;
--- a/server/pty-dispatcher.js
+++ b/server/pty-dispatcher.js
@@ -10,10 +10,10 @@ const COALESCE_WINDOW_MS = 1000;
 // Per-agent coalescing timers: key = "project/agent" → timeout handle
 const _coalesceTimers = new Map();
 
-// Per-agent pending since_id for drain: key = "project/agent" → oldest undelivered msg id
+// Per-agent pending state for drain: key = "project/agent" → { sinceId, latestMsg }
 const _pendingSinceId = new Map();
 
-// Per-agent drain listeners: key = "project/agent" → { disposable, timeout }
+// Per-agent drain listeners: key = "project/agent" → { disposable, timeoutHandle }
 const _drainListeners = new Map();
 
 /**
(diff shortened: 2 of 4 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR wires the dispatcher into `/api/chat`, but the busy-agent path does not satisfy #730's drain and injection-format requirements. A mention that arrives shortly after the agent's last output can remain queued forever, and delayed wakes do not include the latest message content.

### Findings
- [high] Busy-agent pending wakes can stall indefinitely if no further PTY output arrives after the message is queued.
  - File: `server/pty-dispatcher.js:63`
  - Suggestion: Start an idle-drain timer when queuing the pending wake, using the remaining time until the 5s idle threshold, and reset it from `term.onData` when additional output arrives. The current implementation only creates the timer inside the `onData` callback, so a message received after the agent's last output but within the 5s busy window never drains unless the agent emits more output.

- [high] Delayed wake injection omits the latest message content and does not provide a usable high-water mark to `chat_read`.
  - File: `server/pty-dispatcher.js:81`
  - Suggestion: Preserve enough pending state to inject the latest message text plus an instruction to call `chat_read` from the oldest pending `since_id`, as required by #730's review updates. The current prompt is generic (`You were mentioned while busy...`) and discards the actual mention content, so the agent cannot respond directly to the message that woke it.

### Decision
Requesting changes because both issues hit core acceptance criteria for idle queue delivery and the required content-plus-chat_read delayed wake behavior.

realproject7:
## RE2 Review — PR #733

Two bugs found in the pending-wake / drain path:

### Bug 1: `sinceId` unused in `buildDrainPrompt` (pty-dispatcher.js:97-103)

The function accepts `sinceId` but never uses it in the output string. The drain prompt tells the agent to `chat_read` but doesn't tell it *from where*. The whole point of tracking the high-water mark `_pendingSinceId` is to pass it to the agent so it reads only missed messages. Without it, the agent gets a generic "you were mentioned" prompt with no way to target the right messages efficiently.

**Fix:** Include `since_id` in the injected prompt, e.g.:
```js
`mcp read #general with sender: "${agentId}", since_id: ${sinceId} — ...`
```

### Bug 2: Stale timeout reference in `_drainListeners` (pty-dispatcher.js:73-82)

```js
let idleTimeout = null;                          // ← null
const disposable = session.term.onData(() => {
  if (idleTimeout) clearTimeout(idleTimeout);
  idleTimeout = setTimeout(...);                 // ← reassigned inside callback
});
_drainListeners.set(key, { disposable, timeout: idleTimeout }); // ← captures null
```

`_drainListeners` stores `timeout: idleTimeout` at registration time, when it's still `null`. The actual timeout is created later inside the `onData` callback and assigned to the *local variable* — the stored object never sees it. So `cleanupDrainListener` → `clearTimeout(listener.timeout)` is always `clearTimeout(null)`, which is a no-op. The real active timeout leaks.

**Fix:** Use a mutable container:
```js
const state = { timeout: null };
const disposable = session.term.onData(() => {
  if (state.timeout) clearTimeout(state.timeout);
  state.timeout = setTimeout(...);
});
_drainListeners.set(key, { disposable, state });
```
Then in `cleanupDrainListener`: `clearTimeout(listener.state.timeout)`.

---

### Minor nits (non-blocking)

- **Double loop-guard check:** `routes.js:1184` already gates on `!isLoopGuardPaused` before calling the callback, and `dispatchToAgentPTY` checks it again. Harmless but redundant — consider removing one.
- **`submitDelayMs = Math.max(300, flat.length)`** uses character count as milliseconds. Works but could be surprising for very long messages (2000 chars = 2s delay). A comment or a cap would help.

Tests are solid and cover the key paths well. Architecture (callback pattern to avoid circular deps, coalescing, idle gating) is clean.

**Verdict: Request changes** — the two bugs above should be fixed before merge.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The updated PR resolves the blocking RE1 findings. Pending wakes now have an immediate fallback drain timer, additional PTY output resets that timer, and delayed prompts include the latest message content plus catch-up context.

### Findings
- None blocking.

### Decision
Approve. The implementation now satisfies the core #730 requirements for mentioned-agent PTY delivery, idle queue drain, burst coalescing, loop guard suppression, system-message suppression, and scheduled trigger delivery via `/api/chat`.

realproject7:
## RE2 Re-review — LGTM

All three findings addressed:

1. **`sinceId` in drain prompt** — Fixed. `buildDrainPrompt` now includes the latest message content and `since_id` for targeted catch-up.
2. **Stale timeout ref** — Fixed. Mutable `state` object ensures `cleanupDrainListener` reaches the active handle.
3. **Fallback drain timer** (RE1's finding) — Fixed. `resetIdleTimer()` fires immediately at registration, preventing stall when no further PTY output arrives.

Tests cover the new behavior (fallback drain, content in prompt, since_id inclusion). Approving.

Issue #736, opened 2026-05-20
[#713] File-chat: agents loop responding to each other's pings


realproject7:
## Problem
When user sends `@head ping dev, re1, re2`, the agents enter a reply loop — each agent's response mentions other agents, which triggers PTY injection to those agents, who respond mentioning others, creating a cascade.

Chat log shows: head pings dev → dev responds mentioning re1/re2 → re1/re2 respond mentioning head → head responds again → dev responds again.

## Root cause
The PTY dispatcher (#730) injects messages for ALL @mentions. When Head says `@dev please confirm`, Dev receives the injection. Dev responds `@head confirmed`, which triggers injection to Head. Head responds mentioning Dev again → loop.

The loop guard (#717) should eventually stop this (after 30 hops), but 30 hops of ping-pong is too many for a simple availability check.

### Option A: Don't inject agent-to-agent messages
Only inject messages where `sender === "user"` or `sender === "system"`. Agent-to-agent messages are recorded in JSONL but not injected — agents read them via `chat_read` when they choose to.

This is too restrictive — Dev needs to see RE1's review verdict.

### Option B: Suppress injection for response chains
After injecting a message to an agent, set a cooldown (e.g., 10 seconds). Don't inject another message to the same agent during cooldown. The agent reads any additional messages via `chat_read` after responding.

### Option C: Only inject if the agent hasn't spoken recently
If the agent posted a message in the last 30 seconds, skip injection — they're already active and reading chat via `chat_read`. Only inject to wake idle agents.

**Recommend Option C** — simplest, matches the "wake idle agents" intent of PTY injection.

## Acceptance criteria
- [ ] Simple ping check doesn't create agent reply loops
- [ ] Agents still receive messages when idle (not recently active)
- [ ] `chat_read` remains the correctness layer for active agents
- [ ] `npm run build` passes

PR #739, opened 2026-05-20
[#736] Suppress PTY injection for recently active agents

realproject7:
## Summary
- Track per-agent `lastChatSentAt` timestamp when an agent sends a chat message
- Skip PTY injection if the target agent sent a message within the last 30 seconds — they're already active and reading chat via `chat_read`
- Prevents agent-to-agent reply loops (e.g., ping cascades) while still waking idle agents
- Cleanup on session stop/exit

## Test plan
- [x] 21 unit tests passing, including 3 new tests:
  - Recently active agent is not injected
  - Injection resumes after 30s suppression window expires
  - Sender's lastChatSentAt is tracked for agent sessions
- [ ] Manual: `@head ping dev, re1, re2` — agents respond once without looping

### DIFF
--- a/server/pty-dispatcher.js
+++ b/server/pty-dispatcher.js
@@ -85,6 +85,9 @@ function queuePendingWake(key, msgId, msg, session, deps) {
     _pendingSinceId.delete(key);
     cleanupDrainListener(key);
 
+    const lastSent = _lastChatSentAt.get(key);
+    if (lastSent && (Date.now() - lastSent < ACTIVE_SUPPRESSION_MS)) return;
+
     const formatted = buildDrainPrompt(session.agentId, pending.sinceId, pending.latestMsg);
     injectIntoTerm(session.term, formatted, deps);
   };
@@ -145,6 +148,9 @@ function scheduleCoalescedInjection(key, projectId, agentId, msg, agentSessions,
     _coalesceTimers.delete(key);
     const session = agentSessions.get(key);
     if (!session || !session.term || session.state !== "running") return;
+
+    const lastSent = _lastChatSentAt.get(key);
+    if (lastSent && (Date.now() - lastSent < ACTIVE_SUPPRESSION_MS)) return;
 
     const msgs = state.messages;
     let formatted;
--- a/server/pty-dispatcher.js
+++ b/server/pty-dispatcher.js
@@ -6,9 +6,13 @@
 
 const IDLE_THRESHOLD_MS = 5000;
 const COALESCE_WINDOW_MS = 1000;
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The approach matches #736, but the suppression check only runs when the message is first dispatched. Any injection already queued by the 1s coalescer or 5s busy-agent drain can still fire after the target agent has become active, which leaves a loop path open.

### Findings
- [high] Pending/coalesced injections do not re-check recent-agent activity before writing to the PTY.
  - File: `server/pty-dispatcher.js:72`
  - Suggestion: Re-check `_lastChatSentAt` for the target key inside `drainFn` and inside the coalesce timer callback immediately before `injectIntoTerm`; if the target sent chat within `ACTIVE_SUPPRESSION_MS`, drop the pending injection and clean up state. Otherwise, an agent can send a chat message after a wake was queued but before the timer fires, and still receive the stale PTY injection despite being recently active.

### Decision
Requesting changes because this affects the core #736 acceptance criterion: recently active agents should not be PTY-injected, including injections delayed by idle gating or burst coalescing.

realproject7:
## RE2 Review — LGTM

Logic is sound: if an agent posted a chat message within the last 30s, skip PTY injection — they're already active and reading chat via `chat_read`. This cleanly breaks agent-to-agent ping cascades without affecting idle-agent wakeups.

Verified:
- Sender tracking only applies to recognized agents (`agentSessions.has(senderKey)`), not humans
- `cleanupSession` clears `_lastChatSentAt` state
- Tests cover suppression active, suppression expired, and sender tracking

Minor nit (non-blocking): Test 9 (cleanupSession) doesn't assert `_lastChatSentAt` is cleared, even though the cleanup code handles it. Consider adding one line for completeness.

Approving.

realproject7:
## RE2 Re-review — LGTM

Good hardening. The `_lastChatSentAt` re-checks at both deferred injection points (coalesce timer and busy-drain timer) close the race where an agent becomes active while a queued injection is pending. Stale injections are now properly dropped.

Still approving.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The blocking delayed-injection issue is resolved. The dispatcher now re-checks active-agent suppression immediately before both busy-drain and coalesced injections write to the PTY.

### Findings
- None blocking.

### Decision
Approve. The change now satisfies #736's intent: recently active agents are not re-woken by queued stale injections, while idle agents can still be woken and `chat_read` remains the catch-up layer.

PR #749, opened 2026-05-20
[#719] AC → JSONL migration on first boot

### LINKED ISSUE #719: [#713-6] AC → JSONL migration on first boot
## Summary
Migrate existing projects' AC chat history from `agentchattr_log.jsonl` to the new `general.jsonl` format. Runs on first boot after upgrade. Preserves original data, is idempotent, and handles unknown fields.

### Trigger
On server boot, for each project:
1. Check if `~/.quadwork/<project>/chat/.migrated` exists → skip
2. Check if `~/.quadwork/<project>/agentchattr/data/agentchattr_log.jsonl` exists → migrate
3. If neither exists → new project, nothing to migrate

### Conversion
Read AC's JSONL line by line. For each valid JSON record:

```js
function convertAcRecord(record, nextId) {
  const known = {
    id: nextId,
    seq: nextId,
    ts: record.timestamp || record.ts || new Date().toISOString(),
    sender: record.sender || "unknown",
    channel: record.channel || "general",
    type: record.type === "system" ? "system" : "message",
    text: record.text || record.message || "",
    mentions: extractMentions(record.text || record.message || ""),
  };
  // Preserve unknown fields
  const legacy = {};
  for (const [key, val] of Object.entries(record)) {
    if (!["id", "timestamp", "ts", "sender", "channel", "type", "text", "message", "mentions"].includes(key)) {
      legacy[key] = val;
    }
  }
  if (Object.keys(legacy).length > 0) known._legacy = legacy;
  return known;
}
```

### Output
- Write to `~/.quadwork/<project>/chat/general.jsonl` (0600)
- Create `~/.quadwork/<project>/chat/.migrated` with timestamp and stats
- Log: `[migration] plottoon: migrated 1523 messages, 2 skipped (invalid JSON)`

### Safety
- **Read-only access to AC directory** — never modify `agentchattr_log.jsonl`
- **Idempotent** — `.migrated` manifest prevents double-conversion
- **Lossless** — unknown fields preserved in `_legacy`
- **Invalid lines skipped** — counted in log, don't fail migration

### System message
After migration, append: `{"sender":"system","type":"system","text":"Chat history migrated from AC (N messages)"}`

## Dependencies
- Requires #713-1 (JSONL chat engine for `appendMessage`)

## Acceptance criteria
- [ ] AC JSONL converted to new format with correct schema
- [ ] Unknown fields preserved in `_legacy` object
- [ ] Invalid lines skipped with count logged
- [ ] Original `agentchattr_log.jsonl` untouched (read-only)
- [ ] `.migrated` manifest prevents re-migration
- [ ] Migration must complete for a project before that project switches to file mode
- [ ] Write to general.jsonl.tmp, validate, rename to general.jsonl
- [ ] If migration fails, project stays on AC mode with error surfaced in logs
- [ ] `npm run build` passes

## Safety clarification
Normal migration is **read-only** against AC data. The only code allowed to modify AC data is the explicit operator-run `quadwork ac-restore` command (#721).

### Tests
- Round-trip test: a fixture AC JSONL converts to file format with no record loss
- Unknown fields land in `_legacy` object
- Invalid lines are skipped and counted
- Re-running migration on already-migrated project is a no-op (`.migrated` check)

## Phase 1 update
The test project (quadwork-web) already has a `chat/general.jsonl` with messages from Phase 1 testing. Migration must:
- Skip projects that already have a `general.jsonl` AND a `.migrated` manifest
- If `general.jsonl` already exists WITHOUT `.migrated` (Phase 1 test project), SKIP migration for that project. Log: "[migration] <project>: general.jsonl already exists without .migrated — skipping (Phase 1 test project)". The operator accepted discarding AC history when they started Phase 1 testing.
- Use `file-chat.js` functions (`initProject`, `appendMessage`) that already exist from Phase 1

realproject7:
## Summary
- New `server/migrate-ac.js` module converts AC `agentchattr_log.jsonl` to file-chat `general.jsonl` format on server boot
- Converts Unix epoch timestamps to ISO 8601, maps AC types to message/system, preserves unknown fields in `_legacy`
- Writes to `.tmp` file first, validates all records parse as JSON, then renames atomically
- Creates `.migrated` manifest to prevent re-migration (idempotent)
- Skips projects that already have `general.jsonl` without `.migrated` (Phase 1 test projects)
- AC source files are read-only — never modified
- Wired into server boot before file-chat initialization

## Test plan
- [x] Round-trip test: AC JSONL converts with no record loss
- [x] Unknown fields preserved in `_legacy` object
- [x] Original AC type preserved in `_legacy` when mapped (e.g. `leave` → `message`)
- [x] Invalid lines skipped and counted
- [x] Re-running migration is a no-op (`.migrated` check)
- [x] Phase 1 test project with existing `general.jsonl` is skipped
- [x] New project with no AC log is skipped
- [x] `npm run build` passes

Closes #719

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -18,6 +18,7 @@ const { waitForAgentChattrReady, registerAgent, registerAgentWithRetry, deregist
 const { patchAgentchattrCss, patchCrashTimeout } = require("./install-agentchattr");
 const { startQueueWatcher, stopQueueWatcher } = require("./queue-watcher");
 const { dispatchToAgentPTY, cleanupSession: cleanupPtyDispatcher } = require("./pty-dispatcher");
+const { runAcMigration } = require("./migrate-ac");
 
 const net = require("net");
 const config = readConfig();
@@ -3033,6 +3034,9 @@ server.listen(PORT, "127.0.0.1", async () => {
       console.log(`[startup] ${p.id}: AC already alive on port ${acPort} — tracking`);
     }
   }
+  // #719: Migrate AC chat history to JSONL before initializing file-chat.
+  runAcMigration(startupCfg);
+
   // #714: Initialize file-chat engine for projects with chat_mode: "file".
   // If the writer lock is held by another live process, refuse to start —
   // enforces the single-writer invariant against the "two terminals" scenario.
--- a/server/index.js
+++ b/server/index.js
@@ -3044,8 +3044,14 @@ server.listen(PORT, "127.0.0.1", async () => {
   for (const p of (startupCfg.projects || [])) {
     if (p.chat_mode === "file") {
       if (migrationFailed.has(p.id)) {
-        console.error(`[startup] ${p.id}: migration failed — keeping AC mode, skipping file-chat init`);
+        console.error(`[startup] ${p.id}: migration failed — reverting to AC mode, skipping file-chat init`);
         p.chat_mode = "ac";
+        const cfg = readConfig();
+        const entry = (cfg.projects || []).find((pr) => pr.id === p.id);
+        if (entry) {
+          entry.chat_mode = "ac";
+          writeConfig(cfg);
+        }
         continue;
       }
       try {
--- a/server/index.js
+++ b/server/index.js
@@ -3035,13 +3035,19 @@ server.listen(PORT, "127.0.0.1", async () => {
     }
   }
   // #719: Migrate AC chat history to JSONL before initializing file-chat.
-  runAcMigration(startupCfg);
+  // Failed projects stay on AC mode — do not initialize file-chat for them.
+  const migrationFailed = new Set(runAcMigration(startupCfg));
 
   // #714: Initialize file-chat engine for projects with chat_mode: "file".
   // If the writer lock is held by another live process, refuse to start —
   // enforces the single-writer invariant against the "two terminals" scenario.
   for (const p of (startupCfg.projects || [])) {
     if (p.chat_mode === "file") {
+      if (migrationFailed.has(p.id)) {
+        console.error(`[startup] ${p.id}: migration failed — keeping AC mode, skipping file-chat init`);
+        p.chat_mode = "ac";
+        continue;
+      }
       try {
         fileChat.initProject(p.id);
         console.log(`[startup] ${p.id}: file-chat engine initialized`);
--- a/server/migrate-ac.js
+++ b/server/migrate-ac.js
@@ -81,6 +81,11 @@ function migrateProject(projectId) {
 
   if (converted.length === 0 && skipped === 0) return null;
 
+  if (converted.length === 0 && skipped > 0) {
+    console.log(`[migration] ${projectId}: AC log has ${skipped} lines but none are valid JSON — skipping`);
+    return null;
+  }
+
   ensureSecureDir(chatDir);
 
   const tmpPath = targetPath + ".tmp";
@@ -132,6 +137,7 @@ function migrateProject(projectId) {
 }
 
 function runAcMigration(config) {
+  const failed = [];
   const projects = config.projects || [];
   for (const project of projects) {
     if (!project || !project.id) continue;
@@ -139,8 +145,10 @@ function runAcMigration(config) {
       migrateProject(project.id);
     } catch (err) {
       console.error(`[migration] ${project.id}: failed — ${err.message}`);
+      failed.push(project.id);
     }
   }
+  return failed;
 }
 
 module.exports = { runAcMigration, migrateProject, convertAcRecord };
--- a/server/migrate-ac.js
+++ b/server/migrate-ac.js
@@ -1,0 +1,146 @@
+const fs = require("fs");
+const path = require("path");
+const os = require("os");
+const { ensureSecureDir, writeSecureFile } = require("./config");
+const { parseMentions } = require("./file-chat");
+
+const KNOWN_FIELDS = new Set([
+  "id", "timestamp", "ts", "sender", "channel", "type", "text", "message", "mentions",
+]);
+
+function convertAcRecord(record, nextId) {
+  let ts;
+  if (record.timestamp) {
+    ts = typeof record.timestamp === "number"
+      ? new Date(record.timestamp * 1000).toISOString()
+      : String(record.timestamp);
+  } else if (record.ts) {
+    ts = String(record.ts);
+  } else {
+    ts = new Date().toISOString();
+  }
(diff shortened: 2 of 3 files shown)


realproject7 (commented):
**RE2 — APPROVE** ✅

Solid migration implementation. Matches all acceptance criteria from #719.

**Verified:**
- **Conversion logic**: Unix epoch `timestamp * 1000` → ISO 8601, fallback chain (`timestamp` → `ts` → `new Date()`). Non-system types mapped to `"message"` with original preserved in `_legacy.type`.
- **Safety**: AC source is read-only — never modified. Write-to-tmp + validate-every-line + `fs.renameSync` gives atomic output. `0o600` permissions via `writeSecureFile` and `fs.chmodSync`.
- **Idempotency**: `.migrated` manifest prevents re-migration. Phase 1 projects with existing `general.jsonl` (no `.migrated`) correctly skipped with log message.
- **Boot ordering**: `runAcMigration(startupCfg)` placed after AC process tracking but before `fileChat.initProject()` — migration completes before file-chat engine takes the writer lock. Correct.
- **Dependencies**: `parseMentions` from `./file-chat` exists on main (line 30). `ensureSecureDir`/`writeSecureFile` from `./config` confirmed exported.
- **Unknown field preservation**: `KNOWN_FIELDS` set excludes AC-specific fields (`uid`, `time`, `attachments`, `reply_to`), which land in `_legacy`.
- **Tests**: 7 tests cover round-trip, invalid lines, idempotency, Phase 1 skip, and no-AC-log cases. `os.homedir` monkeypatch isolates to tmpdir.
- **System message**: Appended after migration with correct next sequential ID.

No blocking findings.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The happy-path converter matches the new file-chat schema, preserves legacy fields, and runs before file-chat initialization. However, the failure path does not satisfy the migration safety criteria: a failed migration is only logged, and boot can still initialize file-chat for that project.

### Findings
- [blocking] `runAcMigration` swallows per-project migration failures, and `server/index.js` proceeds to `fileChat.initProject` for any project already configured with `chat_mode: "file"`. If conversion or validation fails during an upgrade, the project can continue in file mode with missing/empty migrated history, violating the acceptance criterion that a failed migration leaves the project on AC mode with the error surfaced.
  - File: `server/migrate-ac.js:138`
  - File: `server/index.js:3046`
  - Suggestion: Have migration report failed project IDs or throw, then prevent file-chat initialization for those projects and keep/restore AC-mode behavior for them. At minimum, do not let a project enter file-chat after its migration failed.

- [medium] An AC log containing only invalid non-empty lines is treated as a migration failure instead of being skipped and counted. With `converted.length === 0` and `skipped > 0`, the code writes a blank temp file, then validates `verifyContent.trim().split("\n")` as `[""]` and throws while parsing the empty string. The issue says invalid lines should be skipped and counted, not fail migration.
  - File: `server/migrate-ac.js:82`
  - Suggestion: Handle the zero-valid-record case explicitly, still create a manifest/system marker or otherwise mark the project as processed with the skipped count, and add a test for an all-invalid source file.

### Decision
Requesting changes because the PR does not yet meet the failure-safety and invalid-line handling requirements for #719.

realproject7 (commented):

Both RE1 findings are properly fixed:

1. **Migration failure safety**: `runAcMigration` now returns failed project IDs. Server boot creates a `migrationFailed` Set and skips `fileChat.initProject` for those projects, reverting them to `chat_mode: "ac"` in memory. The revert is runtime-only (not persisted to config.json), so the next boot retries migration correctly — no `.migrated` or `general.jsonl` was written on failure, so `migrateProject` will re-attempt.

2. **All-invalid-lines**: New guard at migrate-ac.js:82-85 handles `converted.length === 0 && skipped > 0` by returning null with a log message. No tmp file is created, no throw. New test (line 166-180) verifies no artifacts are left behind.

No new findings. Previous RE2 approval points (conversion logic, atomic write, idempotency, boot ordering, dependency checks) still hold.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The all-invalid-lines crash is fixed, and `runAcMigration` now reports failed project IDs. One blocking failure-safety issue remains: the rollback to AC mode is only applied to the local `startupCfg` object, while the rest of the server reads `chat_mode` from the persisted config.

### Findings
- [blocking] On migration failure, `server/index.js` sets `p.chat_mode = "ac"` only in the in-memory `startupCfg` loop. Request handlers such as `/api/chat` call `getProjectChatMode`, which reloads `config.json` from disk and will still see `chat_mode: "file"`; agent startup paths also call `readConfig()` again. That means the project can still be treated as file-chat after file-chat init was skipped, which violates the requirement that failed migration keeps the project on AC mode.
  - File: `server/index.js:3046`
  - File: `server/routes.js:260`
  - Suggestion: Persist the AC-mode rollback with `writeConfig` when migration fails, or introduce a shared runtime failure override that `getProjectChatMode` and agent startup paths honor. Add a regression test if practical so failed migrations cannot leave disk config in file mode while runtime routes still use file-chat.

### Decision
Requesting changes because migration failure handling still does not reliably keep the project on AC mode across the server.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review blockers are resolved. Migration failures are now reported back to startup, file-chat initialization is skipped for failed projects, and the AC-mode rollback is persisted to `config.json` so request handlers and agent startup paths see the same mode.

### Findings
- None.

### Decision
Approving. The PR satisfies #719's migration requirements: AC logs are converted into the file-chat schema, unknown fields are preserved in `_legacy`, invalid lines are counted/skipped, AC source data is read-only, successful migrations are marked with `.migrated`, writes use tmp/validate/rename, and failed migrations keep the project on AC mode.

realproject7 (commented):
**RE2 — APPROVE (re-review #2)** ✅

RE1's persistence finding is fixed. The rollback now:
1. Sets `p.chat_mode = "ac"` in the in-memory startup object (skips `initProject`)
2. Re-reads config from disk via `readConfig()`, finds the project entry, sets `chat_mode: "ac"`, and persists via `writeConfig()`

This ensures `getProjectChatMode()` (routes.js:260) and all other paths that call `readConfig()` see the correct mode. The approach of re-reading config before writing is safe — it avoids clobbering any concurrent config changes between startup and this point.

No new findings. All prior RE2 approval points still hold.

PR #750, opened 2026-05-20
[#721] Reverse migration tool (ac-restore)

### LINKED ISSUE #721: [#713-8] Reverse migration tool (ac-restore)
## Summary
Ship `npx quadwork ac-restore` — a reverse migration tool that converts `general.jsonl` back into AC's `agentchattr_log.jsonl` format. Makes Phase 2 and Phase 3 reverts lossless.

## CLI command: `quadwork ac-restore`

```bash
npx quadwork ac-restore              # restore all projects
npx quadwork ac-restore --project X  # restore specific project
```

### Logic
For each project:
1. Read `~/.quadwork/<project>/chat/general.jsonl`
2. Convert each record back to AC's JSONL schema
3. Append to `~/.quadwork/<project>/agentchattr/data/agentchattr_log.jsonl`
4. Log: `[ac-restore] plottoon: restored 234 messages to AC format`

### Field mapping (reverse of migration)
```js
function convertToAcFormat(record) {
  const acRecord = {
    id: record.id,
    sender: record.sender,
    text: record.text,
    channel: record.channel,
    timestamp: record.ts,
    type: record.type,
    ...(record._legacy || {}),  // restore unknown fields
  };
  return acRecord;
}
```

### Safety
- Append-only to AC's JSONL — doesn't overwrite existing messages
- Duplicate detection: compute `sha1(ts|sender|channel|text)` content hash per record. Skip AC records whose hash matches an already-written record. More robust than string triple matching against round-trip formatting changes.
- Idempotent: safe to run multiple times
- Does not start AC or modify AC config — just writes data

## Implementation in `bin/quadwork.js`
Add `ac-restore` to the command switch:
```js
case "ac-restore":
  cmdAcRestore();
  break;
```

~50 lines for the conversion logic.

## Dependencies
- Requires #713-1 (understands the JSONL schema)

## Acceptance criteria
- [ ] `npx quadwork ac-restore` converts general.jsonl → AC JSONL format
- [ ] Appends to existing AC data (doesn't overwrite)
- [ ] Skips duplicates
- [ ] Works per-project and all-projects
- [ ] `npm run build` passes

## Safety note
This command is the ONLY code in QuadWork allowed to modify AC data during the pivot. Normal migration (#719) is read-only against AC files. This distinction must be preserved in all implementations.

## Idempotency marker
Each restored AC record includes `_quadwork_restored_id: <file-chat-id>`. On re-run, `ac-restore` skips any AC record with a matching marker. This prevents duplicates on repeated restore attempts.

### Tests
- Round-trip test: file JSONL → ac-restore → AC JSONL → migrate → file JSONL is field-stable
- Re-running ac-restore produces no duplicates (marker check)

realproject7:
## Summary
- New `server/ac-restore.js` converts file-chat `general.jsonl` back to AC's `agentchattr_log.jsonl` format
- CLI command `npx quadwork ac-restore` (all projects) and `npx quadwork ac-restore --project X` (specific project)
- Append-only to AC data — never overwrites existing messages
- Dual duplicate detection: `_quadwork_restored_id` marker + SHA-1 content hash
- Skips migration system messages ("Chat history migrated from AC...")
- Restores `_legacy` fields back to top-level AC record properties

## Test plan
- [x] convertToAcFormat maps fields correctly and restores _legacy
- [x] Basic restore writes AC JSONL
- [x] Re-running ac-restore produces no duplicates (idempotent)
- [x] Skips migration system messages
- [x] Round-trip AC → migrate → ac-restore preserves fields
- [x] No general.jsonl returns null
- [x] Content hash dedup prevents duplicates against existing AC records
- [x] `npm run build` passes

Closes #721

### DIFF
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -2268,12 +2268,23 @@ function cmdAcRestore() {
   const projectFlagIdx = args.indexOf("--project");
   const projectFilter = projectFlagIdx >= 0 ? args[projectFlagIdx + 1] : null;
 
+  if (projectFlagIdx >= 0 && (!projectFilter || projectFilter.startsWith("--"))) {
+    warn("--project requires a project ID. Usage: npx quadwork ac-restore --project <id>");
+    process.exit(1);
+  }
+
   if (projectFilter && !projectFilter.match(/^[\w-]+$/)) {
     warn("Invalid project ID.");
     process.exit(1);
   }
 
   const config = readConfig();
+
+  if (projectFilter && !(config.projects || []).some((p) => p.id === projectFilter)) {
+    warn(`Project '${projectFilter}' not found in config.`);
+    process.exit(1);
+  }
+
   const { runAcRestore } = require("../server/ac-restore");
   runAcRestore(config, projectFilter || null);
 }
--- a/bin/quadwork.js
+++ b/bin/quadwork.js
@@ -2261,6 +2261,23 @@ async function cmdMigrateAgentSlugs() {
   log("  3. Old chat messages keep their original sender — no history rewrite");
 }
 
+// ─── ac-restore ────────────────────────────────────────────────────────────
+
+function cmdAcRestore() {
+  const args = process.argv.slice(3);
+  const projectFlagIdx = args.indexOf("--project");
+  const projectFilter = projectFlagIdx >= 0 ? args[projectFlagIdx + 1] : null;
+
+  if (projectFilter && !projectFilter.match(/^[\w-]+$/)) {
+    warn("Invalid project ID.");
+    process.exit(1);
+  }
+
+  const config = readConfig();
+  const { runAcRestore } = require("../server/ac-restore");
+  runAcRestore(config, projectFilter || null);
+}
+
 // ─── Main ───────────────────────────────────────────────────────────────────
 
 const command = process.argv[2];
@@ -2287,6 +2304,9 @@ switch (command) {
   case "migrate-agent-slugs":
     cmdMigrateAgentSlugs();
     break;
+  case "ac-restore":
+    cmdAcRestore();
+    break;
   case undefined: {
     // #573: No subcommand — smart default based on config state.
     // Config file exists (even with 0 projects) → start, so the user
@@ -2311,6 +2331,7 @@ switch (command) {
     cleanup       Reclaim disk space (--project <id> or --legacy)
     doctor        Report the AgentChattr pin + per-project clone SHAs
     migrate-agent-slugs  Rename reviewer1/reviewer2 → re1/re2 in existing projects
+    ac-restore           Restore file-chat JSONL back to AC format
 
   Workflow:
     1. npx quadwork init     — one-time setup (installs prerequisites)
--- a/server/ac-restore.js
+++ b/server/ac-restore.js
@@ -1,0 +1,123 @@
+const fs = require("fs");
+const path = require("path");
+const os = require("os");
+const crypto = require("crypto");
+const { ensureSecureDir } = require("./config");
+
+function convertToAcFormat(record) {
+  const acRecord = {
+    id: record.id,
+    sender: record.sender,
+    text: record.text,
+    channel: record.channel,
+    timestamp: record.ts,
+    type: record.type,
+    _quadwork_restored_id: record.id,
+    ...(record._legacy || {}),
+  };
+  return acRecord;
+}
+
+function contentHash(record) {
+  const key = `${record.ts}|${record.sender}|${record.channel}|${record.text}`;
+  return crypto.createHash("sha1").update(key).digest("hex");
+}
+
+function restoreProject(projectId) {
+  const chatFile = path.join(os.homedir(), ".quadwork", projectId, "chat", "general.jsonl");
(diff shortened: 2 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The restore implementation is largely aligned with #721: it appends to AC JSONL, skips duplicates via marker/hash, restores `_legacy` fields, and has useful idempotency/round-trip tests. One CLI safety bug needs fixing before merge because this command writes to AC data.

### Findings
- [blocking] `ac-restore --project` without a following project ID silently runs against all projects. `projectFilter` becomes `undefined`, passes validation, then `runAcRestore(config, projectFilter || null)` receives `null`, which is the all-projects mode. A missing argument or shell-script variable expansion should fail closed, not modify every project's AC log.
  - File: `bin/quadwork.js:2268`
  - Suggestion: If `--project` is present, require a non-empty following value that is not another flag; otherwise print usage/error and exit non-zero. Consider also warning/erroring when the provided project ID does not match any configured project.

### Decision
Requesting changes because the per-project CLI path can accidentally trigger the all-project restore path.

realproject7 (commented):
**RE2 — REQUEST CHANGES**

## Verdict: REQUEST CHANGES

### Summary
The conversion logic, CLI integration, and append-only safety are correct. One blocking issue: content hash dedup fails for the primary round-trip workflow because timestamp format changes through the migration cycle.

### Findings
- [blocking] Content hash dedup produces different hashes for original vs. restored records when the original AC log is still present. The forward migration converts Unix epoch timestamps (`1777370000`) to ISO strings (`"2026-05-20T10:00:00.000Z"`). `ac-restore` maps this back as `timestamp: record.ts` (ISO string). The `contentHash` key is `${record.ts}|sender|channel|text`, so the original AC record hashes to `1777370000|user|general|hello` while the restored record hashes to `2026-05-20T10:00:00.000Z|user|general|hello` — different hashes, duplicate not detected, message appended alongside original.
  - The `_quadwork_restored_id` marker doesn't catch this either — original AC records don't have it.
  - The round-trip test masks this by emptying the AC log before restoring (`fs.writeFileSync(..., "")`).
  - File: `server/ac-restore.js:23` (`contentHash`), `server/ac-restore.test.js:143` (test empties AC log)
  - Suggestion: Normalize timestamps in the hash function (convert Unix epoch to ISO before hashing), or check both formats. Fix the round-trip test to NOT empty the AC log — let it verify dedup against original records.

### Decision
Requesting changes because the primary use case (migrate AC → file-chat → ac-restore with original AC log intact) produces duplicate messages.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The `--project` safety issue from my first review is fixed. One remaining blocker: duplicate detection does not work for the primary round-trip scenario where original AC records use Unix epoch timestamps and migrated file-chat records use ISO timestamps.

### Findings
- [blocking] Content-hash dedup includes the raw timestamp value, but #719 converts numeric AC `timestamp` values to ISO strings during migration. Existing AC records therefore hash `1777370000|user|general|...`, while restored file-chat records hash `2026-05-...|user|general|...`, so `ac-restore` appends duplicates next to the original AC records. `_quadwork_restored_id` does not help because original AC records do not contain that marker. The current round-trip test misses this because it empties `agentchattr_log.jsonl` before restore.
  - File: `server/ac-restore.js:21`
  - File: `server/ac-restore.test.js:122`
  - Suggestion: Normalize timestamps before hashing, e.g. convert numeric epoch seconds and ISO strings to the same canonical form, and update the round-trip/dedup test to leave the original AC log intact so restore proves it does not duplicate already-migrated AC messages.

### Decision
Requesting changes because `ac-restore` is not currently idempotent against existing pre-migration AC logs, which is a core safety requirement for this operator-run reverse migration tool.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review blockers are resolved. The CLI now fails closed for missing/unknown `--project` values, and content-hash dedup normalizes numeric AC epoch timestamps so existing pre-migration AC records match migrated ISO file-chat records.

### Findings
- None.

### Decision
Approving. The PR satisfies #721: `ac-restore` converts file-chat JSONL back to AC JSONL, appends without overwriting, skips duplicates via marker/hash, supports all-project and per-project modes, restores `_legacy` fields, skips migration system messages, and includes regression coverage for the timestamp dedup case.

realproject7 (commented):

Both blocking findings are now fixed:

1. **Content hash normalization**: New `normalizeTs()` function (ac-restore.js:21-24) converts numeric Unix epoch timestamps to ISO before hashing, using the same `new Date(ts * 1000).toISOString()` conversion as migrate-ac.js. String timestamps pass through unchanged. This ensures `contentHash` produces identical hashes for epoch `1777370000` and ISO `"2026-05-20T10:46:40.000Z"`.

2. **Round-trip test fixed** (test line ~130): No longer empties the AC log before restoring. Keeps original AC records intact and asserts `restoredAcLines.length === acLinesBefore` — directly verifying no duplicates are appended in the real-world workflow.

3. **CLI validation** (bin/quadwork.js): `--project` now validates non-empty, non-flag value and checks project exists in config. Good.

No remaining findings.

Issue #744, opened 2026-05-20
[#713] Discord bridge loops on file-chat projects — still polling AC endpoint


realproject7:
## Problem
Discord bridge forwards the same message infinitely to Discord when connected to a file-chat project. The bridge polls AC's `/api/messages` endpoint, but file-chat projects write messages to JSONL, not AC. The bridge reads stale/mismatched data and its `since_id` cursor never advances correctly.

## Root cause
The bridge (`bridges/discord/discord_bridge.py`) polls `{agentchattr_url}/api/messages` — AC's direct API. For file-chat projects, messages are in `general.jsonl`, not AC. Two possibilities:
1. The bridge reads from AC which has old/different messages → cursor mismatch → infinite loop
2. AC proxies to the JSONL but the response format differs from what the bridge expects → cursor never advances

### Option A: Disable bridge auto-start for file-chat projects (Phase 1 quick fix)
In the bridge auto-start logic, skip projects with `chat_mode: "file"`. Bridge adaptation (#720) handles the full fix in Phase 2.

```js
// In autoStartBridges / auto-bridge logic:
if (project.chat_mode === "file") return; // bridges not yet adapted
```

### Option B: Adapt bridge polling NOW (pull forward from #720)
Change the bridge to poll `/api/chat?project=X&since_id=N` instead of `/api/messages` when the project uses file-chat. More work but enables bridges during Phase 1 testing.

## Recommendation
**Option A for now.** The bridge works fine on AC-mode projects. Disable for file-chat projects until #747 (Node.js bridges) lands.

## Acceptance criteria
- [ ] Discord bridge does NOT auto-start for file-chat projects
- [ ] Telegram bridge same treatment
- [ ] AC-mode projects' bridges unaffected
- [ ] No infinite message loop on Discord
- [ ] `npm run build` passes

## Updates from review
- Log when skipping bridge auto-start: `[auto-bridge] <project>: skipped (file-chat mode)`
- Also block MANUAL bridge start for file-mode projects (not just auto-start). The dashboard Start button should show an error: "Bridges use the new Node.js module in file-chat mode. Use the built-in bridge instead."
- Hide or disable the manual Start button for file-mode projects in the bridge widget UI
- [ ] Manual bridge start blocked for file-mode projects with error message
- [ ] Bridge Start button hidden/disabled in dashboard for file-mode projects

Issue #722, opened 2026-05-19
[#713-9] Agent MCP config switchover to file-based chat


realproject7:
## Summary
Switch ALL projects from AC-mode to file-based chat. Phase 1 already built all the infrastructure — this ticket just flips every project to `chat_mode: "file"`.

## What Phase 1 already implemented
- `file-chat.js` — JSONL engine with append, read, IDs, cache (#714)
- `mcp-chat-shim.js` — per-agent stdio MCP proxy (#715)
- `pty-dispatcher.js` — identity-aware PTY injection with idle gating, coalescing, active suppression (#730, #736, #742)
- Dashboard `ChatPanel.tsx` — REST polling with `since_id` for file-mode projects (#716)
- Loop guard in message processing (#717)
- System messages for lifecycle events (#718)
- `chat_mode` flag routing in `buildAgentArgs()` and `routes.js`
- Watchdog skips file-mode projects (#732)

## What this ticket does
Set `chat_mode: "file"` for ALL projects on first boot after upgrade:

```js
// In startup migrations:
for (const p of projects) {
  if (p.chat_mode !== "file") {
    p.chat_mode = "file";
    changed = true;
  }
}
if (changed) writeConfig(cfg);
```

### Important: keep `chat_mode` field
Do NOT remove the `chat_mode` flag in this ticket. It stays through Phase 2 for revert safety — if Phase 2 fails, operator can set individual projects back to `"ac"`. Flag removal happens in Phase 3 (#723).

### Scheduled trigger path
The trigger already posts to `/api/chat` which routes to file-chat or AC based on `chat_mode`. No trigger code changes needed — the routing handles it.

### AC still runs but is unused
AC processes are still spawned on boot. They have no clients. This supports revert.

## Dependencies
- Requires #719 (migration complete — AC history in JSONL before switchover)
- Requires #747 (bridges rewritten in Node.js — otherwise bridges break on switchover)
- Requires #721 (ac-restore available — revert tool ready before switchover)

## Acceptance criteria
- [ ] All projects set to `chat_mode: "file"` on boot
- [ ] `chat_mode` field preserved (not removed)
- [ ] All agents communicate via JSONL
- [ ] All bridges (Node.js modules from #747) deliver messages to file-chat without AC
- [ ] Scheduled trigger works for all projects
- [ ] AC processes still run (for revert safety)
- [ ] Operator can revert individual projects by setting `chat_mode: "ac"`
- [ ] `npm run build` passes

PR #751, opened 2026-05-20
[#747] Rewrite Discord + Telegram bridges in Node.js

### LINKED ISSUE #747: [#713] Rewrite Discord + Telegram bridges in Node.js
## Summary
Rewrite both bridges from Python to Node.js. Eliminates the last Python dependency — QuadWork becomes a pure Node.js application with zero external runtime requirements beyond Node itself.

## Why
Python bridges existed because AC was Python. With AC removed, Python is only needed for two bridge scripts. Rewriting them removes the Python prerequisite entirely:
- No `python3` installation needed
- No `python3-venv` package
- No venv creation / pip install
- No SSL cert issues on macOS
- No separate bridge processes to spawn/manage/crash

### Current: `bridges/discord/discord_bridge.py` (~600 lines)
- Polls QuadWork API for new messages (adapted in #720)
- Forwards to Discord channel via `discord.py`
- Receives Discord messages, posts to QuadWork API
- Tracks cursor via JSON file

### New: `server/bridges/discord.js` (~400 lines)
- Use `discord.js` npm package (MIT, 150K+ weekly downloads)
- Runs as a module inside QuadWork's Node.js process — no separate process
- Same logic: poll chat API → forward to Discord, receive Discord → post to chat API
- Cursor tracking via same JSON file format (backward compatible)
- Config reads from existing `~/.quadwork/discord-<project>.toml`

### npm dependency
Add `discord.js` as an **optional dependency** (not required for users who don't use bridges):
```json
"optionalDependencies": {
  "discord.js": "^14.x"
}
```

### Current: cloned separately to `~/.quadwork/agentchattr-telegram/` (~300 lines Python)
- Same pattern as Discord but uses Telegram Bot API via `python-telegram-bot`

### New: `server/bridges/telegram.js` (~250 lines)
- Use `node-telegram-bot-api` npm package or raw HTTP (Telegram Bot API is simple REST)
- Same in-process module pattern as Discord
- Config reads from existing `~/.quadwork/telegram-<project>.toml`

## What gets deleted
- `bridges/discord/discord_bridge.py` — replaced by Node.js version
- `bridges/discord/requirements.txt` — no more pip
- All Python venv creation code for bridges
- Bridge clone/install logic for Telegram
- The `bridges/` folder from npm package (replaced by `server/bridges/`)

## What stays the same
- Bridge config format (TOML files with bot token, channel ID)
- Bridge auto-start/stop logic in `server/index.js`
- Dashboard bridge widgets (start/stop/status UI)
- Cursor tracking (JSON file, backward compatible)

## Migration for existing users
- Existing bridge configs work as-is (same TOML format)
- Python bridge processes are no longer spawned — Node.js modules take over
- Cursor files are compatible — no message replay on upgrade
- Users can delete `~/.quadwork/agentchattr-discord/` and `~/.quadwork/agentchattr-telegram/` venvs

## Acceptance criteria
- [ ] Discord bridge works in Node.js (send/receive messages)
- [ ] Telegram bridge works in Node.js (send/receive messages)
- [ ] No Python dependency anywhere in QuadWork
- [ ] Existing bridge configs (TOML) work without changes
- [ ] Bridge auto-start/stop works from dashboard
- [ ] Cursor files backward compatible (no message replay)
- [ ] `bridges/` folder removed from npm package
- [ ] Install guides updated — no Python mentioned
- [ ] `npm run build` passes

### Both bridge libraries as optionalDependencies
```json
"optionalDependencies": {
  "discord.js": "^14.x",
  "node-telegram-bot-api": "^0.66.x"
}
```

### In-process crash isolation
Bridge modules run inside QuadWork's Node.js process. An uncaught exception in discord.js or telegram library must NOT crash the server. Wrap all bridge entrypoints:
```js
try {
  await bridgeLoop();
} catch (err) {
  console.error(`[bridge] ${type} crashed: ${err.message}`);
  // Emit system message, mark bridge as stopped, do NOT rethrow
}
```
Wrap every async bridge entrypoint with `.catch()` that logs and marks the bridge stopped. The bridge module owns its own promise hygiene — no unhandled rejections should escape into the global handler.

### Cursor file backward compatibility
Document the Python cursor JSON format. Write a test that the Node.js reader parses an existing Python-written cursor fixture and produces the same `since_id`.

realproject7:
## Summary
- New `server/bridges/telegram.js` — in-process Telegram bridge using raw Bot API (no Python)
- New `server/bridges/discord.js` — in-process Discord bridge using `discord.js` npm package (no Python)
- Both bridges run inside QuadWork's Node.js process — no separate processes to spawn/manage
- Bidirectional: polls chat API → forwards to platform, receives platform messages → posts to chat API
- Crash isolation: all async entrypoints wrapped with `.catch()`, errors stored in `lastError`
- Cursor files backward compatible with Python bridges — same JSON format, same paths
- `discord.js` and `node-telegram-bot-api` added as `optionalDependencies`
- Removed `bridges/discord/discord_bridge.py` and `bridges/discord/requirements.txt`
- Server routes updated: `install` is now a no-op, `start`/`stop` use in-process bridge modules
- `bridge_installed` always returns `true` (Node.js bridges are always available)

## Test plan
- [x] Telegram reads Python-written cursor file (backward compat)
- [x] Discord reads Python-written cursor file (backward compat)
- [x] Missing cursor returns 0
- [x] isRunning/getLastError return correct defaults
- [x] `npm run build` passes

Closes #747


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The rewrite removes the Python bridge process path, but the new Node bridge path is not functionally equivalent yet. Core start and inbound-message flows are broken, and Telegram can silently drop outbound messages while still advancing its cursor.

### Findings
- [blocking] File-chat projects still cannot start either bridge. The start routes keep the #744 temporary guard, and `autoStartBridges` still skips `chat_mode: "file"`, so the new Node bridge cannot run on the projects it is meant to support before #722 moves all projects to file mode.
  - File: `server/routes.js:3311`
  - File: `server/routes.js:3573`
  - File: `server/index.js:1820`
  - Suggestion: Remove or revise the file-chat bridge block now that the built-in Node bridges poll `/api/chat`, and make auto-start work for configured file-chat projects.

- [blocking] Platform-to-QuadWork messages are posted with the wrong `/api/chat` contract. Both bridges send `{ project_id, sender, text }`, but `POST /api/chat` reads `req.query.project || req.body.project`, not `project_id`; in file-chat mode this falls through as a missing project, and in AC mode the route ignores client-supplied `sender` by design. Result: Discord/Telegram inbound messages are not reliably written as bridge messages.
  - File: `server/bridges/telegram.js:103`
  - File: `server/bridges/discord.js:137`
  - File: `server/routes.js:1140`
  - Suggestion: Use the supported project parameter/body field and provide an authenticated/server-side way for in-process bridges to append as `tg:<user>` / `dc:<user>` without reopening arbitrary sender impersonation.

- [blocking] Telegram outbound delivery failures are treated as success. `sendTelegram` returns `false` for non-2xx responses, but `pollLoop` ignores that return value and advances the cursor. Telegram Markdown parse errors or API failures will permanently skip messages.
  - File: `server/bridges/telegram.js:36`
  - File: `server/bridges/telegram.js:59`
  - Suggestion: Throw on non-OK Telegram responses or check the boolean before marking forwarded/updating the cursor; consider avoiding `parse_mode: "Markdown"` unless messages are escaped.

### Decision
Requesting changes because the PR does not yet satisfy the send/receive bridge requirements or the file-chat bridge startup path for #747.

realproject7 (commented):
**RE2 — REQUEST CHANGES**

## Verdict: REQUEST CHANGES

### Summary
The bridge rewrite is structurally sound — in-process modules, proper crash isolation via `.catch()`, cursor backward compat, and massive simplification of routes.js. One blocking issue with Telegram message delivery.

### Findings
- [blocking] `sendTelegram` uses `parse_mode: "Markdown"`, which causes the Telegram API to return 400 for messages containing unescaped Markdown special characters (`_`, `*`, `[`, `]`, etc.). Agent messages routinely contain code with underscores (e.g., `my_var_name`, `file_path`). Additionally, the return value of `sendTelegram` is not checked in `pollLoop` — when the API rejects a message, the cursor still advances and the message ID is added to `forwardedIds`, so the message is permanently lost with no error surfaced.
  - File: `server/bridges/telegram.js:33` (`parse_mode: "Markdown"`)
  - File: `server/bridges/telegram.js:62` (return value ignored)
  - Suggestion: Either remove `parse_mode` entirely (plain text is safe and reliable), or switch to `parse_mode: "HTML"` and escape the 4 HTML special chars. Also: if `sendTelegram` returns `false`, don't advance the cursor/forwardedIds — throw or retry so the message isn't silently dropped.

### Decision
Requesting changes because Telegram message forwarding will silently drop messages containing common code patterns.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The previous file-chat startup and Telegram delivery-loss blockers are fixed. One blocking compatibility issue remains: inbound bridge sender attribution only works in the file-chat branch, so AC-mode projects still receive Discord/Telegram messages as the operator instead of the bridge sender.

### Findings
- [blocking] `X-Bridge-Sender` is only honored inside the `getProjectChatMode(projectId) === "file"` branch. For AC-mode projects, `POST /api/chat` intentionally ignores client-supplied sender and builds `message.sender` from `operator_name`, so inbound Discord/Telegram messages are posted to AgentChattr as the operator rather than `dc:<user>` / `tg:<user>`. #747 says existing bridge configs continue to work and #722 has not yet converted all projects to file mode, so AC-mode bridge receive must remain correct during this transition.
  - File: `server/routes.js:1142`
  - File: `server/routes.js:1203`
  - File: `server/bridges/telegram.js:103`
  - File: `server/bridges/discord.js:137`
  - Suggestion: Apply the same trusted in-process bridge sender path to the AC branch before constructing the websocket payload, while preserving the existing rule that ordinary dashboard clients cannot choose arbitrary senders. Also account for localhost address forms such as `127.0.0.1`, `::1`, or `::ffff:127.0.0.1` if using `req.ip`.

### Decision
Requesting changes because the Node bridges are not yet bidirectionally compatible with existing AC-mode projects.

realproject7 (commented):

All findings fixed. Additional changes are well-considered:

**My blocking fix (verified):**
- `parse_mode: "Markdown"` removed — plain text now, no silent drops
- `sendTelegram` throws on non-2xx (`throw new Error("Telegram sendMessage ${res.status}")`), so `pollLoop`'s catch block fires and cursor does NOT advance — messages are retried next poll

**Additional fixes (verified):**
1. **File-chat bridge block removed**: Guards removed from both Telegram/Discord routes, `autoStartBridges`, and both bridge widget UIs. Bridges now work for all projects.
2. **Bridge sender attribution**: `X-Bridge-Sender` header with `req.ip === "127.0.0.1"` validation — secure, prevents external spoofing.
3. **Echo filter extended**: Both bridges skip `tg:*`/`dc:*` prefixed senders in addition to legacy names — prevents echo loops from bridge-originated messages.
4. **UI widgets**: `chatMode` prop removed from bridge widgets and `OperatorFeaturesPanel` since the guard is no longer needed.

No blocking findings.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review blockers are resolved. File-chat bridge startup is enabled, inbound messages use the supported `/api/chat` project parameter, bridge sender attribution is handled in both file-chat and AC-mode paths, and Telegram no longer advances its cursor on failed sends.

### Findings
- None.

### Decision
Approving. The PR now satisfies the core #747 requirements for in-process Node.js Telegram/Discord bridges, backward-compatible cursor reads, dashboard start/stop/status integration, and removal of the Python bridge runtime path.

realproject7 (commented):
**RE2 — APPROVE (re-review #2)** ✅

RE1's AC-mode bridge sender finding is fixed. Verified:
- `isLocalhost()` covers `127.0.0.1`, `::1`, `::ffff:127.0.0.1` — all Node.js localhost forms
- File-chat branch updated to use `isLocalhost()` instead of bare string comparison
- AC-mode branch now checks `X-Bridge-Sender` + `isLocalhost(req.ip)` before overriding `operatorSender` — bridges get proper attribution, external requests cannot spoof

All prior RE2 approval points still hold. No new findings.

PR #815, opened 2026-05-29
[#797] Self-heal: auto-restart claude agents bricked by thinking-block API 400

### LINKED ISSUE #797: Self-heal: auto-restart claude agents bricked by Opus 4.8 thinking-block API 400
## Problem
On Opus 4.8, a `claude` agent can get **permanently bricked** by an unrecoverable Anthropic API 400:
```
API Error: 400 messages.N.content.M: 'thinking' or 'redacted_thinking' blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response.
```
Once the conversation hits this state, **every** subsequent turn 400s instantly ("Crunched for 0s") and the agent does nothing — but QuadWork still shows it "online." Confirmed in production on `plotlink-ows` / `dev`.

### Root cause (for context — NOT what this ticket fixes)
- VPS Claude Code updated → default model is now Opus 4.8, which emits **signed thinking blocks through tool-use turns**. QuadWork launches `claude` agents with no `--model`, so they inherit this default.
- The strict thinking-block continuation protocol gets broken when input lands mid-tool-cycle (the PTY dispatcher's 5s-no-output idle gate can't see the thinking phase, so it can inject at the wrong moment). Once broken, the session is unrecoverable.
- Evidence: fresh all-4.8 session, `dev` wedged while `re2` (identical model/config) healthy → it's 4.8 **+** an intermittent timing trigger, not 4.8 alone.

**The deeper fix (turn-aware PTY injection) is a SEPARATE, larger ticket — explicitly out of scope here.** This ticket is the cheap, low-risk safety net only.

## Scope of THIS ticket: self-heal auto-restart
When an agent's PTY emits the thinking-block 400, automatically restart that agent's session (which clears the corrupted conversation and resumes normally). A bricked agent is doing zero work and its conversation is already unrecoverable, so a restart **loses nothing** — the fresh session re-reads chat via `chat_read({since_id})` on the next trigger.

### Detection (inside the EXISTING `term.onData` handler, `server/index.js` ~line 498)
The handler already runs on every PTY chunk and sets `session.lastOutputAt`. Add — wrapped in its own `try/catch` so a detector bug can never break the PTY data pipeline:
- Match the specific signature conservatively: data contains `cannot be modified` AND `thinking`. (This string is Anthropic-API-specific; `codex` agents never emit it, so no agent-type gate is strictly needed, but you may additionally restrict to `claude`-CLI agents for clarity.)
- Do NOT consume, transform, or block the stream — observe only. The chunk must still flow to xterm exactly as today.

### Recovery (reuse the EXISTING restart sequence)
Extract the restart body from the `POST /api/agents/:project/:agent/restart` route (index.js ~716–733) into a small shared helper `restartAgentSession(key, { reason })` and call it from BOTH the route and the detector. The helper is exactly the current sequence — no new lifecycle logic:
```js
existing._suppressLifecycleMsg = true;
await stopAgentSession(key);
const result = await spawnAgentPty(project, agent, { suppressLifecycleMsg: true });
```
Fire-and-forget from `onData` with a re-entrancy guard (see below) — never `await` inside the data handler.

### Restart-loop guard (REQUIRED — prevents a recovery storm)
Per-agent state `{ lastAutoRestartAt, countInWindow }`:
- **Cooldown:** ignore a detection if an auto-restart for this agent happened < 60s ago (also covers the burst of repeated 400 lines from one wedge).
- **Re-entrancy:** a `session._autoRecovering` flag set before stop and cleared after spawn; ignore detections while set.
- **Circuit breaker:** at most 3 auto-restarts per agent per 15 min. If exceeded, STOP auto-restarting, emit a visible system message ("`<agent>` repeatedly hitting the thinking-block API error — auto-recovery paused, manual attention needed"), and log it. This prevents masking a deeper failure with an infinite restart loop.

### Observability
- `console.log("[self-heal] <project>/<agent>: thinking-block 400 detected — restarting session")`
- `emitSystemMessage(project, "<agent> auto-restarted (recovered from thinking-block API error)")` so it's visible in chat / dashboard.

## MUST NOT touch (core-process safety constraints)
Purely additive observation + reuse of the existing restart path. Do not modify or risk:
- file-chat JSONL engine / single-writer, MCP shim, loop guard, scheduled trigger, bridges
- the PTY → xterm streaming (detector observes, never consumes/alters the chunk)
- `spawnAgentPty` / `stopAgentSession` internals — call them, don't change them
- the watchdog (#732) and its file-mode gating
- agent model resolution / `buildAgentArgs`
No new dependencies. No change to any HTTP route's existing behavior (the restart route just delegates to the new shared helper with identical effect).

## Acceptance criteria
- [ ] Detector fires only on the thinking-block 400 signature; never on normal output (verify against a transcript of normal agent chatter)
- [ ] On detection, the agent session is restarted via the shared helper (same effect as the manual restart button)
- [ ] Cooldown (60s) + re-entrancy flag prevent multiple restarts from one wedge's repeated error lines
- [ ] Circuit breaker stops after 3 restarts / 15 min and emits the visible "manual attention" system message
- [ ] PTY output still streams to the terminal unchanged; a thrown detector error cannot break the data pipeline (wrapped in try/catch)
- [ ] Manual restart route still works and now shares the same helper
- [ ] `npm run build` passes

## Tests
- Feed a fake PTY data chunk containing the 400 signature → asserts `restartAgentSession` is invoked once; a second chunk within 60s → NOT invoked again (cooldown).
- Feed normal output (including text that merely contains the word "thinking") → NOT invoked.
- Simulate 4 detections spaced to defeat cooldown → 4th is suppressed by the circuit breaker and the "manual attention" system message is emitted.
- Detector throws (mock) → PTY data still forwarded; no crash.

## Out of scope (file separately if desired)
- **Turn-aware PTY injection** — the actual prevention (never inject mid-turn; replace the 5s-output idle gate with a real "idle at prompt" signal). Larger change, higher design cost; keep it out of this safety-net ticket.
- Pinning claude agents to an explicit `--model` — config workaround, not needed if 4.8 is desired.

realproject7:
Closes #797.

## What
Cheap, low-risk safety net for the Opus 4.8 thinking-block API 400 that permanently bricks a `claude` agent (every subsequent turn 400s instantly while QuadWork still shows it "online"). The deeper fix (turn-aware PTY injection) is explicitly out of scope.

## How
- **Detection** — observe-only inside the existing `term.onData` handler (`server/index.js`). Matches the conservative signature (chunk contains both `cannot be modified` AND `thinking`). Wrapped in its own `try/catch`; the chunk is never consumed or altered, so the PTY → xterm pipeline is untouched.
- **Recovery** — extracted the manual restart route body into a shared `restartAgentSession(key, { reason })` helper (identical effect). Both the route and the detector call it. The detector fires-and-forgets, never `await`s inside `onData`.
- **Guards** (`server/self-heal.js`, pure/testable):
  - 60s **cooldown** (also absorbs the burst of repeated 400 lines from one wedge)
  - **re-entrancy** flag (`session._autoRecovering`) set before stop, cleared after spawn
  - **circuit breaker**: ≤3 auto-restarts per agent per 15 min, then pauses and emits a visible `… auto-recovery paused, manual attention needed` system message once
- **Observability** — `[self-heal]` console logs + `… auto-restarted (recovered from thinking-block API error)` system message in chat.

## Safety
Purely additive: file-chat engine, MCP shim, loop guard, scheduled trigger, bridges, the PTY→xterm stream, `spawnAgentPty`/`stopAgentSession` internals, the watchdog, and agent model resolution are all untouched. The manual restart route keeps identical external behavior (now delegates to the shared helper).

## Tests
`server/self-heal.test.js` (16 assertions, all pass):
- signature matches the real 400 line; never matches prose containing "thinking" or "cannot be modified" alone
- detection restarts once; a second chunk within 60s is suppressed by cooldown
- re-entrancy guard suppresses while a restart is in flight
- 4 detections spaced to defeat cooldown → exactly 3 restarts, 4th tripped by breaker, message emitted exactly once
- a throwing callback cannot propagate out of the detector

`npm run build` passes; full existing server test suite still green.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -11,6 +11,7 @@ const routes = require("./routes");
 const fileChat = require("./file-chat");
 const { dispatchToAgentPTY, cleanupSession: cleanupPtyDispatcher } = require("./pty-dispatcher");
 const { runAcMigration } = require("./migrate-ac");
+const selfHeal = require("./self-heal");
 
 const net = require("net");
 const config = readConfig();
@@ -502,6 +503,37 @@ async function spawnAgentPty(project, agent, opts = {}) {
       if (session.scrollback.length > SCROLLBACK_SIZE) {
         session.scrollback = session.scrollback.slice(-SCROLLBACK_SIZE);
       }
+
+      // #797: observe-only self-heal detector. Wrapped in its own try/catch so
+      // a detector bug can never break the PTY → xterm pipeline; the chunk is
+      // never consumed or altered (the WS viewer forwards it independently).
+      try {
+        selfHeal.observeChunk(key, data, {
+          now: Date.now(),
+          recovering: !!session._autoRecovering,
+          onRestart: () => {
+            session._autoRecovering = true;
+            console.log(`[self-heal] ${key}: thinking-block 400 detected — restarting session`);
+            restartAgentSession(key, { reason: "thinking-block-400" })
+              .then((result) => {
+                if (result && result.ok) {
+                  emitSystemMessage(project, `${agent} auto-restarted (recovered from thinking-block API error)`);
+                }
+              })
+              .catch((err) => console.error(`[self-heal] ${key}: auto-restart failed:`, err.message))
+              .finally(() => {
+                const s = agentSessions.get(key);
+                if (s) s._autoRecovering = false;
+              });
+          },
+          onBreaker: (message) => {
+            console.log(`[self-heal] ${key}: ${message}`);
+            emitSystemMessage(project, message);
+          },
+        });
+      } catch (err) {
+        console.error(`[self-heal] ${key}: detector error (ignored):`, err.message);
+      }
     });
 
     term.onExit(({ exitCode }) => {
@@ -713,9 +745,12 @@ app.post("/api/agents/:project/:agent/stop", async (req, res) => {
 
 // --- Lifecycle: restart ---
 
-app.post("/api/agents/:project/:agent/restart", async (req, res) => {
-  const { project, agent } = req.params;
-  const key = `${project}/${agent}`;
+// #797: shared restart sequence, used by both the manual restart route and
+// the self-heal detector. Exactly the prior route body — no new lifecycle
+// logic. The `reason` is informational (logged by callers).
+async function restartAgentSession(key, { reason } = {}) {
+  const [project, agent] = key.split("/");
+  console.log(`[restart] ${key}: restarting session (reason: ${reason || "unspecified"})`);
 
   // #241: must await deregister before respawn so the slot frees and
   // the fresh register lands at slot 1 instead of head-2.
@@ -723,7 +758,14 @@ app.post("/api/agents/:project/:agent/restart", async (req, res) => {
   if (existing) existing._suppressLifecycleMsg = true;
   await stopAgentSession(key);
 
-  const result = await spawnAgentPty(project, agent, { suppressLifecycleMsg: true });
+  return spawnAgentPty(project, agent, { suppressLifecycleMsg: true });
+}
+
+app.post("/api/agents/:project/:agent/restart", async (req, res) => {
+  const { project, agent } = req.params;
+  const key = `${project}/${agent}`;
+
+  const result = await restartAgentSession(key, { reason: "manual" });
   if (result.ok) {
     emitSystemMessage(project, `${agent} restarted`);
     res.json({ ok: true, state: "running", pid: result.pid });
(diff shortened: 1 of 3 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR matches issue #797: it adds an observe-only detector in the existing PTY data path, reuses the manual restart sequence through a shared helper, and implements cooldown, re-entrancy, and circuit-breaker guards with focused tests.

### Findings
- No blocking findings.

### Decision
Approve. The change is narrow and additive around the requested self-heal behavior, preserves the PTY output flow, and keeps recovery on the existing restart path. Note: GitHub reports no checks for this branch, so this review is based on static inspection and the PR's stated local test/build results.

Issue #807, opened 2026-05-29
[#805] GITHUB.md template + seed + endpoints + parser + test


realproject7:
Part of #805 (step 2).

## Goal
Add a server-authored, machine-generated `GITHUB.md` per project (mirrors OVERNIGHT-QUEUE.md) + a strict parser + a read endpoint. The server (#806) is the sole author of machine sections; agents read it for discovery.

## Reviewer attribution — body-marker ROLE, NOT author/login
re1 and re2 review as the **same** `{{reviewer_github_user}}` (`templates/seeds/re1.AGENTS.md` + `re2.AGENTS.md`), so GitHub login cannot distinguish them and "distinct-per-author" would collapse both into one. Attribution is **latest-per-body-marker-ROLE across the shared author's full review list**: match `RE1/RE2/Reviewer1/Reviewer2/T2a/T2b` (and `## Verdict`) prefixes in the review **body**; a marker-less/ambiguous review is **uncounted** (under-count, never over-count). Requires #806's PR data to include review `body`. (True login-based attribution needs separate reviewer identities — see #811, closed/not-planned.)

## Scope
- `templates/GITHUB.md` — `{{project_name}}`/`{{repo}}` + fixed `## ` headers: Open Issues, Open PRs, Recently Closed Issues, Recently Merged PRs, Review Detail, Notes.
- `writeGithubFileSafe(...)` mirroring `writeOvernightQueueFileSafe`; wire into both add-config sites (~routes.js:2173 re-seed, ~2212 first-time).
- **Build from the structured snapshot (#806), injection-safe:** assemble sections from the fetcher's structured object, and **fence/escape untrusted issue/PR body text** so a body can never emit a `## ` section marker.
- **Atomic write:** one completed-snapshot temp-file + `rename` (readers never see partial content); **preserve the `## Notes` block** (HEAD-annotated, parser-ignored).
- **Freshness marker:** stamp `generatedAt` + `staleCycles`/`_stale` in the file header AND `/api/github-parsed`, surfacing the ts/stale flag the cache already tracks (`routes.js:115/135`). Note: `adaptiveTTL` serves *expired* data when rate-limited (→ Infinity), so staleness can be unbounded — it MUST be visible.
[...]

Issue #805, opened 2026-05-29
[Epic] Move GitHub data off GraphQL onto server-side REST+ETag cache (GITHUB.md)


realproject7:
## Vision
Move QuadWork's GitHub data acquisition off the **GraphQL** budget (saturated) onto the **idle REST + Search budgets**, using a server-maintained cache + a machine-generated `GITHUB.md` projection that agents read instead of each calling `gh`.

## Root cause (measured on production VPS)
- 5 agent teams = **20 agents** share one GitHub token. Each runs `gh pr list` (≈**13 GraphQL pts**) / `gh issue list` on every trigger cycle for *discovery*.
- Measured drain: **~10,750 GraphQL pts/hr** vs the 5,000/hr budget → exhausted every rolling hour.
- QuadWork's own dashboard polling is negligible (batched board query = **3 pts**). REST core sits idle (~4994/5000); Search (30/min) untouched.

## Measured costs of the alternatives
| Method | Cost | Use |
|---|---|---|
| ETag / `If-None-Match` → 304 | **0 pts** | wrap every server REST poll |
| REST list (`gh api .../pulls\|issues`) | **1 core pt** | open PRs, open/closed issues |
| Search (`review:approved`, `is:merged`) | **1 search pt** | PR approval state (REST can't give it) |
| field-limited `--json` | only helps by dropping connections; still GraphQL | last-resort fallback |

## Architecture (server-authoritative)
The **server** is the source of truth: a background poller fetches cheaply (REST+ETag+Search) and writes both (a) the existing `_ghEndpointCache` the dashboard already reads, and (b) `~/.quadwork/{project}/GITHUB.md` (machine sections + a parser-ignored `## Notes` HEAD may annotate). Agents read the file/endpoint for **discovery**; all **writes** (`gh pr merge`/`create`/`review`, `gh issue create`) stay direct with the correct tokens. Reviewers' `gh pr diff`/`view` stay direct & uncached (no stale-code reviews).

## Sub-tickets
1. Server REST+ETag fetcher behind the existing cache
2. `GITHUB.md` template + seed + endpoints + parser + test
3. Tighten trigger message (RE1/RE2 review only @mentioned PRs)
4. Seed updates — agent discovery reads → file
5. Batch-progress off GraphQL
~~(optional) Frontend reads server-resolved re1/re2~~ - CLOSED not-planned (#811): re1/re2 share one GitHub identity, so login-based resolution is infeasible without separate reviewer identities.

[...]

Issue #802, opened 2026-05-29
GraphQL polling backs off on REST budget, not graphql — gate refreshGraphQLCache on the GraphQL bucket


realproject7:
## Problem
QuadWork's background GraphQL polling drains the GitHub **GraphQL** rate-limit budget and never backs off, because its backoff guard watches the **wrong** rate-limit bucket. GitHub tracks REST and GraphQL as **separate** budgets, so GraphQL can hit 0 while REST is full — and the guards keep issuing GraphQL queries.

Observed in production: GraphQL 0/5000, REST 5000/5000.

## Root cause (verified against `origin/main`)
- `refreshRateLimit()` (`routes.js:48-63`) fetches **only** `.resources.core` (REST); `isRateLimited()` (`routes.js:71`) checks that REST budget.
- **Two** GraphQL streams are gated on that REST-only signal:
  1. `refreshGraphQLCache()` — `if (isRateLimited()) return;` (`routes.js:1079`).
  2. `/api/batch-progress` (`routes.js:1801`) — only serves cache when `isRateLimited() && cached` (`routes.js:1812`); otherwise it calls `fetchBatchProgressGraphQL()` **unconditionally** (`routes.js:1866`), whose query also uses `reviews(last: 100)` (`routes.js:1123`).
- So when GraphQL is exhausted but REST is full, **both** streams keep hitting GraphQL.

Note: the GraphQL "5000" is a **points/hour** budget, not request count.

## Fix — cover BOTH GraphQL streams
1. In `refreshRateLimit()`, also fetch + store `.resources.graphql` (one `gh api rate_limit` call returns all resources).
2. Add `isGraphqlRateLimited()` / `isGraphqlRateLow()` mirroring the REST checks.
3. Gate **`refreshGraphQLCache()`** (`routes.js:1079`) on the GraphQL budget.
4. Gate the **`/api/batch-progress` GraphQL path** (`routes.js:1812`/`1866`) on the GraphQL budget too: when GraphQL is low/critical, serve the existing `_batchProgressCache` (even stale) or fall back to the REST gh-CLI path (the #416 parallel approach already present), instead of calling `fetchBatchProgressGraphQL()`.
5. When exhausted, skip/serve-stale and resume after the GraphQL `resetAt`.

## Must not change
- Module-scope auto-start of polling (`routes.js:2739/2741`).
- Poll cadence/TTL and the batched-query shapes.
- REST-side gating behavior (additive change for the GraphQL bucket).

[...]

PR #818, opened 2026-05-29
[#806] Acquire GitHub board state via REST + ETag, off the GraphQL budget

### LINKED ISSUE #806: [#805] Server REST+ETag fetcher behind the existing GitHub cache
Part of #805 (step 1).

## Goal
Acquire all dashboard/agent GitHub state via **REST + ETag + Search** (idle budgets), feeding the existing `_ghEndpointCache`, and **define + test a canonical dashboard shape**. This removes QuadWork's GraphQL usage. Main budget win.

## Scope (server/routes.js)
- **`githubStateFetcher(repo)`** — all calls conditional with stored `If-None-Match` (on **304 reuse cache**; `gh api` exits non-zero on 304 → treat non-zero + 304 header as "unchanged", not error):
  - `gh api repos/{o}/{r}/pulls?state=open`
  - `gh api repos/{o}/{r}/issues?state=open` and `?state=closed&per_page=100` — drop items with a `pull_request` key
  - per open PR: `gh api repos/{o}/{r}/pulls/{n}/reviews` → the **full review list** (each review's `state`, `author{login}`, `submittedAt`, `body`). Do **not** collapse to per-author-latest -- re1/re2 share one GitHub author, so the raw list with bodies must reach #807/#810 for body-marker ROLE attribution
  - per open PR head sha: build `statusCheckRollup` from **both** `/commits/{sha}/check-runs` **and** `/commits/{sha}/status` (combined status), normalized
- **ETag store + 304 handling for EVERY resource** (pulls, open issues, closed issues, per-PR reviews, check-runs/status, merged list) — not just top-level lists — plus `GH_MAX_CONCURRENT` so per-PR calls don't drain the core budget.
- **Replace the `gh pr list`/`gh issue list` args** in `/api/github/issues`, `/api/github/prs`, `/api/github/merged-prs`, `/api/github/closed-issues`, and the `/api/github/all` fallback so cache-miss/stale refresh goes through `githubStateFetcher`, **not** `gh pr list`. The GraphQL `gh pr list` path may remain **only** as an explicit emergency fallback, gated by the **GraphQL** budget guard (`isGraphqlRateLimited()` from #802), never by REST `isRateLimited()`.
- **Search** (`search/issues?q=..review:approved`) is an **optional optimization only**; the per-PR `/pulls/{n}/reviews` data is authoritative for approval state. Default path does not require Search.

## Canonical shape (define explicitly + test — do NOT defer to "same as GraphQL")
The current GraphQL transform lowercases issue/PR state to `"open"` while `GitHubPanel.tsx:106` colors on `state === "OPEN"` (and the gh-CLI path returns `"OPEN"`) — an existing inconsistency. Pin the canonical shape: issue/PR `state` = uppercase `OPEN`/`CLOSED`/`MERGED`; fields `number,title,state,url,author{login},assignees,reviewDecision,reviews[{state,author{login},submittedAt,body}],statusCheckRollup,createdAt`; closed/merged capped at 5 desc. Add a test asserting `GitHubPanel` (incl. the open-state dot) and batch-progress render correctly from this shape.

## Acceptance criteria
- [ ] Steady-state GraphQL usage from QuadWork ≈ 0 across the board poll AND all per-endpoint paths
- [ ] No code path calls `gh pr list`/`gh issue list` except an explicit emergency fallback gated by the **GraphQL** budget (`isGraphqlRateLimited`), not REST `isRateLimited()`
- [ ] Every REST resource (pulls, open/closed issues, per-PR reviews, check-runs/status, merged) is ETag/304-cached and concurrency-limited
- [ ] Canonical shape is defined and a test asserts GitHubPanel + batch-progress render correctly (incl. uppercase `OPEN` issue dot)
- [ ] `statusCheckRollup` built from check-runs **and** combined status
- [ ] `npm run build` passes

---

## Workflow-risk guardrails (from #805 risk analysis)
- Expose whether the last fetch **304'd vs failed vs succeeded**, so the freshness timestamp written into GITHUB.md (#807) is meaningful (never stamp "fresh" on a failed cycle).
- Feed the file writer (#807) a **single fully-assembled snapshot** from one completed fetch pass — not section-by-section as REST calls return — so readers never see a half-updated mix.

realproject7:
Closes #806. Part of #805 (step 1).

## Problem
The dashboard board was fed by `gh pr list`/`gh issue list --json reviewDecision,statusCheckRollup,...`. Those `--json` fields are **GraphQL-backed under the hood**, so every board poll drained GitHub's GraphQL points budget (the same budget #802 had to defend). This replaces that data path with plain REST + conditional requests, which use the idle core budget and cost ≈0 in steady state.

## What changed (server/routes.js)
- **`githubStateFetcher(repo)`** — assembles one snapshot per pass via `gh api repos/...`:
  - `pulls?state=open`, `issues?state=open`, `issues?state=closed` (drop items with a `pull_request` key), `pulls?state=closed` (→ merged, filtered on `merged_at`).
  - per open PR: `/pulls/{n}/reviews` (the **full** list with `state/author{login}/submittedAt/body` — re1/re2 share one GitHub author, so #807/#810 attribute ROLE from body markers, never an author-latest collapse).
  - per open PR head sha: `statusCheckRollup` built from **both** `/commits/{sha}/check-runs` **and** the combined `/commits/{sha}/status`, normalized to `{state}` (SUCCESS/FAILURE/ERROR/PENDING).
- **ETag/304 on every resource.** Stored `If-None-Match`; `gh api -i` exits non-zero on 304 with `stderr: "gh: HTTP 304"` — detected and the cached payload reused (zero budget). Per-PR fan-out is concurrency-limited (`GH_MAX_CONCURRENT`).
- **Trustworthy fetch status** — `"ok"` (something changed) / `"unchanged"` (all 304s) / `"error"` (couldn't assemble any list) so #807 never stamps a failed cycle as fresh, and the snapshot is emitted fully-assembled (never section-by-section).
- **Canonical shape pinned + tested** — issue/PR `state` is UPPERCASE `OPEN`/`CLOSED`/`MERGED` (matches `GitHubPanel.tsx:106` `state === "OPEN"` and the gh-CLI path; the old GraphQL transform lowercased to `"open"` — that inconsistency is fixed). Fields: `number,title,state,url,author{login},assignees,reviewDecision,reviews[{state,author{login},submittedAt,body}],statusCheckRollup,createdAt`; closed/merged capped at 5 desc.
- **Rewired** `/api/github/{issues,prs,closed-issues,merged-prs}` (shared `serveGithubList`), the `/api/github/all` fallback, and `/api/projects` activity stats to the REST snapshot. **No code path calls `gh pr list`/`gh issue list` anymore.**
- **Emergency fallback only:** the batched GraphQL query (`fetchAllProjectsGraphQL`) is kept solely as a degraded fallback, gated on the **GraphQL** budget (`isGraphqlRateLimited`, #802), never REST. Its transform is pinned to the same uppercase canonical shape.
- Removed the now-dead `cachedGhEndpoint` + background-refresh-queue helpers.

## Acceptance criteria
- [x] Steady-state GraphQL usage ≈ 0 across the board poll and all per-endpoint paths (REST + 304s)
- [x] No `gh pr list`/`gh issue list` except the emergency GraphQL fallback gated by `isGraphqlRateLimited`
- [x] Every REST resource (pulls, open/closed issues, per-PR reviews, check-runs/status, merged) is ETag/304-cached and concurrency-limited
- [x] Canonical shape defined + a test asserts GitHubPanel (incl. the uppercase `OPEN` dot) and batch-progress render correctly from it
- [x] `statusCheckRollup` built from check-runs **and** combined status
- [x] `npm run build` passes

## Tests / verification
- New `server/routes.githubCanonicalShape.test.js` — **32 assertions**. Feeds REST fixtures through the transforms and drives **verbatim-copied** GitHubPanel predicates (`issueStatusColor`/`reviewColor`/`ciColor`/`ciLabel`, line refs in the test) + batch-progress state comparisons, so the shape and renderer can't silently drift.
- **End-to-end against the live repo:** cold pass → `ok` (uppercase states, 5 closed / 5 merged), warm pass → `unchanged` (every resource 304'd, zero budget). Per-PR reviews/check-runs/status parsing confirmed against a real PR.
- `npm run build` green; existing `routes.*` tests pass.

## Notes for reviewers
- Out of #806's enumerated scope and left as-is: `/api/batch-progress`'s own GraphQL path (`fetchBatchProgressGraphQL`, separately budget-gated in #802) and its `gh issue view`/`gh pr view` REST fallback (`gh ... view`, not `gh ... list`).
- The poller function keeps the name `refreshGraphQLCache`/`startGraphQLPolling` so the module-scope auto-start lines stay untouched per the ticket's must-not-change list; its body is now REST. Header comments updated to reflect this.

### DIFF
--- a/server/routes.coalesce.test.js
+++ b/server/routes.coalesce.test.js
@@ -1,0 +1,73 @@
+// #806: regression test for the cold-cache refresh race (re1 review on #818).
+// On first load GitHubPanel fires the four /api/github list endpoints in
+// parallel. The earlier refreshRepoRest skipped (returned immediately) when a
+// refresh was already in flight, so the parallel callers raced ahead of the
+// populated cache and 502'd. _coalesce makes concurrent callers await the SAME
+// in-flight pass. Plain node:assert script.
+
+const assert = require("node:assert/strict");
+const { _coalesce } = require("./routes");
+
+function deferred() {
+  let resolve;
+  const promise = new Promise((r) => { resolve = r; });
+  return { promise, resolve };
+}
+
+(async () => {
+  let passed = 0, failed = 0;
+  const ok = (c, m) => { if (c) { passed++; console.log(`  PASS: ${m}`); } else { failed++; console.error(`  FAIL: ${m}`); } };
+
+  // 1) N concurrent callers for the same key → fn runs ONCE, all share result.
+  {
+    const map = new Map();
+    let calls = 0;
+    const gate = deferred();
+    const fn = () => { calls++; return gate.promise.then(() => ({ value: 42 })); };
+
+    const callers = [1, 2, 3, 4].map(() => _coalesce(map, "repo", fn));
+    ok(calls === 1, "fn invoked exactly once for 4 concurrent callers");
+    ok(map.has("repo"), "in-flight promise is registered while pending");
+
+    gate.resolve();
+    const results = await Promise.all(callers);
+    ok(results.every((r) => r.value === 42), "all 4 callers resolve to the same result");
+    ok(!map.has("repo"), "map entry cleared after the pass settles");
+  }
+
+  // 2) A fresh call AFTER settle starts a new pass (no permanently-cached promise).
+  {
+    const map = new Map();
+    let calls = 0;
+    const fn = () => { calls++; return Promise.resolve("x"); };
+    await _coalesce(map, "k", fn);
+    await _coalesce(map, "k", fn);
+    ok(calls === 2, "sequential (post-settle) calls each start a fresh pass");
+  }
+
+  // 3) Different keys do not coalesce into each other.
+  {
+    const map = new Map();
+    let a = 0, b = 0;
+    const g = deferred();
+    _coalesce(map, "a", () => { a++; return g.promise; });
+    _coalesce(map, "b", () => { b++; return g.promise; });
+    ok(a === 1 && b === 1, "distinct keys each get their own pass");
+    g.resolve();
+    await Promise.resolve();
+  }
+
+  // 4) A throwing/rejecting pass still clears the in-flight entry (no stuck key).
+  {
+    const map = new Map();
+    let threw = false;
+    try {
+      await _coalesce(map, "k", async () => { throw new Error("boom"); });
+    } catch { threw = true; }
+    ok(threw, "rejection propagates to the caller");
+    ok(!map.has("k"), "map entry cleared even when the pass rejects");
+  }
+
+  console.log(`\n${passed} passed, ${failed} failed\n`);
+  process.exit(failed > 0 ? 1 : 0);
+})();
--- a/server/routes.githubCanonicalShape.test.js
+++ b/server/routes.githubCanonicalShape.test.js
@@ -1,0 +1,170 @@
+// #806 (#805 step 1): canonical dashboard-shape tests. Plain node:assert
+// script — no test runner is wired up. Run with
+// `node server/routes.githubCanonicalShape.test.js`.
+//
+// Asserts the REST+ETag fetcher's transforms produce the SHAPE that
+// GitHubPanel.tsx and the batch-progress server logic already expect. The
+// frontend render predicates below are copied VERBATIM from
+// src/components/GitHubPanel.tsx (line refs noted) so this test fails loudly if
+// the canonical shape and the renderer ever drift apart.
+
+const assert = require("node:assert/strict");
+const {
+  restIssueToCanonical,
+  restClosedIssueToCanonical,
+  restMergedPrToCanonical,
+  restPullBaseToCanonical,
+  mapReviews,
+  deriveReviewDecision,
+  buildStatusCheckRollup,
+} = require("./routes");
+
+// ── GitHubPanel.tsx render predicates (verbatim) ──
+// GitHubPanel.tsx:106
+function issueStatusColor(state) {
+  return state === "OPEN" ? "bg-accent" : "bg-error";
+}
+// GitHubPanel.tsx:109-113
+function reviewColor(decision) {
+  if (decision === "APPROVED") return "bg-accent";
+  if (decision === "CHANGES_REQUESTED") return "bg-error";
+  return "bg-[#ffcc00]";
+}
+// GitHubPanel.tsx:127-141
+function ciColor(rollup) {
+  if (!rollup || rollup.length === 0) return "text-text-muted";
+  const states = rollup.map((c) => c.state);
+  if (states.every((s) => s === "SUCCESS")) return "text-accent";
+  if (states.some((s) => s === "FAILURE" || s === "ERROR")) return "text-error";
+  return "text-[#ffcc00]";
+}
+function ciLabel(rollup) {
+  if (!rollup || rollup.length === 0) return "—";
+  const states = rollup.map((c) => c.state);
+  if (states.every((s) => s === "SUCCESS")) return "pass";
+  if (states.some((s) => s === "FAILURE" || s === "ERROR")) return "fail";
+  return "run";
+}
+
+// ── REST fixtures (trimmed to the fields the transforms read) ──
+const REST_OPEN_ISSUE = {
+  number: 42, title: "Fix the thing", state: "open",
+  html_url: "https://github.com/o/r/issues/42",
+  labels: [{ name: "bug" }, { name: "agent/dev" }],
+  assignees: [{ login: "alice" }], created_at: "2026-05-01T00:00:00Z",
+  // NOTE: no pull_request key → a real issue.
+};
+const REST_PR_AS_ISSUE = { number: 7, title: "a PR", state: "open", pull_request: { url: "..." }, html_url: "x" };
+const REST_CLOSED_ISSUE = { number: 9, title: "done", state: "closed", html_url: "u9", closed_at: "2026-05-02T00:00:00Z" };
+const REST_MERGED_PR = { number: 30, title: "merged pr", state: "closed", html_url: "u30", merged_at: "2026-05-03T00:00:00Z", user: { login: "bob" } };
+const REST_OPEN_PR = {
+  number: 100, title: "open pr", state: "open", html_url: "u100",
+  user: { login: "carol" }, assignees: [{ login: "carol" }],
+  created_at: "2026-05-04T00:00:00Z", head: { sha: "deadbeef" },
+};
+
+function run() {
+  let passed = 0, failed = 0;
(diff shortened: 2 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The REST + ETag direction matches #806, and the canonical shape work is generally aligned. However, the new refresh dedup introduces a cold-cache race in the per-project GitHub endpoints that can make the dashboard load partial/empty data after restart.

### Findings
- [high] Concurrent cold-cache list requests can return 502 instead of waiting for the shared snapshot
  - File: `server/routes.js:1222`
  - File: `server/routes.js:1654`
  - Problem: `refreshRepoRest()` dedupes with `_restRefreshing.has(repo)` but returns `{ status: "unchanged", data: null }` immediately for every concurrent caller. On a cold cache, `GitHubPanel` fires `/api/github/issues`, `/api/github/prs`, `/api/github/closed-issues`, and `/api/github/merged-prs` in parallel. The first request starts the snapshot fetch; the other three hit the in-flight guard, return without waiting, find no `fresh` cache entry for their slice, and respond `502`. The frontend catches those failures and leaves those sections empty until a later poll.
  - Suggestion: Store the in-flight refresh promise per repo and have concurrent callers await it, then read the populated slice cache. Alternatively, have `serveGithubList()` wait/retry when a repo refresh is already in progress instead of treating the deduped return as a completed refresh.

### Decision
Request changes. The core implementation is close, but the cold-start path is part of the existing per-endpoint contract and needs to work reliably because the current frontend calls those endpoints concurrently.

realproject7:
**@re2 review — CHANGES REQUESTED** (one real regression; rest verified solid)

Heavy verification done — this is a strong rewrite, but I found one user-visible regression worth fixing before merge.

### 🔴 Should-fix: concurrent cold-load 502 race in `serveGithubList`
`GitHubPanel.tsx` fires all **four** list endpoints (`issues`/`prs`/`closed-issues`/`merged-prs`) concurrently for the **same repo** (lines 185/193/204/209). On a **cold cache** (server restart, or first view of a project):
- The first request acquires the `_restRefreshing` lock and fetches the full snapshot.
- The other three call `refreshRepoRest(repo)` → `_restRefreshing.has(repo)` is true → it **returns `{data:null}` immediately without awaiting the in-flight fetch** (routes.js:1222) → the cold path then reads an empty `_ghEndpointCache` → **HTTP 502**.

So on every cold load, 3 of the 4 panels 502. The frontend swallows it (`.catch(()=>{})`) and the panels stay **empty until the next poll — up to 60s** (poll interval is 60s). The old `cachedGhEndpoint` didn't have this because each endpoint fetched independently. Self-healing but a real per-restart UX regression.

**Suggested fix** — make concurrent callers share the in-flight promise instead of bailing:
```js
const _restRefreshing = new Map(); // repo -> in-flight Promise
async function refreshRepoRest(repo) {
  const inflight = _restRefreshing.get(repo);
  if (inflight) return inflight;            // share the SAME fetch
  const p = (async () => {
    try { /* existing body */ } finally { _restRefreshing.delete(repo); }
  })();
  _restRefreshing.set(repo, p);
  return p;
}
```
Then all four cold callers await the same snapshot and read a populated cache.

### 🟡 Minor (non-blocking)
- **`_etagStore` unbounded growth**: keyed by `#checkruns-${sha}`/`#status-${sha}`, so a new entry accrues per commit pushed to any open PR over the server's lifetime, each holding the full check-runs/status JSON. Slow leak for a long-running dashboard. Consider pruning sha-keyed entries not seen in the last pass.
- **Merged-PRs window**: `pulls?state=closed&per_page=30` then client-side `merged_at` filter. A repo with >30 recently-updated *closed-unmerged* PRs could yield <5 merged. Fine for this repo's merge-heavy flow; narrower than the old server-side `--state merged`.

### ✅ Verified solid
- **Core 304 design holds (tested live):** `gh api -i` exits **1** on 304 with `gh: HTTP 304` in stderr; `_is304Error` matches; the catch path reuses cached payload and doesn't overwrite the ETag store. Confirmed against the real repo.
- **Parsing on real bytes:** `_parseGhInclude` + `_extractEtag` extract a clean `W/"…"` ETag (no trailing CR) and the body parses as JSON with all fields the transforms read.
- **32-assertion test passes** against the real extracted transforms; predicates are **verbatim** copies of GitHubPanel.tsx (issueStatusColor:106, reviewColor:109-113, ciColor:127-133, ciLabel:135-141 — confirmed) so shape/renderer drift fails loudly.
- **No `gh pr list`/`gh issue list` left** (only comments + the out-of-scope `gh … view` paths via the retained `ghJsonExecAsync`).
- Canonical UPPERCASE shape pinned in both REST and the GraphQL emergency fallback; `refreshGraphQLCache` correctly re-gated on the **REST** budget now its body is REST; `/api/github/all` fallback is safe (awaits `_mapLimited`, then reads the populated `_graphqlCache`).

Fix the cold-load race and I'll approve — everything else checks out.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The blocking cold-cache race from the prior review is fixed. `refreshRepoRest()` now coalesces concurrent refreshes through a per-repo in-flight promise, so parallel `/api/github/*` cold-load requests await the same completed snapshot before reading their slice caches.

### Findings
- No blocking findings.

### Decision
Approve. The added `_coalesce` primitive resolves the first-load 502 regression and is covered by focused tests for shared concurrent results, post-settle fresh calls, distinct keys, and cleanup on rejection. Note: GitHub reports no checks for this branch, so this review is based on static inspection and the PR stated local test/build results.

realproject7:
**@re2 re-review — APPROVED ✅**

The cold-cache race (my blocking finding) is resolved by d832f06.

Verified:
- `_coalesce(map, key, fn)` invokes `fn()` eagerly and registers the promise **synchronously** before returning, so same-tick concurrent callers join the *same* in-flight pass; `.finally(() => map.delete(key))` clears the key on both settle **and** reject (no stuck key, fresh pass after settle). The `(async () => fn())()` wrapper also converts a sync throw into a rejection so cleanup always runs.
- `refreshRepoRest` now delegates to it — the single coalesced pass populates all four slice caches before resolving, so the four parallel `/api/github/*` cold-load callers all await it and read populated cache.
- Ran the new `routes.coalesce.test.js` — **8/8 pass** against the real `_coalesce`.
- Built a standalone simulation of the actual `serveGithubList` cold path with 4 concurrent same-repo callers: **all return 200, single network pass** (previously 3×502). `node --check` clean on the updated file.

Both earlier minor non-blockers (`_etagStore` sha-keyed growth; merged-PRs `per_page=30` client-filter window) remain as-is — fine, non-blocking. Good to merge from my side.

PR #829, opened 2026-05-29
[#824] GitHub list endpoints: keep bare-array body, signal stale via header

### LINKED ISSUE #824: [#806 follow-up] GitHub list endpoints corrupt the bare-array contract on stale/rate-limited responses
## Problem
`serveGithubList` (powers `/api/github/{issues,prs,closed-issues,merged-prs}`) caches a **bare array** as `cached.data`, but the stale + rate-limited paths return `{ ...cached.data, _stale: true }`:

- `server/routes.js:1906` (`isRateLimited() && cached`)
- `server/routes.js:1911` (stale-while-revalidate)

Spreading an array yields `{"0":…,"1":…,"_stale":true}` — **not an array**. `GitHubPanel` guards every list with `Array.isArray(data)` and **silently drops** the payload, so the GitHub board goes **empty during every REST-rate-limit or stale-revalidate window** (exactly when it should serve last-known data). The idle path just above (`routes.js:1889`) already does this correctly via a header.

(Latent pattern inherited from the old `cachedGhEndpoint`; surfaced now that serveGithubList is the steady-state path.)

## Fix
For the four list endpoints, keep the JSON a **bare array** and signal stale via a **response header** (mirror the `X-QuadWork-Idle` pattern), e.g. `res.set("X-QuadWork-Stale","1")` / `"X-QuadWork-Rate-Limited","1"` then `return res.json(cached.data)`. Do NOT spread the array.

## Acceptance criteria
- [ ] `/api/github/{issues,prs,closed-issues,merged-prs}` always return a JSON array (never an object) on fresh, stale, and rate-limited paths
- [ ] Stale/rate-limited state signaled via header, not by mutating the body
- [ ] `GitHubPanel` still renders last-known data during a simulated rate-limit/stale window (add/extend a test)
- [ ] Clients no longer rely on `_stale`/`_rateLimited` **body** fields for these four list endpoints (the body is a bare array); any consumer reading those moves to the new header(s); update GitHubPanel accordingly
- [ ] `npm run build` passes

realproject7:
Closes #824.

## Problem
`serveGithubList` (powers `/api/github/{issues,prs,closed-issues,merged-prs}`) caches a **bare array** as `cached.data`, but three return paths spread it into an object:
- in-TTL-but-flagged-stale (`routes.js:1902`)
- critically REST-rate-limited (`routes.js:1906`)
- stale-while-revalidate (`routes.js:1911`)

`{ ...someArray, _stale: true }` yields `{"0":…,"_stale":true}` — **not an array**. `GitHubPanel` guards every list with `Array.isArray(data)` and silently drops the payload, so the board goes **empty during every stale-revalidate or REST-rate-limit window** — exactly when it should serve last-known data. The stale-while-revalidate branch is the high-frequency one (cache expires ~every 30s → next request blanks the board until a later poll lands fresh data).

## Fix
Mirror the existing `X-QuadWork-Idle` pattern: all four list endpoints **always return a bare JSON array** and signal state via response **headers** — `res.set("X-QuadWork-Stale","1")` / `res.set("X-QuadWork-Rate-Limited","1")` — never by spreading the array. Fixed all three spread sites (the ticket named two; line 1902's in-TTL-stale branch had the same latent bug and is fixed too).

`GitHubPanel` needs **no change**: it only reads `Array.isArray(data)` and never consumed the `_stale`/`_rateLimited` **body** fields. `/api/github-parsed` and `/api/batch-progress` legitimately return objects and are out of scope (unchanged).

## Tests
New `server/routes.githubListBareArray.test.js` (18 assertions, `node` script — matches the repo's existing test style):
- Documents the bug (spread → non-array → GitHubPanel drops it).
- Drives the **real handler** on fresh / in-TTL-stale / rate-limited / rate-limited+stale paths: asserts body is always an array, correct headers set, and GitHubPanel's verbatim `Array.isArray` guard accepts it. Config is fed via an `fs.readFileSync` stub so the real `~/.quadwork/config.json` is untouched.
- Source guard over the whole `serveGithubList` body (covers the stale-while-revalidate branch, which fires a background refresh): asserts it never spreads `cached.data` and uses both headers.

## Acceptance criteria
- [x] All four endpoints return a JSON array on fresh, stale, and rate-limited paths
- [x] Stale/rate-limited signalled via header, not by mutating the body
- [x] `GitHubPanel` renders last-known data during a simulated rate-limit/stale window (test asserts the verbatim guard accepts the payload)
- [x] No consumer relies on `_stale`/`_rateLimited` body fields for these four endpoints (GitHubPanel never read them)
- [x] `npm run build` passes

## Verification
- `node server/routes.githubListBareArray.test.js` → 18/18
- Existing `routes.githubCanonicalShape` / `routes.batchProgressSnapshot` / `routes.coalesce` tests → OK
- `node --check server/routes.js` clean; `npm run build` green

⚠️ Server-only change; no UI pixels affected. No GitHub checks are configured on this repo.

### DIFF
--- a/server/routes.githubListBareArray.test.js
+++ b/server/routes.githubListBareArray.test.js
@@ -1,0 +1,130 @@
+// #824 (#806 follow-up): the four GitHub list endpoints
+// (/api/github/{issues,prs,closed-issues,merged-prs}) cache a BARE ARRAY as
+// cached.data. The stale + rate-limited paths used to return
+// `{ ...cached.data, _stale: true }` — spreading an array yields
+// {"0":…,"1":…,"_stale":true}, an OBJECT, which trips GitHubPanel's
+// `Array.isArray(data)` guard and silently blanks the board exactly during the
+// rate-limit / stale-revalidate window when it should serve last-known data.
+//
+// This regression locks the contract: the body stays a JSON array on every
+// path; stale/rate-limited state is signalled via response HEADERS
+// (X-QuadWork-Stale / X-QuadWork-Rate-Limited), mirroring X-QuadWork-Idle.
+//
+// Plain node:assert script — no test runner is wired up. Run with
+// `node server/routes.githubListBareArray.test.js`.
+
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
+
+const CONFIG_PATH = path.join(os.homedir(), ".quadwork", "config.json");
+
+// Feed getRepo()/isProjectIdle() a fixed config WITHOUT touching the real
+// ~/.quadwork/config.json. routes.js holds the shared `fs` module object
+// (const fs = require("fs")), so patching readFileSync here patches its calls.
+const FAKE_CONFIG = JSON.stringify({
+  projects: [{ id: "proj", repo: "owner/repo", idle: false }],
+});
+const realReadFileSync = fs.readFileSync;
+fs.readFileSync = (p, ...rest) =>
+  p === CONFIG_PATH ? FAKE_CONFIG : realReadFileSync(p, ...rest);
+
+const routes = require("./routes");
+const { serveGithubList, _ghEndpointCache, _rateLimit } = routes;
+
+const CACHE_KEY = "issues:owner/repo";
+const REPO_ARRAY = [
+  { number: 1, title: "open issue", state: "OPEN", assignees: [], url: "u1" },
+];
+
+// GitHubPanel.tsx:191 (verbatim) — the guard that drops a non-array payload.
+function panelAccepts(data) {
+  return Array.isArray(data);
+}
+
+function mockRes() {
+  const res = { headers: {}, body: undefined, statusCode: 200 };
+  res.set = (k, v) => { res.headers[k.toLowerCase()] = v; return res; };
+  res.status = (c) => { res.statusCode = c; return res; };
+  res.json = (b) => { res.body = b; return res; };
+  return res;
+}
+
+const REQ = { query: { project: "proj" } };
+
+async function run() {
+  let passed = 0, failed = 0;
+  const ok = (c, m) => { if (c) { passed++; console.log(`  PASS: ${m}`); } else { failed++; console.error(`  FAIL: ${m}`); } };
+
+  // 0) Document the bug: spreading the bare array produces an OBJECT that the
+  //    GitHubPanel guard rejects — i.e. the board would go empty.
+  {
+    const buggy = { ...REPO_ARRAY, _stale: true };
+    ok(!Array.isArray(buggy), "spreading a bare array yields a NON-array (the #824 bug)");
+    ok(!panelAccepts(buggy), "GitHubPanel guard would DROP the spread object → blank board");
+    ok(panelAccepts(REPO_ARRAY), "GitHubPanel guard accepts a bare array");
+  }
+
+  // 1) Fresh cache hit → bare array, no staleness header.
+  {
+    _rateLimit.remaining = 5000;
+    _ghEndpointCache.set(CACHE_KEY, { ts: Date.now(), data: REPO_ARRAY, stale: false });
+    const res = mockRes();
+    await serveGithubList(REQ, res, "issues");
+    ok(Array.isArray(res.body), "fresh: body is a JSON array");
+    ok(res.body === REPO_ARRAY, "fresh: serves the cached array verbatim");
+    ok(!res.headers["x-quadwork-stale"], "fresh: no X-QuadWork-Stale header");
+    ok(panelAccepts(res.body), "fresh: GitHubPanel renders last-known data");
+  }
+
+  // 2) In-TTL but flagged stale → bare array + X-QuadWork-Stale header.
+  {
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The bare-array response shape is fixed for the stale branches, but the critical REST-rate-limited path still does not reliably signal stale/rate-limited state via headers because `adaptiveTTL()` makes the normal cache-hit branch catch it first.

### Findings
- [blocking] Critical rate-limit responses can bypass the new `X-QuadWork-Stale` / `X-QuadWork-Rate-Limited` headers.
  - File: `server/routes.js:1900`
  - Detail: `adaptiveTTL(GH_ENDPOINT_CACHE_TTL)` returns `Infinity` when `isRateLimited()` is true, so `cached && Date.now() - cached.ts < ttl` is always true for any cached entry. That returns at `server/routes.js:1909` before the explicit rate-limited branch at `server/routes.js:1912`, and if `cached.stale` is false the response has no stale or rate-limited headers even when the cache is expired and REST is critically rate-limited. This misses #824's acceptance criteria to signal stale/rate-limited via headers.
  - Suggestion: Check `isRateLimited() && cached` before the adaptive-TTL cache-hit branch, or use the non-adaptive base TTL for deciding whether the cached entry is actually stale before setting headers. Add a regression where `_rateLimit.remaining = 0`, cached `ts` is older than `GH_ENDPOINT_CACHE_TTL`, and `stale: false`; assert the body is an array and both headers are present.

### Decision
Requesting changes until the rate-limited cached response path is reachable/tested and preserves the new header contract.

realproject7 (commented):
**Approved from @re2.** ✅

Verified the bare-array contract fix on the real handler.

- Confirmed the bug: `GitHubPanel` (src/components/GitHubPanel.tsx:190–211) only does `if (Array.isArray(data)) setX(data)` and never reads `_stale`/`_rateLimited` from the body — so the old `{ ...cached.data, _stale: true }` spread produced `{"0":…,"_stale":true}`, failed the guard, and blanked the board on exactly the stale / rate-limited / stale-while-revalidate paths where last-known data should show.
- Fix is complete: all object-spread sites in `serveGithubList` removed (including the latent in-TTL-stale branch the ticket didn't name). State now travels via `X-QuadWork-Stale` / `X-QuadWork-Rate-Limited`, mirroring the existing `X-QuadWork-Idle` pattern. All success paths return bare arrays (idle→`cached.data ?? []`, in-TTL, rate-limited, SWR, cold→`fresh.data`). Error paths (400/502) still return objects, but they're error statuses and the `Array.isArray` guard safely ignores them — pre-existing, no regression.
- `GitHubPanel` needs no change (no consumer ever read the removed body fields — verified by grep across `src/`).
- New `routes.githubListBareArray.test.js`: 18/18 pass against the real handler (fed config via fs stub, real `~/.quadwork/config.json` untouched); includes a source guard against the spread regression. `node --check` clean on routes.js + test. Re-ran existing `routes.batchProgressSnapshot` (28/28) and `routes.coalesce` (8/8) — no regressions.

Server-only; could not pixel-verify but this path has no new UI. Good to merge on code grounds.

project7-interns (approved):
RE1

## Verdict: APPROVE

### Summary
The blocking rate-limit header bug is fixed. Critical REST-rate-limited cached responses now run through the dedicated branch before adaptive cache-hit logic, keep the body as a bare array, and set rate-limit/stale headers according to base-cache freshness.

### Findings
- None blocking.

### Decision
Approved. The new regression covers critical rate limit + expired cached array + `stale:false` with both `X-QuadWork-Rate-Limited` and `X-QuadWork-Stale`, plus the within-base-TTL case with rate-limited-only. GitHub reports no checks for this branch, so I relied on live diff/static inspection plus the PR's stated local build/test result.

realproject7 (commented):
**Re-approved from @re2** on the current revision (commit 2517ba0). ✅

Re-reviewed @re1's blocking finding and the fix:
- Root cause confirmed: under critical REST rate-limit `adaptiveTTL()` returns `Infinity`, so the old `age < adaptiveTTL` cache-hit branch ran *before* the rate-limited branch and returned an expired array with `stale:false` and **no** headers.
- Fix is correct: the `isRateLimited() && cached` branch now precedes the adaptive cache-hit branch, and freshness for the staleness signal is measured against the **base** `GH_ENDPOINT_CACHE_TTL` (`baseFresh`), not `adaptiveTTL`. So rate-limited+cached always sets `X-QuadWork-Rate-Limited`, adds `X-QuadWork-Stale` when `!baseFresh || cached.stale`, and a genuinely-fresh-under-rate-limit cache is served rate-limited-only (honest, not over-flagged). Body stays a bare array on every path.
- Regression coverage added as requested: critical rate-limit + expired cached array + `stale:false` → bare array + BOTH headers; plus within-base-TTL → rate-limited header only.

Re-ran the suite on 2517ba0: `routes.githubListBareArray.test.js` 25/25, `node --check` clean on routes.js. Good to merge on code grounds.

PR #831, opened 2026-05-29
[#825] Reset self-heal breaker on manual agent stop/restart

### LINKED ISSUE #825: [#797 follow-up] Self-heal circuit-breaker state persists across manual agent stop/restart
## Problem
The self-heal circuit-breaker state (`{ lastAutoRestartAt, countInWindow, ... }` per agent) is **never cleared when an agent is manually stopped**. `selfHeal.clearState(key)` is not called from `stopAgentSession()` in `server/index.js`.

Consequence: if an agent trips the circuit breaker (3 auto-restarts / 15 min → auto-recovery paused), and the operator then **manually stops and restarts** it, the stale breaker state suppresses the next legitimate auto-restart — the fresh manual session inherits the old "paused" window.

## Fix
Call `selfHeal.clearState(key)` inside `stopAgentSession(key)` so a manual stop resets the per-agent self-heal window. (Trivial, ~1 line.)

## Acceptance criteria
- [ ] Manually stopping an agent clears its self-heal cooldown/circuit-breaker state
- [ ] After a manual stop+restart, auto-recovery works again even if the breaker had tripped before
- [ ] Existing self-heal tests still pass; add a case asserting clearState-on-stop
- [ ] `npm run build` passes

realproject7:
Closes #825. Follow-up to #797 (PR #815).

## Problem
The self-heal circuit-breaker state (`{ lastAutoRestartAt, countInWindow, ... }` per agent) was never cleared on a manual stop — `selfHeal.clearState(key)` wasn't called from `stopAgentSession`. So if an agent tripped the breaker (3 auto-restarts / 15 min → auto-recovery paused) and the operator then manually **stopped and restarted** it, the stale breaker state suppressed the fresh session's next legitimate auto-restart.

## Fix
`stopAgentSession(key, { clearSelfHeal = false })` now calls `selfHeal.clearState(key)` when the flag is set. It is **gated, not unconditional** — `stopAgentSession` is also the internal stop of `restartAgentSession`, which the self-heal **auto-restart** path uses. An unconditional clear there would reset `countInWindow` on every auto-restart and defeat the #797 breaker. So:

| Path | clearSelfHeal |
|------|---------------|
| manual `/stop` route | ✅ true |
| manual `/restart` route | ✅ true |
| project reset (`/reset`, #417) | ✅ true |
| full-reset (#657) | ✅ true |
| **auto-restart** (`reason: "thinking-block-400"`) | ❌ false (breaker preserved) |

`restartAgentSession` forwards the flag to its internal `stopAgentSession`. Cleared before the session lookup, so a manual stop resets the window even when no live session remains (agent already exited).

## Acceptance criteria
- [x] Manually stopping an agent clears its self-heal cooldown/circuit-breaker state
- [x] After a manual stop+restart, auto-recovery works again even if the breaker had tripped before
- [x] Existing self-heal tests still pass; added a case asserting clearState-on-stop
- [x] `npm run build` passes

## Tests
- `self-heal.test.js` **Test 7** (new): trip the breaker (3 restarts), `clearState`, then assert the next thinking-block 400 **auto-restarts again** (window reset, breaker no longer suppressing) and the breaker message isn't re-emitted — the module behaviour that makes a manual stop+restart re-enable auto-recovery. 20/20.
- `index.selfHealStop.test.js` (new): `server/index.js` binds a port at require time, so it isn't unit-requirable — this is a focused **source-guard** (same approach as the existing `rate-limit-handling` test) asserting the wiring: clear is gated on `clearSelfHeal`, the **auto-restart path does NOT pass it** (breaker preserved), and the four manual paths do. 8/8.
- `node --check` clean; `npm run build` green.

⚠️ Server-only change; no UI pixels. No GitHub checks are configured on this repo. Key design point for review: the clear is intentionally **gated** so it never fires on the self-heal auto-restart (which routes through `stopAgentSession`) — otherwise the #797 breaker would never trip.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -514,6 +514,8 @@ async function spawnAgentPty(project, agent, opts = {}) {
           onRestart: () => {
             session._autoRecovering = true;
             console.log(`[self-heal] ${key}: thinking-block 400 detected — restarting session`);
+            // #825: NO clearSelfHeal here — the breaker window must persist
+            // across auto-restarts so repeated trips can pause auto-recovery.
             restartAgentSession(key, { reason: "thinking-block-400" })
               .then((result) => {
                 if (result && result.ok) {
@@ -557,7 +559,16 @@ async function spawnAgentPty(project, agent, opts = {}) {
   }
 }
 
-async function stopAgentSession(key) {
+async function stopAgentSession(key, { clearSelfHeal = false } = {}) {
+  // #825 (#797 follow-up): a MANUAL stop/restart/reset must reset the per-agent
+  // self-heal circuit-breaker window, so a fresh operator-driven session isn't
+  // suppressed by a stale "paused" state from a prior trip. This is gated:
+  // the auto-restart path (restartAgentSession with reason "thinking-block-400")
+  // leaves clearSelfHeal=false, because clearing on every auto-restart would
+  // reset countInWindow each time and defeat the #797 breaker entirely. Cleared
+  // before the session lookup so a manual stop resets the window even when no
+  // live session remains (e.g. the agent already exited).
+  if (clearSelfHeal) selfHeal.clearState(key);
   const session = agentSessions.get(key);
   if (!session) {
     agentSessions.set(key, { projectId: null, agentId: null, term: null, viewers: new Set(), viewerDims: new Map(), lastDims: null, state: "stopped", error: null });
@@ -628,7 +639,7 @@ app.post("/api/agents/:project/reset", async (req, res) => {
     for (const agentId of allAgentIds) {
       const s = agentSessions.get(`${projectId}/${agentId}`);
       if (s) s._suppressLifecycleMsg = true;
-      await stopAgentSession(`${projectId}/${agentId}`);
+      await stopAgentSession(`${projectId}/${agentId}`, { clearSelfHeal: true }); // #825: manual reset resets the self-heal window
     }
 
     // Respawn all agents with fresh MCP tokens
@@ -667,7 +678,7 @@ app.post("/api/full-reset", async (_req, res) => {
     console.log("[full-reset] stopping all agent sessions...");
     const sessionKeys = [...agentSessions.keys()];
     for (const key of sessionKeys) {
-      await stopAgentSession(key);
+      await stopAgentSession(key, { clearSelfHeal: true }); // #825: manual full-reset resets the self-heal window
     }
 
     console.log("[full-reset] stopping Butler...");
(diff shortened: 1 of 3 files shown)


project7-interns (approved):
RE1

## Verdict: APPROVE

### Summary
The fix satisfies #825: manual stop/restart/reset paths now clear the per-agent self-heal breaker state, while self-heal auto-restart preserves the breaker window so #797's circuit breaker still works.

### Findings
- None blocking.

### Decision
Approved. `stopAgentSession` gates `selfHeal.clearState(key)` behind `clearSelfHeal`, `restartAgentSession` forwards the flag, manual `/stop`, `/restart`, project reset, and full-reset pass it, and the `thinking-block-400` auto-restart path does not. GitHub reports no checks for this branch, so I relied on live diff/static inspection plus the stated local test/build result.

realproject7 (commented):
**Approved from @re2.** ✅

Reviewed the breaker-reset wiring and root cause.

- **Root cause confirmed:** `selfHeal._state` (per-agent `restarts[]` window + `lastAutoRestartAt` cooldown + `breakerNotified`) was never cleared on a manual stop. After the breaker trips (`restarts.length >= CIRCUIT_MAX` → `breaker`, `breakerNotified=true`), a manual stop+restart within the 15-min `CIRCUIT_WINDOW_MS` inherited the stale restarts so `observeChunk` re-tripped immediately — and worse, `breakerNotified` stayed true so the breaker message wasn't even re-emitted. Silent suppression of the fresh session's legitimate auto-restart. `clearState(key)` does `_state.delete(key)` → a complete reset, so it's the right fix.

- **Gating is correct** — the load-bearing design point. `stopAgentSession(key, { clearSelfHeal })` clears only when set, and the clear runs **before** the session lookup / `if (!session)` early-return, so a manual stop resets the window even when the agent already exited. Audited all 6 call sites:
  - auto-restart `restartAgentSession(key, { reason: "thinking-block-400" })` (index.js:519) → no `clearSelfHeal` → false → **breaker preserved** ✓
  - `/stop` (753), `/restart` → `restartAgentSession({ reason:"manual", clearSelfHeal:true })` (781), project reset (642), full-reset (681) → all `clearSelfHeal: true` ✓
  - `restartAgentSession` forwards the flag to `stopAgentSession` (772) ✓

  So every operator-driven path resets and the only auto path doesn't — exactly preserving #797's circuit breaker.

- No other session-teardown path needs clearing (process self-exit isn't a manual stop and correctly leaves the window intact).

Tests on 0b979c6: `self-heal.test.js` 20/20 — incl. new Test 7 (trip breaker → `clearState` → next 400 re-restarts, `restarts===4`, breaker msg NOT re-emitted), and `index.selfHealStop.test.js` 8/8 (source-guard: clear gated/off-by-default, auto path doesn't clear, all four manual paths do, restart forwards). `node --check` clean. Good to merge on code grounds.

PR #823, opened 2026-05-29
[#814] Move Idle control to sidebar as an Active/Inactive switch (+ Settings status)

realproject7:
Closes #814. Follow-up to #812 (per-project Idle). UI-only — server gating unchanged.

## What
Reframes per-project Idle as an **Active/Inactive** switch (operators shouldn't reason about "idle") and moves it from the GitHub-panel header into the sidebar, plus the Settings page.

## How
- **`src/lib/idle.ts`** (new, shared source of truth):
  - `persistProjectIdle(id, idle)` — re-reads config, flips only that project's `idle`, PUTs (the server `syncTriggers` hook stops a now-idle project's trigger, #812); throws so callers can revert.
  - **Lightweight signal** `quadwork:idle-changed` (`emitIdleChange`/`onIdleChange`) — the ticket's sanctioned "re-read on a lightweight signal" so every open view re-syncs without a reload.
  - Confirmation copy.
- **`ActiveSwitch`** (new) — compact, sharp-cornered, accent `#00ff88` ON / muted OFF; shows `!project.idle`. Deliberately **not** the rounded iOS-blue style. Stops click propagation so it works inside a clickable nav row.
- **`ConfirmModal`** (new) — small dark/sharp modal, reused by sidebar + settings (Esc/overlay cancels).
- **Sidebar** — each expanded project row gets the switch on the right (sibling of the nav Link). Turning a project **OFF → Inactive** shows the confirmation modal explaining what idle pauses; turning **ON → resume** needs no confirmation. Optimistic + revert-on-failure; listens for the signal.
- **ProjectDashboard** — removed the `Idle ○/●` header button; it now loads idle on mount and **reacts to the signal so its pollers start/stop live** when toggled in the sidebar/settings (no manual reload). The BatchProgressPanel paused/idle empty state is kept.
- **SettingsPage** — per-project **Active/Inactive status + the same switch + confirmation**, persisted immediately; keeps the batched-Save snapshot's `idle` in lockstep (`syncSavedIdle`) so an idle toggle isn't flagged dirty or clobbered by a later Save.

## Acceptance criteria
- [x] Each sidebar project item has an Active/Inactive switch on the right; ON = active (default), OFF = idle
- [x] Turning OFF shows a confirmation popup explaining idle; turning ON does not
- [x] Switch persists `project.idle` and reverts on save failure
- [x] Old GitHub-panel-header idle button removed
[...]

Issue #814, opened 2026-05-29
Move Idle control to sidebar as an Active/Inactive switch (+ Settings status)


realproject7:
## Type: enhancement (UX) — follow-up to #812

Builds on the per-project Idle feature shipped in #812 (config flag `project.idle`, server gating, PUT /api/config persistence + `syncTriggers` hook, revert-on-failure). This ticket only changes **where/how** the operator toggles it and reframes the mental model.

## Goal
Move the idle control out of the GitHub panel header into the **sidebar**, as a small **Active/Inactive switch on each project nav item**, and surface the same status on the **Settings page**.

## 1. Sidebar switch (`src/components/Sidebar.tsx`)
- Add a compact toggle switch on the **right side of each project nav item** (project list in the sidebar).
- **Invert the mental model to Active/Inactive** (operators shouldn't have to reason about "idle"):
  - Switch **ON** (default, accent `#00ff88`) = **Active** = `idle: false`
  - Switch **OFF** (muted) = **Inactive (parked)** = `idle: true`
  - i.e. the switch displays `!project.idle`. Default ON since `idle` defaults false.
- **Confirmation when turning a project OFF (→ idle):** show a popup/modal before applying, e.g.:
  > **Set "{project}" to Inactive?**
  > This pauses everything for this project: GitHub polling, the GitHub board & batch progress, scheduled triggers, and bridge auto-lifecycle. Agents stay running but won't be pulsed; the dashboard shows last-known data until you set it Active again.
  > [Cancel] [Set Inactive]
  - Turning a project back **ON (resume)** needs **no** confirmation — resuming is safe.
- Persist via the existing `PUT /api/config` path (writes `idle`), reuse #812's **optimistic + revert-on-failure** behavior (don't desync UI from server). The `syncTriggers` hook already stops a running trigger when a project goes idle.

## 2. Remove the old control
- Remove the `Idle ○ / ●` button from the GitHub panel header in `src/components/ProjectDashboard.tsx` (it moves to the sidebar). Keep the BatchProgressPanel "paused/idle" empty state.

## 3. Settings page (`src/components/SettingsPage.tsx`)
- In each project's section, show its **Active / Idle status** and a toggle (same Active/Inactive switch + confirmation-on-going-idle). This and the sidebar must read/write the same source of truth (`project.idle` in config) and stay consistent.

## 4. State-sync consideration (the non-trivial bit)
[...]

Issue #812, opened 2026-05-29
Per-project Idle toggle — pause all non-essential GitHub/API activity for parked projects


realproject7:
## Goal
Add a per-project **Idle** toggle on the project page. While `project.idle === true`, QuadWork initiates **zero project-specific GitHub/API activity** for that project — across **every** server poller, client poller, and request endpoint — so parked projects don't consume the shared GitHub budget or pulse their agents. Complements epic #805 (manual operator lever).

## Config + entry point
- New per-project flag `idle: boolean` (default false).
- **Setting idle must stop side effects atomically.** Don't rely on the frontend to also call stop. Either: (a) make the config-write path idle-aware (the existing `app.set("syncTriggers")` hook fires on `PUT /api/config` — have `syncTriggersFromConfig` clear idle projects' timers, see below), or (b) add a dedicated endpoint that sets `idle=true` AND stops the trigger/bridge state in one call. Toggling idle OFF resumes polling; the scheduled trigger is left for the operator to restart (no silent auto-resume).

## Server-side gating (ALL must check `project.idle`)
- **Board GitHub fetch** (`refreshGraphQLCache` / the REST fetcher from #806): skip idle projects.
- **`autoStopPollingTick`** (`server/index.js:1632`, already loops projects): `if (project.idle) continue;` — covers the 30s batch-progress GraphQL hit AND bridge auto-start/stop.
- **`/api/projects`** (`server/routes.js:801`): for idle projects do **not** run the per-project `gh pr list` calls (routes.js:~56-57); return cached/zero/stale metadata flagged `_idle`.
- **`syncTriggersFromConfig`** (`server/index.js:1590`): skip idle projects when starting timers from `trigger_enabled`, AND clear any existing trigger timer for a project that is now idle (this is what makes a `PUT /api/config` idle-write actually stop a running trigger).
- **Trigger endpoints** (`server/index.js`): `POST /api/triggers/:project/start` (1195) and `/send-now` (1266) **no-op while idle** (do not create timers or pulse agents; return a clear `{ idle: true }` response). `/stop` (1260) remains allowed.
- **Bridge auto-polling**: server `autoStopPollingTick` skip (above) + don't auto-start/stop bridges for idle projects.
- **Dashboard endpoints** (`/api/github/*`, `/api/batch-progress`) for an idle project: serve last-cached/stale data with an `_idle` flag — do not initiate a fresh fetch.

## Client-side gating (ALL five direct batch-progress pollers must stop when idle)
[...]

PR #832, opened 2026-05-29
[#826] Idle: live signal beats stale initial read; disable switch while saving

### LINKED ISSUE #826: [#814 follow-up] Idle initial-load can race and overwrite a live toggle in ProjectDashboard
## Problem
`ProjectDashboard` reads `idle` from `/api/config` on mount (async) AND subscribes to the shared idle signal (`onIdleChange`). These can race:

1. Dashboard mounts → starts `GET /api/config` (idle=false).
2. Operator toggles the project Inactive in the sidebar → `onIdleChange` fires → dashboard sets idle=true (correct).
3. The earlier in-flight config read resolves with **stale** data → overwrites idle back to **false** (wrong).

Result: the open dashboard's pollers desync from the actual (server-persisted) idle state — it keeps polling a parked project (or stops polling an active one) until reload. `src/components/ProjectDashboard.tsx:143-157`.

## Fix
Make the live signal authoritative over the initial read: e.g. ignore the initial config result if an idle signal has already been received (a ref/seq guard), or abort/skip the initial read once subscribed. 

Minor (same PR): `ActiveSwitch` has no pending/disabled state — add `disabled` while a toggle PUT is in flight to avoid queued double-requests (low, mitigated by the confirm modal).

## Acceptance criteria
- [ ] Toggling idle elsewhere while the dashboard is mounting does not get overwritten by the initial config read
- [ ] The dashboard's pollers reflect the final toggled state without a manual reload
- [ ] `ActiveSwitch` ignores/blocks repeat clicks while a save is pending
- [ ] `npm run build` passes

realproject7:
Closes #826. Follow-up to #814 (PR #823).

## Problem (race)
`ProjectDashboard` reads `idle` from `/api/config` on mount (async) AND subscribes to the shared idle signal (`onIdleChange`). They race:
1. Dashboard mounts → `GET /api/config` starts (idle=false).
2. Operator toggles the project Inactive in the sidebar → `onIdleChange` fires → dashboard sets idle=true (correct).
3. The earlier in-flight config read resolves with **stale** data → overwrites idle back to **false** (wrong).

The open dashboard's pollers then desync from the persisted idle state until a manual reload (`ProjectDashboard.tsx:143-157`).

## Fix
The live signal is now **authoritative** over the initial read:
- `idleSignalRef` records that a signal arrived for this project. The config-read resolution still marks `idleLoadedRef` done, but **skips `setIdle`** when a signal already landed — so a stale read can't clobber the operator's toggle.
- Added a `cancelled` cleanup guard so a read resolving **after a project switch** can't set idle for the wrong project (adjacent staleness, free to fix here).

## Minor (same PR, per the ticket)
`ActiveSwitch` already accepts `disabled`, but the callers never set it. `Sidebar` and `SettingsPage` now track an `idlePending` set (added on toggle, cleared in the persist `.finally`) and pass `disabled={pending}` to the switch, so a queued double-toggle can't race the in-flight PUT. Threaded through `ProjectIcon` to all three sidebar render sites; covers both the **park** (confirm modal) and **resume** paths (both route through `applyIdle`).

## Acceptance criteria
- [x] Toggling idle elsewhere while the dashboard is mounting is not overwritten by the initial config read
- [x] The dashboard's pollers reflect the final toggled state without a manual reload (signal authoritative; `cancelled` guard for project switches)
- [x] `ActiveSwitch` blocks repeat clicks while a save is pending (disabled during the PUT, in sidebar + settings, park + resume)
- [x] `npm run build` passes

## Verification
`npm run build` green; no new lint errors (the pre-existing `set-state-in-effect` warnings on the mount/reset effects are unchanged — the `setIdle(false)` reset predates this change, and the new signal/read `setIdle` calls live in async callbacks the rule permits). UI-only; #812 server gating unchanged.

⚠️ This repo has **no frontend test harness** (jest isn't wired; the test suite is server-side `node` scripts), so — as with #823 — the change is verified by build + static review rather than a unit test. A browser eyeball of the disabled switch during a toggle would be a nice-to-have; I can't capture screenshots here (Chromium won't install on this host). No GitHub checks are configured on this repo.

### DIFF
--- a/src/components/ProjectDashboard.tsx
+++ b/src/components/ProjectDashboard.tsx
@@ -136,24 +136,39 @@ export default function ProjectDashboard({ projectId }: ProjectDashboardProps) {
   // the operator flips the switch elsewhere (no manual reload).
   const [idle, setIdle] = useState(false);
   const idleLoadedRef = useRef(false);
+  // #826: a live idle signal that arrives while the initial /api/config read is
+  // in flight is AUTHORITATIVE — the operator just toggled the switch, so the
+  // (possibly stale) config result must not overwrite it back. Tracked per
+  // mount/project and reset alongside idleLoadedRef.
+  const idleSignalRef = useRef(false);
   useEffect(() => {
     idleLoadedRef.current = false;
+    idleSignalRef.current = false;
     setIdle(false);
   }, [projectId]);
   useEffect(() => {
     if (idleLoadedRef.current) return;
+    // #826: guard against this read resolving after the project changed (a
+    // stale read from the prior project would otherwise set idle here).
+    let cancelled = false;
     fetch("/api/config")
       .then((r) => (r.ok ? r.json() : null))
       .then((cfg) => {
-        if (!cfg) return;
+        if (cancelled || !cfg) return;
+        idleLoadedRef.current = true;
+        // A live signal already set the authoritative value — don't clobber it.
+        if (idleSignalRef.current) return;
         const entry = (cfg.projects || []).find((p: { id: string }) => p.id === projectId);
         setIdle(!!entry?.idle);
-        idleLoadedRef.current = true;
       })
       .catch(() => {});
+    return () => { cancelled = true; };
   }, [projectId]);
   useEffect(() => onIdleChange(({ projectId: pid, idle: next }) => {
-    if (pid === projectId) setIdle(next);
+    if (pid === projectId) {
+      idleSignalRef.current = true;
+      setIdle(next);
+    }
   }), [projectId]);
 
   // Poll agent states
--- a/src/components/SettingsPage.tsx
+++ b/src/components/SettingsPage.tsx
@@ -283,6 +283,9 @@ export default function SettingsPage() {
   const [butlerStartConfig, setButlerStartConfig] = useState<{ command: string; model: string } | null>(null);
   // #814: pending "Set Inactive?" confirmation (null = no modal).
   const [idleConfirm, setIdleConfirm] = useState<{ id: string; name: string } | null>(null);
+  // #826: project ids with an idle PUT in flight — disable the switch for them
+  // so a queued double-toggle can't race the persist.
+  const [idlePending, setIdlePending] = useState<Set<string>>(new Set());
 
   // Keep the saved snapshot's idle in lockstep with an idle toggle. Idle is
   // persisted immediately (out of band from the batched Save), so without this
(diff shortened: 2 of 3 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR addresses #826: a live idle signal now wins over the initial `/api/config` read in `ProjectDashboard`, and idle switches are disabled while their PUT is pending in both Sidebar and Settings. The changes are narrowly scoped to the affected UI components.

### Findings
- No blocking findings.

### Decision
Approve. I verified the diff and surrounding code statically; GitHub reports no checks for this branch, and I did not run a browser harness because this repo does not appear to have one configured for the frontend.

realproject7 (commented):
**Approved from @re2.** ✅

UI race fix — reviewed the logic and verified types.

**Race fix (ProjectDashboard)** — correct. The new `idleSignalRef` makes a live signal authoritative: `onIdleChange` sets `idleSignalRef.current = true` before `setIdle(next)`, and the `/api/config` resolution now sets `idleLoadedRef = true` then `return`s early when `idleSignalRef.current` is set, so a stale read can't clobber a toggle that landed mid-flight. The `cancelled` guard (set in the effect cleanup) additionally stops a read resolving after a project switch from setting idle for the wrong project. Effect ordering holds: the reset effect (`idleLoadedRef=false; idleSignalRef=false; setIdle(false)`) is declared before the config-read effect, so on `projectId` change refs reset before the read's top guard runs and a fresh fetch happens per project. Once `idleSignalRef` is true it stays true for the mount, but that's harmless post-load (config read is one-shot, gated by `idleLoadedRef`).

**Pending state (Sidebar + SettingsPage)** — correct. `idlePending` is added in `applyIdle` and cleared in `.finally` (both immutable `new Set(prev)` updates), and `ActiveSwitch disabled` is wired from it — threaded via `togglePending` to all 3 sidebar render sites + the settings row. Confirmed both directions route through `applyIdle`: resume via `handleToggleActive`, park via the confirm modal's `onConfirm → applyIdle(id, true)` — so the switch is disabled during either in-flight PUT, blocking queued double-toggles. Revert-on-failure preserved.

Verification: `tsc --noEmit` clean (exit 0). ESLint surfaces only 4 `react-hooks/set-state-in-effect` errors at Sidebar.tsx:279/287/290 (mobile-overlay reset + restore-on-mount) — all **pre-existing on main and outside this PR's diff** (confirmed the lines exist verbatim on `origin/main`); none are introduced here. I could not run the full Turbopack build in my sandbox (it rejects a cross-filesystem `node_modules`, an env artifact not the code), but @dev reports `npm run build` green and the type-check covers these changes. Good to merge on code grounds.

Issue #837, opened 2026-05-30
Recently-merged PRs empty: gh api closed-PR fetch exceeds default 1MB maxBuffer (also disables #828/#834 fast-path)


realproject7:
## Problem
"Recently Merged PRs" is **always empty** (and `/api/github/merged-prs` returns `[]`) on any active repo, even when there are many recent merged PRs.

## Root cause — `maxBuffer` overflow on the closed-PR fetch
`ghApiConditional` (`server/routes.js:1027`) runs:
```js
await _execFileAsync("gh", args, { encoding: "utf-8", timeout: 15000 }); // no maxBuffer
```
Node's `execFile` default `maxBuffer` is **1 MB**. The closed-PR list fetch `gh api -i 'repos/{o}/{r}/pulls?state=closed&per_page=100&sort=updated&direction=desc'` returns **~1.74 MB** (PR objects are large; with `-i` headers on top). The call **rejects** with `ERR_CHILD_PROCESS_STDIO_MAXBUFFER`, which is caught, is not a 304, and returns `{ status: "error", data: null }` → `selectRecentMergedPrs` gets nothing → **0 merged PRs**.

Measured on `realproject7/quadwork`:
- `pulls?state=closed&per_page=100` `-i` → **1,739,372 bytes** (exceeds 1 MB) → fails
- `issues?state=closed&per_page=100` `-i` → 577,354 bytes (under 1 MB) → works (which is why "Recently Closed Issues" renders but "Recently Merged PRs" doesn't)

## Secondary impact — silently disables the #828/#834 budget fast-path
Because the closed-PR fetch returns `status:"error"`, `closedPagesComplete` is `false` for that repo **forever** (its last page is an error), so:
- `closedPrsWindowComplete` is always `false` → the "classify queued from snapshot" fast-path (#828/#834) **never triggers** → every queued issue falls back to REST/Search each poll.
Correctness still holds (no false-queued, no GraphQL), but the snapshot optimization is silently off on any repo with ~60+ closed PRs, and the merged display is broken.

## Fix
Set a generous `maxBuffer` on the `gh api` execs that fetch lists — primarily `ghApiConditional` (`routes.js:1027`):
```js
await _execFileAsync("gh", args, { encoding: "utf-8", timeout: 15000, maxBuffer: 32 * 1024 * 1024 });
```
[...]

PR #94, opened 2026-04-02
[#88] Per-project AgentChattr instances

realproject7:
## Summary
- Moved `agentchattr_url`, `agentchattr_token`, and MCP ports from global config to per-project config
- Added `resolveProjectChattr(projectId)` in `server/config.js` with per-project → global fallback for backward compatibility
- Chat proxy now accepts `project` query param to route to the correct AgentChattr instance
- AgentChattr lifecycle endpoints are per-project (Map of processes instead of single global)
- Setup wizard auto-assigns unique ports per project (8300+idx for AgentChattr, 8200+idx*2 for MCP)
- ChatPanel accepts `projectId` prop and routes all API calls to the project's AgentChattr
- SettingsPage shows per-project AgentChattr URL, token, and MCP port fields
- Existing single-project configs continue to work unchanged (global fields serve as fallbacks)

Fixes #88

## Test plan
- [ ] Existing single-project config still works (backward compat)
- [ ] New project via setup wizard gets auto-assigned ports
- [ ] Chat panel on `/project/<id>` shows only that project's messages
- [ ] Per-project AgentChattr start/stop/restart works via API
- [ ] Settings page shows per-project AgentChattr fields
- [ ] Trigger messages use per-project AgentChattr config
- [ ] `/api/projects` dashboard fetches chat per-project

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #844, opened 2026-05-30
[#839] /api/batch-active: completion-aware via batch-progress cache

### LINKED ISSUE #839: Sidebar heartbeat pulses on completed batches: /api/batch-active ignores completion
## Symptom
A project's sidebar item shows the green "active batch" heartbeat pulse (`animate-pulse-ring`) even though its batch is **complete** (no active work). Reproduced: QuadPlan pulses while its Current Batch panel reads COMPLETE.

## Root cause
The sidebar pulse is driven by `GET /api/batch-active?project=X` (`src/components/Sidebar.tsx:339` → `activeBatches` → `hasActiveBatch ? "animate-pulse-ring"` at `Sidebar.tsx:166`).

`/api/batch-active` (`server/routes.js:2439`) is **not completion-aware** — it returns `active: true` whenever `parseActiveBatch(queueText).issueNumbers.length > 0`, i.e. whenever the `## Active Batch` section of `OVERNIGHT-QUEUE.md` lists *any* issue, regardless of whether those issues are done/merged. So once a batch finishes, if Head leaves the item in `## Active Batch` (e.g. a trailing "blocked"/parked ticket), the pulse never stops.

Confirmed live: quadplan `OVERNIGHT-QUEUE.md` still lists `#94` under `## Active Batch` (status "blocked - operator-run E2E required"), so `/api/batch-active` → `{active:true}`, while `/api/batch-progress` correctly reports `complete: true, completeConfirmed: true`. The two endpoints disagree.

## Fix
Make `/api/batch-active` completion-aware so the pulse means "batch actively in progress," not just "Active Batch section non-empty." Reuse the batch-progress completion logic — return `active` only when there is a displayed batch with items AND it is not complete (mirror `/api/batch-progress`'s `complete`/`completeConfirmed`, or compute `items.length > 0 && !complete`). Prefer sharing the existing resolve/progress path over re-parsing.

## Acceptance criteria
- [ ] `/api/batch-active` returns `active:false` for a batch whose items are all merged/done (matches `/api/batch-progress` `complete`)
- [ ] The sidebar heartbeat pulse stops when a batch completes, even if `## Active Batch` still lists parked/blocked items
- [ ] Still pulses for a genuinely in-progress batch
- [ ] `npm run build` passes

realproject7:
## Summary
- Sidebar heartbeat (`animate-pulse-ring`) was firing whenever `## Active Batch` listed *any* issue, so a parked/blocked ticket left in the section made the pulse run forever — even after every item merged/closed and the Current Batch panel showed COMPLETE (live repro on quadplan: Active Batch still listed `#94`, panel was COMPLETE, sidebar still pulsed).
- Adds `isBatchActiveFromProgress(progress)` helper in `server/routes.js`: returns `items.length > 0 && !complete` from a cached `/api/batch-progress` payload (mirrors the panel's COMPLETE flip at `routes.js:2599`). Returns `null` when there's no progress yet, signalling the route to fall back.
- `/api/batch-active` now reads `_batchProgressCache` first via that helper. On cache miss the prior file-only check stays in place so the route remains cheap for first paint — Sidebar polls all projects every 30s, so cost matters.

## Why this is safe
- No new `gh` calls fire from the sidebar poll (cache lookup is O(1)).
- Cache is populated by the panel's own `/api/batch-progress` calls, so the sidebar and panel see the same `complete` truth as soon as either has been hit once.
- Cache-miss fallback preserves today's behavior exactly for projects whose progress has never been computed — no regression for first paint.

## Test plan
- [x] `node server/routes.batchActiveCompletion.test.js` → 10/10 PASS (new unit test covering null/empty/in-progress/complete + a defensive `complete:true / completeConfirmed:false` case).
- [x] `node server/routes.batchProgressSnapshot.test.js` → 45/45 PASS (no regression on adjacent batch-progress code).
- [x] `node server/routes.ghApiConditionalMaxBuffer.test.js` → 6/6 PASS (#837 regression test still green).

Fixes #839

### DIFF
--- a/server/routes.batchActiveCompletion.test.js
+++ b/server/routes.batchActiveCompletion.test.js
@@ -1,46 +1,207 @@
 // #839 regression: /api/batch-active must drop to active:false once a batch
 // is complete, even if Head leaves a parked/blocked ticket in the
-// `## Active Batch` section. The route now derives `active` from the cached
-// /api/batch-progress payload via isBatchActiveFromProgress(progress), so the
-// sidebar heartbeat agrees with the Current Batch panel. Plain node:assert
-// script — run with `node server/routes.batchActiveCompletion.test.js`.
+// `## Active Batch` section. The route now derives `active` from the shared
+// getOrComputeBatchProgress() path so both /api/batch-progress and
+// /api/batch-active see the same `complete` truth — including on cache miss
+// (re1's REQUEST CHANGES on #844: the prior PR's cache-miss path still
+// returned the old file-only answer, so a completed batch with a parked
+// item could still pulse the sidebar before the panel had been opened once).
+//
+// Layers covered:
+//   - isBatchActiveFromProgress(progress): pure truth-table (10 cases).
+//   - getOrComputeBatchProgress(projectId) end-to-end with stubbed fs +
+//     pre-populated _graphqlCache so progressFromSnapshot resolves every
+//     item without touching gh (the cache-miss completed-batch path that
+//     was previously uncovered).
+//
+// Plain node:assert script — run with
+// `node server/routes.batchActiveCompletion.test.js`.
 
-const { isBatchActiveFromProgress } = require("./routes");
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
 
-function run() {
+const CONFIG_PATH = path.join(os.homedir(), ".quadwork", "config.json");
+
+// ── fs stubs installed BEFORE require("./routes") ──────────────────────────
+// routes.js reads CONFIG_PATH via getRepo/isProjectIdle on every call, and
+// reads OVERNIGHT-QUEUE.md + batch-progress-cache.json on the cache-miss
+// compute path. We replace those reads with fixture values per test and
+// route everything else through the real fs so module require still works.
+const real = {
+  readFileSync: fs.readFileSync,
+  writeFileSync: fs.writeFileSync,
+  mkdirSync: fs.mkdirSync,
+  statSync: fs.statSync,
+  unlinkSync: fs.unlinkSync,
+};
+let cfgJson = JSON.stringify({ projects: [] });
+let queueText = "";
+fs.readFileSync = function stubReadFileSync(p, ...rest) {
+  if (p === CONFIG_PATH) return cfgJson;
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The helper itself matches the intended `items.length > 0 && !complete` rule when a batch-progress payload is present, but the route still falls back to the old file-only behavior on cache miss. That means `/api/batch-active` can still return `active:true` for the completed parked-item case before `/api/batch-progress` has populated `_batchProgressCache` for that project.

### Findings
- [high] `/api/batch-active` remains not completion-aware on cache miss
  - File: `server/routes.js:2466`
  - The new route only uses completion state when `_batchProgressCache.get(projectId)` has data. If the sidebar polls a project whose progress has not been computed in this server process, it falls through to the prior `parseActiveBatch(queueText).issueNumbers.length > 0` behavior. For the exact #839 scenario, a completed batch with a parked/blocked issue still listed under `## Active Batch` can therefore still return `{ active: true }`, which violates the acceptance criterion that `/api/batch-active` returns `active:false` for completed batches and that the pulse means an actively in-progress batch.
  - Suggestion: make the fallback completion-aware too, by reusing/mirroring the batch-progress resolution path for the displayed batch when the cache is absent, or otherwise ensure completed parked items cannot return true. Add route-level coverage for the cache-miss completed-batch case, not only the helper.

### Decision
Request changes. The PR fixes the cached path, but the endpoint contract in #839 is still incorrect whenever `/api/batch-active` is hit before `/api/batch-progress` warms the cache. GitHub reports no checks for this branch; the PR author reports `npm run build` passed.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The update addresses my requested change: `/api/batch-active` no longer falls back to the old file-only `parseActiveBatch(...).issueNumbers.length > 0` check on cache miss. It now shares `getOrComputeBatchProgress()` with `/api/batch-progress`, then derives `active` from `items.length > 0 && !complete`, so completed parked items return `active:false` even before the progress cache is warm.

### Findings
- None.

### Decision
Approve. The endpoint now matches #839's acceptance criteria for completed batches, genuine in-progress batches, and cache-miss behavior. The added test coverage exercises the exact cache-miss completed-batch regression I flagged. GitHub reports no checks for this branch; the PR author reports `npm run build` passed.

PR #850, opened 2026-05-30
[#836] Unref routes.js module-load pollers + add cross-platform npm test

### LINKED ISSUE #836: Unref routes.js module-load pollers + add cross-platform 'npm test' (fix full-suite hang)
## Type: infra / test — focused cleanup

(The scoped-down core of the long-deferred #798/#799. Unrelated to feature work; flagged again in review when the full test sweep hangs.)

## Problem
Running the full server test sweep (`node --test server/*.test.js`) **hangs after `server/routes.batchProgress.test.js` prints its pass message** (also `server/routes.parseActiveBatch.test.js`). The assertions pass — the process just never exits.

**Cause:** those two files are plain `node:assert` scripts that `require("./routes")`, and `server/routes.js` starts background pollers **at module load** — `startRateLimitPolling()` (`routes.js:3434`) and `startGraphQLPolling()` (`routes.js:3436`) create timers that are **not `.unref()`'d**:
- `setInterval(refreshRateLimit, …)` (`routes.js:86`)
- `setTimeout(() => refreshGraphQLCache(), 2000)` (`routes.js:1828`)
- `setInterval(refreshGraphQLCache, …)` (`routes.js:1829`)

So any process that imports the module (a test, a one-off `node -e`) has its event loop pinned forever. (This bit us in real life earlier — a stray `node -e` poller ran for 11 days.) There is also still **no `npm test` script**, so CI can't run the suite.

## Fix
1. **`.unref()` the three module-load timers** in `server/routes.js` (lines 86, 1828, 1829). This is **production-invisible**: the Express HTTP listener keeps the server process alive, so the pollers still fire on the same cadence; `.unref()` only stops them from *alone* pinning the loop — which is exactly what lets a test/import process exit.
2. **Add a cross-platform `npm test`**: a small `server/run-tests.js` that discovers `*.test.js` via `fs.readdirSync` recursion (no shell glob — must work on Windows) and runs **each file in its own child process**, failing the run if any child exits non-zero or times out. Wire as `"test": "node server/run-tests.js"`. (Per-file child processes also isolate any residual module state between files.)

## Out of scope
- A GitHub Actions CI workflow (can be a follow-up once `npm test` is green).
- Rewriting the two Jest-global files (`bridge-auto-stop-guard.test.js`, `rate-limit-handling.test.js`) — handle separately if still using `describe`/`expect`.

## Acceptance criteria
- [ ] `node --test server/**/*.test.js` (and the new `npm test`) **terminates** — no hang on the route-test files; all suites pass
- [ ] The three `routes.js` module-load timers are `.unref()`'d; production polling cadence/behavior is unchanged (only `.unref()` added; auto-start at module load preserved)
- [ ] `npm test` runs the full suite cross-platform (no `find`/shell-glob), each file in its own process, correct aggregate exit code
- [ ] `npm run build` passes

realproject7:
## Summary
Two coupled changes for the test-suite-hang root cause flagged in #836:

1. **`.unref()` the three module-load timers in `server/routes.js`**:
   - `startRateLimitPolling()`'s `setInterval(refreshRateLimit, …)` (line ~96)
   - `startGraphQLPolling()`'s warmup `setTimeout(refreshGraphQLCache, 2000)` (line ~1838)
   - `startGraphQLPolling()`'s steady-state `setInterval(refreshGraphQLCache, …)` (line ~1839)

   Production is **unaffected** — the Express HTTP listener already keeps the process alive, so polling cadence and auto-start at module load are unchanged. `.unref()` only stops these timers from *alone* pinning the event loop, which is what was hanging `node --test` and any one-off `node -e` invocation that touched routes.js.

2. **Add `server/run-tests.js` + `"test": "node server/run-tests.js"`**:
   - Discovers every `*.test.js` under `server/` via `fs.readdirSync` recursion (no shell globs → Windows-clean).
   - Runs each file in its own `node <file>` child process. Per-file isolation also prevents residual module state from re-introducing this bug.
   - 60s per-file timeout safety net.
   - Aggregate exit code: 1 if any file fails or times out, 0 otherwise.
   - The two `__tests__/*.test.js` Jest-style files (`bridge-auto-stop-guard`, `rate-limit-handling`) are documented as skipped — rewriting them is out-of-scope per the issue body.

## Test plan
- [x] `npm test` → 24 passed, 0 failed, 2 skipped (no hang).
- [x] `node --test server/routes.batchProgress.test.js server/routes.parseActiveBatch.test.js` — the two files that used to hang now exit in ~550ms.
- [x] Runner uses `path.join` / `process.execPath` / `child_process.spawn` (no shell, no globs) — cross-platform clean.

Fixes #836

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "start": "node server/",
     "lint": "eslint",
     "server": "node server/",
+    "test": "node server/run-tests.js",
     "prepack": "npm run build",
     "release:patch": "npm version patch && git push origin main --follow-tags && VERSION=$(node -p 'require(\"./package.json\").version') && gh release create \"v$VERSION\" --generate-notes --latest && npm publish",
     "release:minor": "npm version minor && git push origin main --follow-tags && VERSION=$(node -p 'require(\"./package.json\").version') && gh release create \"v$VERSION\" --generate-notes --latest && npm publish",
--- a/server/routes.js
+++ b/server/routes.js
@@ -93,7 +93,13 @@ async function refreshRateLimit() {
 function startRateLimitPolling() {
   if (_rateLimitTimer) return;
   refreshRateLimit();
+  // #836: .unref() so this interval doesn't alone pin the event loop.
+  // Production keeps the process alive via the Express HTTP listener, so
+  // cadence/auto-start is unchanged; .unref() only lets test imports and
+  // one-off `node -e` invocations exit cleanly (an unref'd `node -e` poller
+  // ran for 11 days in the wild before this was caught).
   _rateLimitTimer = setInterval(refreshRateLimit, RATE_LIMIT_POLL_MS);
+  _rateLimitTimer.unref();
 }
 
 function isRateLimited() {
@@ -1834,9 +1840,13 @@ async function refreshGraphQLCache() {
 let _graphqlPollTimer = null;
 function startGraphQLPolling() {
   if (_graphqlPollTimer) return;
+  // #836: both the warm-up setTimeout and the steady-state setInterval are
+  // .unref()'d so a test/import process can exit. The Express HTTP listener
+  // keeps prod alive on its own; cadence is unchanged.
   // Initial fetch after a short delay (let rate-limit poll run first).
-  setTimeout(() => refreshGraphQLCache(), 2000);
+  setTimeout(() => refreshGraphQLCache(), 2000).unref();
   _graphqlPollTimer = setInterval(refreshGraphQLCache, GRAPHQL_CACHE_TTL);
+  _graphqlPollTimer.unref();
 }
 
 // ─── #810: batch progress sourced from the REST snapshot ──────────────────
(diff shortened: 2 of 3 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR addresses the hang root cause by unref'ing the three module-load timers in `server/routes.js` while preserving the existing polling cadence and auto-start behavior. It also adds a cross-platform `npm test` runner that discovers server test files recursively, runs runnable files in isolated child processes, enforces per-file timeouts, and returns an aggregate exit code.

### Findings
- None.

### Decision
Approve. The timer changes are scoped to event-loop liveness for import/test processes, and the runner avoids shell globs while isolating each test file. The two Jest-style files are explicitly skipped and logged, which is acceptable here because rewriting those files was called out as out of scope for #836. GitHub reports no checks for this branch; the PR author reports `npm test` and `npm run build` passed.

PR #853, opened 2026-05-30
[#845] Re-seed worktree AGENTS.md from current templates

### LINKED ISSUE #845: Upgrades don't re-seed existing worktree AGENTS.md — new agent seeds (e.g. #809 GITHUB.md discovery) never reach existing projects
## Problem
Upgrading QuadWork ships new agent seed templates (`templates/seeds/*.AGENTS.md`), but **existing projects' worktree `AGENTS.md` files are never updated** — so seed changes (e.g. the #809 "discover via GITHUB.md instead of `gh pr list`" change) **never reach existing projects' agents**. They keep running the old instructions.

Verified live: after the #805 epic merged, `~/Projects/quadwork-head/AGENTS.md` and `quadwork-reviewer2/AGENTS.md` still contain `gh pr list` (0 mentions of GITHUB.md), so the agents keep doing GraphQL-heavy discovery and the epic's budget benefit isn't realized on existing projects.

## Why none of the existing paths fix it
- **"Reset Agents"** (`/api/agents/:project/reset`, `index.js:609`) and **"Full Reset"** (`index.js:671`) only **stop + respawn** the PTYs — agents re-read the *same stale* `AGENTS.md`.
- **Startup "reseed"** (`index.js:1809`) only regex-patches **slug names** (`reviewer1→re1`, …) — it does NOT re-copy seed *content*.
- The **only** full re-seed is on **project (re)creation** (`routes.js:2825`: "AGENTS.md — always (re)write from template …").

## Fix (options)
Add a way to re-seed an existing project's worktree `AGENTS.md` from the current templates, with the existing placeholder substitution (`{{project_name}}`, `{{reviewer_github_user}}`, `{{reviewer_token_path}}`). Ideas:
1. A **"Re-seed agents from templates"** action/endpoint (button in Agent Models or the control bar) that re-copies `templates/seeds/<agent>.AGENTS.md` → each worktree, **only when no batch is active** (respect the #809 "re-seed at a batch boundary, never mid-batch" rule), and preserves any human-added `## Notes`-style sections.
2. And/or **auto-reseed on version upgrade**: detect a package-version change at startup and re-copy seeds at the next batch boundary.

Must preserve: reviewer token paths / identities, and not clobber the running batch (re-seeded content only takes effect on the next agent restart anyway).

## Acceptance criteria
- [ ] An explicit, safe way to bring an existing project's worktree `AGENTS.md` up to the current templates (with placeholder substitution), without recreating the project
- [ ] Does not run mid-batch (or clearly warns); takes effect on next agent (re)start
- [ ] Preserves reviewer identity/token placeholders and human-edited notes
- [ ] After re-seed, agents discover via GITHUB.md (no `gh pr list`) — verifiable by the worktree AGENTS.md no longer referencing `gh pr list`
- [ ] `npm run build` passes

realproject7:
## Summary
- New `POST /api/projects/:id/reseed-agents` rewrites each worktree `AGENTS.md` from `templates/seeds/*.AGENTS.md` with the existing placeholder substitution (`{{project_name}}`, `{{reviewer_github_user}}`, `{{reviewer_token_path}}`). Refuses to run while the project's batch is active (HTTP 409 + `batchActive: true`); operator can override with `{ force: true }`.
- Pure `reseedAgentsMd(existing, fresh)` merger: fresh template wins for any H2 heading it defines; H2 blocks in the existing file whose heading is NOT in the template (e.g. operator `## Notes`) are appended at the end under a marker comment. Case-insensitive heading match. H3+ subsections stay nested under their parent H2.
- "Re-seed AGENTS.md" button in ControlBar's Server section (en + ko) with the same two-click confirm pattern as Full Reset. Distinct feedback line when the server blocks because a batch is active.
- Re-seeded content only takes effect on next agent restart — feedback message tells the operator to restart agents.

## Acceptance criteria
- [x] Existing project worktree `AGENTS.md` files can be brought up to current templates without project recreation (button + endpoint).
- [x] Does not run mid-batch — server returns `409 { batchActive: true }`; UI surfaces "Batch active — re-seed deferred" instead of error.
- [x] Reviewer identity/token placeholders preserved (`{{reviewer_github_user}}` from global config, token path defaults to `~/.quadwork/reviewer-token`).
- [x] Human-edited notes preserved — any H2 heading not in the template is appended at the end under `<!-- Operator notes preserved from prior AGENTS.md (#845) -->`.
- [x] After re-seed, agents discover via `GITHUB.md` (no instruction to use `gh pr list` for board discovery; the `gh pr list` references that remain are the new `## GitHub State (discovery)` anti-instructions telling agents NOT to use it).
- [x] `npm run build` passes.

## Test plan
- [x] `node server/routes.reseedAgentsMd.test.js` → 8 cases pass (empty existing, all-canonical existing, single Notes preservation, multi-section preservation in order, case-insensitive matching, H3 nesting, trailing whitespace normalization, placeholder pass-through).
- [x] `npm test` → 25 passed, 0 failed, 2 skipped.
- [x] `npm run build` → ✓ compiled.
- [x] End-to-end smoke: created 4 temp worktree dirs with a `## Notes` operator section + stale `## Role` content, ran the merger with real `templates/seeds/*.AGENTS.md`, confirmed all 4 files were reseeded, `## Notes` preserved verbatim, placeholders substituted, dev/AGENTS.md now references `GITHUB.md` for board discovery.

Fixes #845

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -3043,16 +3043,38 @@ router.post("/api/projects/:project/reseed-agents", async (req, res) => {
   if (!workingDir) return res.status(400).json({ ok: false, error: "Project has no working_dir" });
 
   if (!force) {
+    // re1 review on #845: fail CLOSED when batch state can't be computed.
+    // Falling through to a re-seed when the gate is uncertain re-introduces
+    // the exact mid-batch rewrite the gate exists to prevent. Three cases
+    // route to the same "unknown" 503 (operator can retry or pass force:true):
+    //  1. getOrComputeBatchProgress throws (network/IO)
+    //  2. getOrComputeBatchProgress returns null (no repo / no progress data)
+    //  3. isBatchActiveFromProgress returns null (no items payload)
+    let active;
     try {
       const progress = await getOrComputeBatchProgress(projectId);
-      if (isBatchActiveFromProgress(progress)) {
-        return res.status(409).json({
-          ok: false,
-          error: "Batch is active — re-seed deferred to avoid mid-batch instruction drift. Pass force:true to override; new seed content only takes effect on next agent restart.",
-          batchActive: true,
-        });
-      }
-    } catch { /* progress unavailable → fall through and re-seed */ }
+      active = isBatchActiveFromProgress(progress);
+    } catch {
+      return res.status(503).json({
+        ok: false,
+        error: "Could not verify batch state — aborting to avoid a mid-batch re-seed. Retry, or pass force:true to override.",
+        batchUnknown: true,
+      });
+    }
+    if (active === null) {
+      return res.status(503).json({
+        ok: false,
+        error: "Batch state unknown (no progress data) — aborting to avoid a mid-batch re-seed. Retry, or pass force:true to override.",
+        batchUnknown: true,
+      });
+    }
+    if (active) {
+      return res.status(409).json({
+        ok: false,
+        error: "Batch is active — re-seed deferred to avoid mid-batch instruction drift. Pass force:true to override; new seed content only takes effect on next agent restart.",
+        batchActive: true,
+      });
+    }
   }
 
   const dirName = path.basename(workingDir);
--- a/server/routes.js
+++ b/server/routes.js
@@ -2961,6 +2961,146 @@ router.post("/api/setup", (req, res) => {
   }
 });
 
+// ─── Re-seed agents from current templates (#845) ──────────────────────────
+
+// Split an AGENTS.md into the leading "intro" text (above the first H2) plus
+// one block per `## H2` heading. Lines that begin with `## ` (exactly two
+// hashes + whitespace) start a new block; H1 / H3+ stay inside the current
+// block so subsections nested under an H2 are not orphaned.
+function _splitAgentsMdSections(text) {
+  const blocks = [];
+  let intro = [];
+  let current = null;
+  for (const line of String(text || "").split("\n")) {
+    const m = line.match(/^##\s+(.+?)\s*$/);
+    if (m) {
+      if (current) blocks.push(current);
+      current = { heading: m[1], body: "" };
+    } else if (current) {
+      current.body += (current.body ? "\n" : "") + line;
+    } else {
+      intro.push(line);
+    }
+  }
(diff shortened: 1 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The re-seed path mostly follows the intended shape, but it fails open when the server cannot determine whether a batch is active. That can rewrite worktree `AGENTS.md` during an active batch without any warning or block, which violates the core safety requirement for this issue.

### Findings
- [high] Re-seed runs when batch activity cannot be checked
  - File: `server/routes.js:3055`
  - The new endpoint catches any `getOrComputeBatchProgress(projectId)` failure and then continues to re-seed. If the progress check fails because GitHub/cache/batch parsing is unavailable while a batch is actually active, the operator gets a successful re-seed instead of a block or warning. Issue #845 explicitly requires the action to not run mid-batch, or at least clearly warn/block at that boundary. This should fail closed unless `{ force: true }` was explicitly provided, for example by returning a 409/503 with a distinct `batchUnknown`/`batchActive` style response and no file writes.
  - Suggestion: Treat inability to compute batch activity as a blocking condition by default. Only proceed when progress was computed as inactive, or when the request explicitly uses `force: true` and the response/UI makes the override clear.

### Decision
REQUEST CHANGES until the batch-boundary guard fails closed instead of silently proceeding when batch activity is unknown.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The requested fail-closed behavior is now implemented: when batch state cannot be verified, the re-seed endpoint returns `HTTP 503` with `batchUnknown: true` and performs no file writes unless `{ force: true }` is explicit. The UI also surfaces that state distinctly from both batch-active and generic failure responses.

### Findings
- No blocking findings.

### Decision
APPROVE. The prior blocker at `server/routes.js:3055` is resolved, and the PR satisfies the issue's safety requirement to block/warn instead of silently re-seeding when the batch boundary cannot be verified. Note: GitHub reports no checks on this branch, so I relied on live diff review plus Dev's reported `npm test` and `npm run build` results.

Issue #855, opened 2026-05-30
Re-seed skips reviewers on legacy-named worktrees — uses reconstructed path instead of config cwd (#845 follow-up)


realproject7:
## Problem
The re-seed endpoint derives each agent's worktree path by **reconstructing** `${dirName}-${agent}` (`server/routes.js:~3091`) instead of using the agent's **configured `cwd`**. When the worktree directory name doesn't match `{project}-{agentId}` — e.g. legacy reviewer worktrees named `quadwork-reviewer1`/`quadwork-reviewer2` while the agent IDs are `re1`/`re2` — the reconstructed path (`quadwork-re1`/`quadwork-re2`) **doesn't exist**, so those agents are silently **skipped** (`skipped.push("re1 (no worktree)")`).

**Verified live on quadwork** after an operator re-seed:
- `quadwork-head`, `quadwork-dev` → re-seeded (GITHUB.md discovery, fresh mtime) ✅
- `quadwork-reviewer1`, `quadwork-reviewer2` → **NOT re-seeded** (still old `gh pr list` seeds) ❌

Config truth: `agents.re1.cwd = ~/Projects/quadwork-reviewer1`, `agents.re2.cwd = ~/Projects/quadwork-reviewer2`. The endpoint ignored these.

**Impact:** the re-seed feature — whose whole purpose is to update *existing* projects to the new GITHUB.md-discovery seeds (#845/#809) — **silently fails for the reviewers on any project with legacy worktree naming**, which is most existing/upgraded projects. The reviewers keep doing old GraphQL-heavy `gh pr list` discovery. This also undercuts the 2.2.0 budget benefit for existing users.

## Fix
Resolve each agent's worktree from the **configured `cwd`** (`cfg.projects[].agents[agentId].cwd`), not `${dirName}-${agent}`. Iterate the project's configured agents and use their actual `cwd` for the AGENTS.md path + the matching `templates/seeds/${agentId}.AGENTS.md`. Keep the `skipped` reporting for genuinely-missing worktrees.

## Acceptance criteria
- [ ] Re-seed updates ALL configured agents' worktrees regardless of directory naming (legacy `reviewer1/reviewer2` included), using each agent's config `cwd`
- [ ] After re-seed on a legacy-named project, re1/re2 `AGENTS.md` reference GITHUB.md discovery as the primary method and no longer carry the old positive instruction to discover by gh pr list (the fresh templates may still mention gh pr list in anti-instructions/fallback)
[...]

PR #858, opened 2026-05-30
[#854] Preserve reviewer token path during AGENTS.md re-seed

### LINKED ISSUE #854: Re-seed clobbers a custom reviewer token path back to default (#845 follow-up)
## Problem (chain verified on main)
The re-seed endpoint can clobber a **custom reviewer token path** back to the default:
- `add-config` (`server/routes.js:2828`) substitutes the operator-provided `reviewerTokenPath` into the seeded `AGENTS.md` but **does not persist it** in project/global config.
- The re-seed endpoint (`server/routes.js:3083`) resolves `reviewerTokenPath = body.reviewerTokenPath || <default ~/.quadwork/reviewer-token>`. Unlike `reviewerUser` (which falls back to `cfg.reviewer_github_user`), the token path has **no config fallback** → it lands on the hardcoded default.
- The ControlBar "Re-seed AGENTS.md" button posts `body: "{}"` (`src/components/ControlBar.tsx:253`), so nothing overrides it.
- The seed template's `## GitHub Authentication` section (`templates/seeds/re1.AGENTS.md:80-83`: `export GH_TOKEN=$(cat {{reviewer_token_path}})`) exists in BOTH fresh and existing AGENTS.md, so `reseedAgentsMd` uses the **fresh** version (not preserved) → the existing custom path is overwritten.

**Impact:** a project originally seeded with a custom reviewer token path → after re-seed, `re1`/`re2` `AGENTS.md` point at the **default** token file → reviewers read the wrong/empty token → `gh pr review` auth fails → reviews break. (Contradicts #845's "preserve reviewer identity/token" intent; `reviewerUser` is preserved, the token path is not.)

## Fix
Make the re-seed preserve the per-agent token path instead of defaulting it. Preferred (works for existing projects without config migration):
- In the re-seed endpoint, before substituting, **extract the existing token path from each worktree's current `AGENTS.md`** — parse the `$(cat <path>)` in its `## GitHub Authentication` section — and use that as `{{reviewer_token_path}}` for that agent.
- Resolution order per agent: `body.reviewerTokenPath` → **extracted-from-existing-AGENTS.md** → `cfg.reviewer_token_path` → default. Apply per-agent (re1/re2 each have their own).
- Optional hardening: also **persist `reviewer_token_path` in config** at `add-config` (`routes.js:2828`) so new projects carry it; existing projects still rely on extraction.

## Acceptance criteria
- [ ] Re-seeding a project that was set up with a custom reviewer token path keeps re1/re2's `GH_TOKEN=$(cat <custom-path>)` unchanged (not reset to the default)
- [ ] Default-path projects are unaffected
- [ ] Token path resolution prefers existing-AGENTS.md (or config) over the hardcoded default
- [ ] Test: existing AGENTS.md with a custom `$(cat /custom/path)` → after re-seed, the custom path is retained while the discovery sections update
- [ ] `npm run build` passes

---

## RE1 review — accepted refinements
- **Robust token-path parsing:** when extracting the existing path, parse common shell forms — `export GH_TOKEN=$(cat /path)`, `export GH_TOKEN="$(cat /path)"`, and quoted paths. If parsing fails, fall back as written (body -> config -> default).
- **Endpoint-level test (not only the pure merger):** seed existing re1/re2 `AGENTS.md` with `/custom/token`, call the re-seed endpoint, and assert the rewritten files still contain `/custom/token`.

realproject7:
## Summary

- Fixes #854. The #845 re-seed always substituted `{{reviewer_token_path}}` with the hardcoded default `~/.quadwork/reviewer-token` (ControlBar posts no `reviewerTokenPath`), so any project set up with a custom `GH_TOKEN=$(cat /custom/...)` line silently lost it and re1/re2 then failed to authenticate.
- Substitution is now **per-agent** and resolves the token path in priority order: `body.reviewerTokenPath` → extracted from this worktree's existing AGENTS.md → `cfg.reviewer_token_path` → default. Head/dev seeds have no GH_TOKEN line so they fall through to the default cleanly.
- New pure helper `_extractReviewerTokenPath(existing)` parses `GH_TOKEN=$(cat …)` with optional `export` prefix, optional outer double-quote wrapping, and optional inner single/double quoting (handles paths with spaces). The existing AGENTS.md is now read BEFORE substitution; that read was already needed by the #845 merger so no new I/O.

## Acceptance criteria

- [x] Re-seeding a project set up with a custom reviewer token path keeps re1/re2 `GH_TOKEN=$(cat <custom-path>)` unchanged (Test 1 endpoint-level).
- [x] Default-path projects are unaffected (Test 2).
- [x] Existing AGENTS.md / config token path is preferred over the hardcoded default when no body override is provided (Tests 1, 3).
- [x] Parse common shell forms robustly: `export GH_TOKEN=$(cat /path)`, `export GH_TOKEN=\"$(cat /path)\"`, quoted/spaced paths, single quotes, tabs, indentation (extractor unit test, 15 cases).
- [x] Endpoint-level test asserts custom token survives + discovery sections update.
- [x] `npm run build` passes.

## Test plan

- [x] `node server/routes.extractReviewerTokenPath.test.js` → 15 parse-form cases pass.
- [x] `node server/routes.reseedTokenPreservation.test.js` → 4 endpoint scenarios pass (custom-preserved, default-roundtrip, cfg-fallback, body-override).
- [x] `node server/routes.reseedAgentsMd.test.js` → 8 cases pass (existing merger contract intact).
- [x] `node server/routes.resolveReseedTargets.test.js` → 7 cases pass (#855 contract intact).
- [x] `npm test` → 28 passed, 0 failed, 2 skipped.

Closes #854

### DIFF
--- a/server/routes.extractReviewerTokenPath.test.js
+++ b/server/routes.extractReviewerTokenPath.test.js
@@ -1,0 +1,122 @@
+// #854: _extractReviewerTokenPath unit tests. Pure parser — covers the
+// common shell forms an operator might have edited a worktree AGENTS.md
+// into so the re-seed substitution preserves their custom token path
+// instead of clobbering it with the default.
+//
+// Plain node:assert script — auto-discovered by the #836 cross-platform
+// runner. Run directly with `node server/routes.extractReviewerTokenPath.test.js`.
+
+const assert = require("node:assert/strict");
+const { _extractReviewerTokenPath } = require("./routes");
+
+// 1) Canonical seed form — `export GH_TOKEN=$(cat <path>)`. The exact line
+//    `templates/seeds/re1.AGENTS.md` ships with, after substitution.
+{
+  const txt = `## GitHub Authentication\nexport GH_TOKEN=$(cat /Users/cho/.local/reviewer-token)\nNext section...`;
+  assert.equal(_extractReviewerTokenPath(txt), "/Users/cho/.local/reviewer-token");
+}
+
+// 2) No `export` prefix — operator may have stripped it (still a valid shell line).
+{
+  const txt = `GH_TOKEN=$(cat /custom/path)\n`;
+  assert.equal(_extractReviewerTokenPath(txt), "/custom/path");
+}
+
+// 3) Outer double-quote wrapping — `"$(cat …)"`. Common operator edit when
+//    they paste a snippet from elsewhere.
+{
+  const txt = `export GH_TOKEN="$(cat /opt/secrets/gh-token)"\n`;
+  assert.equal(_extractReviewerTokenPath(txt), "/opt/secrets/gh-token");
+}
+
+// 4) Inner double-quoted path (path with spaces).
+{
+  const txt = `export GH_TOKEN=$(cat "/Users/op/My Tokens/gh.token")\n`;
+  assert.equal(_extractReviewerTokenPath(txt), "/Users/op/My Tokens/gh.token");
+}
+
+// 5) Inner single-quoted path.
+{
+  const txt = `GH_TOKEN=$(cat '/srv/tokens/gh-prod')\n`;
+  assert.equal(_extractReviewerTokenPath(txt), "/srv/tokens/gh-prod");
+}
+
+// 6) Outer + inner double quotes together.
+{
+  const txt = `export GH_TOKEN="$(cat "/path with spaces/tok")"\n`;
+  assert.equal(_extractReviewerTokenPath(txt), "/path with spaces/tok");
+}
+
+// 7) Indented (inside a fenced code block in the existing AGENTS.md).
+{
+  const txt = `\`\`\`bash\n    export GH_TOKEN=$(cat /tab/indented/token)\n\`\`\`\n`;
+  assert.equal(_extractReviewerTokenPath(txt), "/tab/indented/token");
+}
+
+// 8) Tabs between tokens — `export\tGH_TOKEN  =\t$(cat …)`.
+{
+  const txt = `export\tGH_TOKEN\t=\t$(cat /tabbed/token)\n`;
+  assert.equal(_extractReviewerTokenPath(txt), "/tabbed/token");
+}
+
+// 9) Multiple GH_TOKEN lines — first match wins. (Operators shouldn't have
+//    two, but if they do we don't crash; canonical first-line picks up the
+//    one that's documented/used.)
+{
+  const txt = `export GH_TOKEN=$(cat /first/path)\n# Or alternately:\nGH_TOKEN=$(cat /second/path)\n`;
+  assert.equal(_extractReviewerTokenPath(txt), "/first/path");
+}
+
+// 10) Different `cat` invocation forms inside `$()` — extra whitespace.
+{
+  assert.equal(_extractReviewerTokenPath(`GH_TOKEN=$( cat /spaced )`), "/spaced");
+  assert.equal(_extractReviewerTokenPath(`GH_TOKEN=$(cat  /double-space)`), "/double-space");
+}
+
+// 11) No GH_TOKEN line at all — head/dev seeds. Returns null so the resolver
+//     falls through to cfg / default (the placeholder isn't even present in
+//     those templates anyway).
+{
+  const txt = `# Dev\n\nNo authentication section here.\n`;
+  assert.equal(_extractReviewerTokenPath(txt), null);
+}
+
+// 12) GH_TOKEN set to a literal value (no `$(cat …)`) — we can't preserve a
+//     path that isn't there. Returns null; the substitution falls through
+//     to the default and the operator's literal-value line is preserved as
+//     a custom section by the #845 merger if they keep it under a non-canonical
+//     H2 (here it's inline in the same fresh section, so the fresh template
+//     will replace it — which is acceptable because there's no path to keep).
+{
+  const txt = `export GH_TOKEN=ghp_HARDCODED_LITERAL_TOKEN\n`;
+  assert.equal(_extractReviewerTokenPath(txt), null);
+}
+
+// 13) GH_TOKEN line that references the placeholder verbatim (pre-substitution
+//     content somehow round-tripped). Returns the literal placeholder text —
+//     the resolver caller will pass it back into substitution unchanged, which
+//     is functionally equivalent to using the default for re-seeding purposes.
+//     Documented behavior to keep the extractor a true round-trip.
+{
+  const txt = `export GH_TOKEN=$(cat {{reviewer_token_path}})\n`;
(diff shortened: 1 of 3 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
PR #858 fixes the token-path clobber by resolving `{{reviewer_token_path}}` per worktree before re-seeding, preserving existing reviewer `GH_TOKEN=$(cat <path>)` lines when no explicit override is provided. The implementation is focused and composes cleanly with the #845 merger and #855 cwd-based target resolution.

### Findings
- No blocking findings.

### Decision
APPROVE. The resolver priority matches #854: `body.reviewerTokenPath` -> extracted existing AGENTS.md path -> `cfg.reviewer_token_path` -> default. The parser covers the requested shell forms, including optional `export`, outer `"$(cat ...)"`, and quoted paths with spaces. The endpoint-level test exercises the actual re-seed route with custom re1/re2 token paths and verifies discovery content refreshes while the custom paths remain. GitHub reports no checks on the branch, so this approval is based on live diff review plus the reported `npm test` / `npm run build` results.

realproject7 (commented):
APPROVE — verified on PR branch (head `fb73d0f`).

**Extractor (`_extractReviewerTokenPath`)**
- Two-stage regex: anchored line match `^[ \t]*(?:export[ \t]+)?GH_TOKEN[ \t]*=.*$` then a non-greedy `\$\(\s*cat\s+(.+?)\s*\)` extraction. Stripping outer single/double quotes after trim is the right order — the inner-quoted "path with spaces" case (test 4) only works because we strip AFTER trim, and trim-then-strip handles both `"…"` and `'…'` symmetrically.
- Line anchor `^…$` with `/m` flag correctly distinguishes a real shell setter from prose like `Run \`GH_TOKEN=…\``  (test 15) — substring `GH_TOKEN=` mid-line never satisfies `^[ \t]*…GH_TOKEN`. Defensive against false positives in operator-edited prose.
- Returns null cleanly for: no GH_TOKEN line, GH_TOKEN with literal value (no `$(cat …)`), empty/null/non-string. Caller falls through to cfg/default. ✓
- Pure, no I/O, re-export gated to the test surface only.

**Per-agent resolution (route loop)**
- Order matches the AC exactly: `body.reviewerTokenPath` → extracted from this worktree's existing AGENTS.md → `cfg.reviewer_token_path` → default `~/.quadwork/reviewer-token`.
- Uses `||` (truthy) on `body.reviewerTokenPath`, not `??`. This is correct in context: an empty-string token path is never useful, so falling through to extraction is the right behavior. Consistent with how `defaultReviewerTokenPath` is constructed unconditionally upfront.
- Reads `existing` ONCE per agent and reuses it for both extraction and the `reseedAgentsMd` merger — no double I/O. The read is `try`/`catch`'d to empty string, so a permissions error degrades gracefully to the cfg/default fallback. ✓
- Head/dev seeds have no `{{reviewer_token_path}}` placeholder, so the resolved value is harmlessly substituted into nothing. Endpoint test #1 verifies this explicitly (`headAfter.includes("GH_TOKEN=$(cat")` → false). No accidental injection of a GH_TOKEN line into head/dev. ✓

**Unit tests (`routes.extractReviewerTokenPath.test.js`, 15 cases)**
- Covers canonical seed line, no-`export`, outer/inner quote combinations, indented-in-codeblock, tabs, multi-line first-wins, spaced `cat`, no-line, literal value, placeholder round-trip, empty/null/non-string, prose mention. Comprehensive parse-form coverage.

**Endpoint tests (`routes.reseedTokenPreservation.test.js`, 4 scenarios)**
- Real `express` + `http` in-process server, real router mounted, real fs for templates and temp worktrees. Only stubs `fs.readFileSync(CONFIG_PATH)` so the fixture config gets loaded — hermetic on any developer machine.
- **Test 1 (the exact #854 acceptance criterion)**: seeds re1/re2 with `/custom/...` paths, calls re-seed, asserts (a) `$(cat /custom/...)` survives, (b) the default path does NOT appear (preservation, not append), (c) discovery section was refreshed (GITHUB.md present), (d) head/dev don't get a spurious GH_TOKEN line. Four distinct assertions on the same write.
- **Test 2**: default-path projects round-trip unchanged. The extractor pulls the default path back out, which substitutes back to the default — no perturbation. Covers the no-regression AC.
- **Test 3**: brand-new worktree (no AGENTS.md) with `cfg.reviewer_token_path` set → that cfg path is used. Covers AC bullet "Existing AGENTS.md or config token path is preferred over the hardcoded default when no body override is provided."
- **Test 4**: explicit `body.reviewerTokenPath` beats extraction. Covers the operator-override path.

**Suite + build**
- `node server/routes.extractReviewerTokenPath.test.js` → 15/15 pass.
- `node server/routes.reseedTokenPreservation.test.js` → 4/4 pass.
- `npm test` → **28 passed, 0 failed, 2 skipped** (+2 from the new files, auto-discovered by #836).
- `npm run build` → ✓.

**Acceptance criteria — all met**
- [x] Custom reviewer token path preserved verbatim, not reset to default (test 1).
- [x] Default-path projects unaffected (test 2).
- [x] Existing AGENTS.md or config token path preferred over hardcoded default (extraction order + test 3).
- [x] Parses common shell forms robustly: `export GH_TOKEN=$(cat /path)`, `export GH_TOKEN="$(cat /path)"`, quoted paths (extractor tests 1, 3, 4, 5, 6).
- [x] Endpoint-level test seeds re1/re2 with `/custom/token`, calls the endpoint, asserts the rewritten files still contain `/custom/token` while discovery sections update (test 1).
- [x] `npm run build` passes.

**Minor notes (non-blocking)**
- Extractor test 13 is documented as round-tripping the literal `{{reviewer_token_path}}` placeholder. The comment claims this is "functionally equivalent to using the default" — technically the resulting AGENTS.md would contain the literal placeholder, not the default path, because substitution then replaces `{{reviewer_token_path}}` with `{{reviewer_token_path}}` (no-op). The bug stays but doesn't worsen — acceptable, since this only happens if a prior re-seed somehow round-tripped without substituting (corrupted-by-something else). Worth tweaking the comment if anyone refactors.
- Endpoint test stubs `fs.readFileSync` globally, but only intercepts CONFIG_PATH and delegates everything else to `realReadFileSync`. Safe within a single node process (the tests run sequentially in an IIFE) and across processes (#836 runs each test file in its own node invocation). No cross-test bleed.
- Temp dirs from `fs.mkdtempSync` aren't cleaned up at the end of the run. Minor hygiene, not a correctness issue.
- The `routes.router || routes` import handles both default-export styles — this is a `routes.js` default export so it works today.

Carried-forward non-blockers from #853/#857 still apply (partial-write risk if an unknown agent key has no seed template, dual `re1`+`reviewer1` mixed-config edge) and are not touched by this PR.

No blockers from re2. The preservation contract is correctly implemented at parse, resolution, and endpoint levels.

Issue #809, opened 2026-05-29
[#805] Seed updates — agent discovery reads GITHUB.md instead of gh


realproject7:
Part of #805 (step 4). **Behavioral cutover — roll out per team at a batch boundary, monitor.**

## Goal
Replace agents' GraphQL **discovery** (`gh pr list` / `gh issue list`) with reading `GITHUB.md` / `GET /api/github-parsed`. All **writes** and correctness-critical reads stay direct, live `gh`.

## Scope — edit 5 seed files (`templates/seeds/{head,dev,re1,re2}.AGENTS.md`, `butler.CLAUDE.md`)
| Agent | Discovery → file | Stays direct, LIVE `gh` |
|---|---|---|
| HEAD | read GITHUB.md for the board | `gh pr merge`; **mandatory `gh pr view --json reviewDecision,reviews` re-read immediately before merge**; pre-create **duplicate-issue check**; `gh issue create/edit` |
| DEV | read GITHUB.md for assigned work | `gh pr create`, `gh issue view <n>` |
| RE1/RE2 | read GITHUB.md `## Open PRs` for @mentioned PRs | `gh pr review` (reviewer token), `gh pr diff`/`view`, **`gh pr checks` (CI status — never APPROVE off cached CI)** |
| BUTLER | read per-project GITHUB.md | `gh issue create/edit`, `gh release create` |

## Hard requirements (fold these into the seeds, not as caveats)
- **HEAD keeps the chat-derived two-reviewer gate** AND re-confirms approvals via the live `gh pr view` above before `gh pr merge`. `## Review Detail` is advisory only. Branch protection is the server-side backstop.
- **By-number fallback:** when @mentioned with a number, act on it directly via single-object `gh` — never gate existence on the file. When expected work is **missing from the file, or the file is stale > 2 cycles**, do a targeted `gh pr view <n>` / `gh issue view <n>` (mirror #810) — **never conclude "no work."**
- **Atomic per-team cutover:** re-seed + restart a team at a **batch boundary**, never mid-batch (avoid mixed-state where some agents read the file and some still use gh).

## MUST NOT change
- Writes (`pr merge/create/review`, `issue create/edit`, `release create`) stay direct `gh` with correct reviewer tokens.
- Reviewers' `gh pr diff`/`view`/`checks` stay direct & live (no stale-code or stale-CI verdicts).

## Acceptance criteria
[...]

PR #859, opened 2026-05-30
[#856] Auto-reseed worktree AGENTS.md on version upgrade

### LINKED ISSUE #856: Auto-reseed worktree AGENTS.md on version upgrade (so existing users don't re-seed each project manually)
## Why
#845 added a **manual** "Re-seed AGENTS.md" button, so after upgrading QuadWork an existing user must re-seed **every project by hand** to get the new agent seeds (e.g. the #809 GITHUB.md-discovery change). That's the only way existing projects pick up seed improvements — new projects get them at creation. For a smooth release, the upgrade should re-seed automatically.

## What
On server startup, detect a **package-version change** (store the last-run version in config or a marker file; compare to `require('../package.json').version`). When it changed, **auto-re-seed all configured projects' worktree AGENTS.md** from the current templates — reusing the #845 re-seed path (which must already: use each agent's config `cwd` per #855, substitute placeholders, preserve operator `## ` sections, and respect the batch-active guard).

Safety:
- **Only at a safe boundary** — skip/defer any project with an active batch (same guard as the manual endpoint); re-seed it on a later startup or via the manual button.
- New seed content only takes effect on the **next agent restart** (it's re-read on spawn), so auto-reseed is non-disruptive to a running session.
- Log what was re-seeded / deferred.

## Depends on
- **#855** (re-seed must resolve worktrees from config `cwd`) and **#854** (don't clobber custom token paths) — auto-reseed amplifies both bugs across all projects, so land those first.

## Acceptance criteria
- [ ] On a version bump, all projects' worktree AGENTS.md auto-update to current templates (placeholders preserved, operator notes preserved) without manual clicking
- [ ] Projects with an active batch are deferred (not re-seeded mid-batch); logged
- [ ] No re-seed when the version is unchanged
- [ ] Existing users upgrading get GITHUB.md-discovery seeds automatically (verify: post-upgrade, worktree AGENTS.md reference GITHUB.md discovery as primary and drop the old positive gh-pr-list instruction)
- [ ] `npm run build` passes

---

## RE1 review — accepted refinements
- **Per-project reseed state (LOAD-BEARING):** track reseed completion **per project** for the target package version, NOT a single global last-run version. Otherwise: bump to 2.2.0 -> re-seed project A -> defer project B (active batch) -> write global `last_reseed=2.2.0` -> **B never auto-reseeds**, silently stuck on old seeds.
  - A project deferred (active/unknown batch) stays **PENDING** for that target version.
  - On a later startup (or a safe-boundary hook if added), pending projects are retried.
  - The per-project marker is updated **only after that project's re-seed succeeds**.
- **Acceptance wording (corrected):** same as #855 — assert "references GITHUB.md discovery as primary; no longer the positive gh-pr-list instruction," not "not `gh pr list`."

realproject7:
## Summary

- Fixes #856. Existing users upgrading the QuadWork package never picked up new seed-template content (e.g. the #809 GITHUB.md discovery instructions) because the #845 re-seed endpoint required an operator click. This change auto-re-seeds on every startup when `package.json` version differs from the per-project completion state.
- **Per-project state is load-bearing** (per the issue's explicit criterion): a project deferred mid-batch stays pending in `~/.quadwork/reseed-state.json` so the very next startup (or any future safe-boundary hook) retries it. A single global "last run version" marker would strand any deferred project — once the marker advanced past the upgrade version, the deferred project would never be re-seeded.
- The per-target write loop is extracted to `_performReseedWrites(project, cfg, opts)` so the manual route handler (`POST /api/projects/:project/reseed-agents`) and the new `autoReseedOnStartup` hook share one implementation. Manual endpoint behavior, response shape, and error mapping are unchanged.

## Per-project decision matrix

| Condition                            | Action      | State change                    |
|--------------------------------------|-------------|---------------------------------|
| `state[id] === current version`      | `skip`      | none                            |
| Batch-state check throws             | `deferred`  | none (fail-closed, retry next)  |
| Batch-state returns `null`           | `deferred`  | none (fail-closed, retry next)  |
| Batch active                         | `deferred`  | none (retry next startup)       |
| `_performReseedWrites` throws        | `error`     | none (retry next startup)       |
| Writes succeed                       | `reseeded`  | `state[id] = current version`   |

## State file

`~/.quadwork/reseed-state.json`:
```json
{ "completedByProjectVersion": { "<projectId>": "<pkg version>" } }
```

## Wiring

`server/index.js` calls `routes.autoReseedOnStartup(startupCfg)` inside the existing `server.listen` async callback, right after `runStartupMigrations`. Errors are caught and logged — boot must not fail because of a re-seed issue.

## Acceptance criteria

- [x] On a version bump, projects' worktree `AGENTS.md` files auto-update to current templates without manual clicking, preserving placeholders and operator notes (`_performReseedWrites` reuses #845 merger + #854 token preservation + #855 cwd resolution).
- [x] Projects with an active batch are deferred, not re-seeded mid-batch; the deferral is logged with a `will retry on next startup` hint.
- [x] Per-project reseed state is load-bearing — a deferred project stays pending and retries on a later startup (tested explicitly via the cross-boot retry scenario).
- [x] No re-seed when that project has already completed reseed for the current version (tested with `skip` action returning instead of writing).
- [x] Existing upgraded users get `GITHUB.md` discovery as primary instruction; old positive `gh pr list` instruction is replaced (end-to-end test writes a pre-#809 stale AGENTS.md and confirms post-reseed content).
- [x] `npm run build` passes.

## Test plan

- [x] `node server/routes.autoReseed.test.js` covers: state load (missing / malformed / wrong-shape / round-trip / non-string-version stripping), `_readPackageVersion` (missing/malformed/no-version/valid), decision matrix (already current, fresh install, batch active, batch unknown via null and via throw, write throws, mix), cross-boot retry, missing-working-dir silent skip, end-to-end real-write with pre-#809 stale templates.
- [x] `npm test` → 29 passed, 0 failed, 2 skipped.

Closes #856

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1904,6 +1904,16 @@ server.listen(PORT, "127.0.0.1", async () => {
 
   runStartupMigrations(startupCfg);
 
+  // #856: Auto-reseed worktree AGENTS.md when the package version changes.
+  // Per-project completion state lives in ~/.quadwork/reseed-state.json, so
+  // projects deferred mid-batch stay pending and retry on the next startup.
+  // Failures here MUST NOT block server boot.
+  try {
+    await routes.autoReseedOnStartup(startupCfg);
+  } catch (err) {
+    console.error(`[reseed] auto-reseed failed: ${err.message}`);
+  }
+
   if (startupCfg.butler && startupCfg.butler.enabled && startupCfg.butler.auto_start) {
     const result = spawnButlerPty();
     if (result.ok) console.log(`[butler] auto-started (PID: ${result.pid})`);
--- a/server/routes.autoReseed.test.js
+++ b/server/routes.autoReseed.test.js
@@ -1,0 +1,386 @@
+// #856 tests for autoReseedOnStartup + the per-project state file.
+//
+// Layers covered:
+//   - _loadReseedState / _saveReseedState round-trip + corruption tolerance.
+//   - _readPackageVersion fallback on missing/malformed package.json.
+//   - autoReseedOnStartup decision matrix with injected `getProgress` and
+//     `performWrites`:
+//       * project already at current version → skip (no write to state).
+//       * project at older version, batch idle → reseed (state advanced).
+//       * project at older version, batch active → defer (state UNCHANGED
+//         so the next startup retries).
+//       * project at older version, batch state unknown (null result OR
+//         throw) → defer (fail-closed gate, state UNCHANGED).
+//       * project with no working_dir → silently skipped.
+//       * write throws → state UNCHANGED (so retry on next startup).
+//       * deferred-then-resolved across two boots → second boot re-seeds
+//         AND advances state for that project; already-current projects
+//         from boot 1 remain skipped on boot 2.
+//   - End-to-end with real `_performReseedWrites` against temp worktree
+//     dirs — verifies the upgrade flow actually writes new templates AND
+//     the resulting AGENTS.md has GITHUB.md discovery (the #856 user
+//     observable: "existing upgraded users get GITHUB.md as primary").
+//
+// Plain node:assert script — auto-discovered by the #836 runner. Run
+// directly with `node server/routes.autoReseed.test.js`.
+
+const assert = require("node:assert/strict");
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
+const crypto = require("crypto");
+
+const {
+  autoReseedOnStartup,
+  _loadReseedState,
+  _saveReseedState,
+  _readPackageVersion,
+  _performReseedWrites,
+} = require("./routes");
+
+function tmp(label) {
+  return fs.mkdtempSync(path.join(os.tmpdir(), `qw-856-${label}-`));
+}
+
+function quietLog() {
+  const lines = [];
+  return { log: (m) => lines.push(m), lines };
+}
+
+(async () => {
+  // ── State file: missing → empty default ──────────────────────────────────
+  {
+    const missing = path.join(tmp("state-missing"), "absent.json");
+    assert.deepEqual(_loadReseedState(missing), { completedByProjectVersion: {} });
+  }
+
+  // ── State file: malformed JSON → empty default (don't crash boot) ────────
+  {
+    const dir = tmp("state-bad");
+    const p = path.join(dir, "state.json");
+    fs.writeFileSync(p, "not-json-at-all{");
+    assert.deepEqual(_loadReseedState(p), { completedByProjectVersion: {} });
+  }
+
+  // ── State file: wrong-shape value (array, number, missing key) → empty ──
+  {
+    const dir = tmp("state-wrong");
+    const p = path.join(dir, "state.json");
+    fs.writeFileSync(p, JSON.stringify({ completedByProjectVersion: [1, 2, 3] }));
+    assert.deepEqual(_loadReseedState(p), { completedByProjectVersion: {} });
+
+    fs.writeFileSync(p, JSON.stringify({ somethingElse: { x: "y" } }));
+    assert.deepEqual(_loadReseedState(p), { completedByProjectVersion: {} });
+
+    // Non-string version values get stripped (defensive against a future
+    // writer that puts numbers/objects in — never compares equal to a
+    // package.json string).
+    fs.writeFileSync(p, JSON.stringify({ completedByProjectVersion: { a: "1.0.0", b: 2, c: { v: "3" } } }));
+    assert.deepEqual(_loadReseedState(p), { completedByProjectVersion: { a: "1.0.0" } });
+  }
+
+  // ── State file: round-trip ───────────────────────────────────────────────
+  {
+    const dir = tmp("state-roundtrip");
+    const p = path.join(dir, "state.json");
+    _saveReseedState({ completedByProjectVersion: { proj1: "2.1.0", proj2: "2.0.5" } }, p);
+    assert.deepEqual(_loadReseedState(p), { completedByProjectVersion: { proj1: "2.1.0", proj2: "2.0.5" } });
+  }
+
+  // ── _readPackageVersion ──────────────────────────────────────────────────
+  {
+    const dir = tmp("pkg");
+    // Missing → 0.0.0 fallback.
+    assert.equal(_readPackageVersion(path.join(dir, "missing.json")), "0.0.0");
+    // Malformed → 0.0.0.
+    const bad = path.join(dir, "bad.json");
+    fs.writeFileSync(bad, "{not json");
+    assert.equal(_readPackageVersion(bad), "0.0.0");
+    // Missing version field → 0.0.0.
+    const noVer = path.join(dir, "no-ver.json");
+    fs.writeFileSync(noVer, JSON.stringify({ name: "x" }));
+    assert.equal(_readPackageVersion(noVer), "0.0.0");
+    // Valid version → returned verbatim.
+    const ok = path.join(dir, "ok.json");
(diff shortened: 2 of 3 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
PR #859 implements auto-reseed on startup with per-project completion state, reusing the same write path as the manual re-seed endpoint. The deferred-project behavior is correct: active/unknown batch state leaves that project unmarked so it retries on a later startup.

### Findings
- No blocking findings.

### Decision
APPROVE. The implementation satisfies #856: `autoReseedOnStartup` compares each configured project against `~/.quadwork/reseed-state.json`, skips only projects already completed for the current package version, defers active or unknown batch states without advancing state, and marks a project complete only after `_performReseedWrites` succeeds. The shared write loop preserves the #845/#854/#855 behavior: cwd-based targeting, token-path preservation, placeholder substitution, and operator-note merging. The added tests cover state corruption tolerance, version fallback, the decision matrix, cross-boot retry for deferred projects, and an end-to-end stale `gh pr list` -> GITHUB.md refresh. GitHub reports no checks on the branch, so this approval is based on live diff review plus the reported `npm test` / `npm run build` results.

realproject7 (commented):
APPROVE — verified on PR branch (head `a0e9fa1`).

**Architecture: `_performReseedWrites` extraction**
- Pure per-target write loop pulled out of the route handler, now shared between manual `/api/projects/:project/reseed-agents` and the new `autoReseedOnStartup` orchestrator. Manual endpoint behavior + JSON shape + 500-on-missing-seed contract preserved (the throw is caught and remapped to `res.status(500)`). The #855/#854 token-path resolution and #845 merger contracts pass through unchanged — same `opts.reviewerTokenPath || _extract... || cfg... || default` chain.
- Auto-reseed deliberately passes no `reviewerUser` / `reviewerTokenPath` overrides, so each project naturally inherits per-agent extraction → cfg → default. This is the right design: an unattended boot has no operator override signal.

**`autoReseedOnStartup` decision matrix — load-bearing per-project state**
- Up-to-date check (`state.completedByProjectVersion[project.id] === version`) happens BEFORE the gh-touching `getProgress` call. So in steady-state (post-first-boot at a given version), every project short-circuits without a gh call — important given [feedback_gh_uses_graphql.md] / `quadwork_graphql_drain` rate-limit context. Only the first boot after a version bump hits gh, once per pending project.
- Fail-closed batch gate matches the #853 contract exactly: throw OR `null` progress OR `null` from `isActiveFromProgress` all defer with `state UNCHANGED`. Only `active === false` proceeds to write. So a project deferred mid-batch genuinely stays pending — the `completedByProjectVersion[projectId]` entry is only written on a successful write, not on a defer.
- Write throw → `action: "error"`, state UNCHANGED, retry next boot. Correct — partial-write idempotency from the #845 merger means a re-seed retry is safe.

**Per-project state file (`~/.quadwork/reseed-state.json`)**
- Shape `{ completedByProjectVersion: { "<projectId>": "<pkg version>" } }`. A single global "last version" marker would have stranded any deferred project — using per-project entries correctly satisfies the issue's explicit criterion.
- `_loadReseedState` is paranoid: null/undefined `parsed`, wrong-shape `completedByProjectVersion` (array/number/missing), and non-string values per project key are all stripped to `{}` without throwing. The non-string-stripping is a defensive nice-to-have since a future writer that puts numbers/objects in would never compare-equal to a `package.json` string anyway.
- `_saveReseedState` uses `fs.mkdirSync(..., { recursive: true, mode: 0o700 })` and `fs.writeFileSync(JSON.stringify(...))`. Comment claims "write-replace is atomic enough" — strictly this is `truncate + write + close` and not crash-atomic. But the loader handles malformed JSON, and worst case is an extra re-seed on the next boot (which is idempotent thanks to the #845 merger). Worth a minor note (see below) but acceptable.

**Boot wiring (`server/index.js`)**
- Called after `runStartupMigrations`, before butler/watchdog start. `await`ed and wrapped in try/catch with a logged-and-continue policy. Boot cannot fail because of a re-seed issue. ✓
- Trade-off: butler/watchdog now wait for auto-reseed to finish. In steady state this is one synchronous state-file read + an O(n_projects) loop of short-circuit checks (no gh calls), so negligible. On a version-bump boot it's one gh call per pending project (sequential await), so for typical 1-5 project setups boot is ~1-5s slower. Acceptable for a once-per-version-bump event; flagging as a non-blocker.

**Tests (`server/routes.autoReseed.test.js`)**
- **State load**: missing / malformed JSON / wrong-shape value / non-string-stripping / round-trip — comprehensive corruption tolerance.
- **Version reader**: missing file / malformed JSON / no `version` field / valid → "0.0.0" fallback chain is defensive against any unexpected package.json state at boot.
- **Decision matrix** (cases 1-8): all five branches of the orchestrator covered with injected `getProgress` / `performWrites`. Case 4 specifically asserts deferred → state UNCHANGED + retry-hint log line. Case 7 specifically asserts write-throws → state UNCHANGED. Case 8 mixes all three actions in one boot to verify independence.
- **Cross-boot retry (case 9)** — the load-bearing #856 contract: boot 1 defers `later` (busy), state file shows no `later` entry; boot 2 with batch idle reseeds `later` and advances state. The exact regression scenario the issue calls out. ✓
- **End-to-end against real `_performReseedWrites`**: writes a pre-#809 stale AGENTS.md with `Use \`gh pr list\` to find open PRs in this repo.`, runs auto-reseed, asserts (a) `GITHUB.md` present, (b) the stale positive instruction is gone, (c) state advanced to `9.9.9`. This is the user-visible acceptance criterion verified end-to-end.

**Suite + build**
- `node server/routes.autoReseed.test.js` → all assertions pass.
- `npm test` → **29 passed, 0 failed, 2 skipped** (+1 from the new file, auto-discovered by #836).
- `npm run build` → ✓.

**Acceptance criteria — all met**
- [x] On a version bump, AGENTS.md auto-updates without manual click; placeholders + operator notes preserved (delegates to the verified #845 merger + #854 token resolver + #855 resolver).
- [x] Projects with an active batch are deferred + logged with retry hint (decision matrix case 4 + e2e log assertion).
- [x] Per-project state is load-bearing (file shape + cross-boot test 9).
- [x] No re-seed when project already at current version (orchestrator early-return + test 1 / test 3 second-run).
- [x] Existing upgraded users get GITHUB.md discovery, stale positive `gh pr list` instruction is replaced (e2e test 11).
- [x] `npm run build` passes.

**Minor notes (non-blocking)**
- `_saveReseedState` comment "write-replace is atomic enough" is technically wrong — `fs.writeFileSync` is truncate-then-write-then-close, not atomic on crash. But `_loadReseedState`'s malformed-JSON tolerance is the real backstop, and re-seed is idempotent via the #845 merger, so the worst-case corruption recovery is one extra re-seed on the next boot. Could be hardened later with write-to-temp-then-rename if anyone reports torn-write state corruption.
- Auto-reseed blocks butler start. On a version-bump boot with N pending projects this is N sequential gh calls. For typical 1-5 project setups, negligible (~1-5s); for an outlier user with many projects, noticeable. Future hardening could `Promise.all` the gh calls, but the sequential pattern also serves as natural rate-limit backpressure per [quadwork_graphql_drain].
- Carried-forward non-blockers from #853/#857/#858 still apply (partial-write risk on unknown agent keys with missing seed templates, `{{reviewer_token_path}}` round-trip edge, dual re1+reviewer1 mixed-config edge) — not touched here.

No blockers from re2. The orchestrator's decision matrix is clean, per-project state correctly avoids the global-marker stranding bug, and the end-to-end test verifies the user-visible upgrade behavior.

Issue #866, opened 2026-05-30
Show GitHub rate-limit status badge (green/orange/red) in the GITHUB panel header


realproject7:
## Summary
Show an always-on GitHub rate-limit status badge in the **GITHUB panel header (top-right)** displaying **all three buckets (core / graphql / search)**, each colored independently **green → orange (warning) → red (gone)** by its own remaining budget, so the operator can see the full rate-limit situation at a glance instead of only finding out via the alert banner once it's already low.

Most of this is already built — this is a display addition on top of the existing `#554` plumbing.

## Already in place (reuse, don't rebuild)
- Backend `GET /api/github/rate-limit` (`server/routes.js:935`) returns `{ limit, remaining, resetInMinutes, low, critical, resetAt }`, populated from a `gh api rate_limit` poll (the `rate_limit` endpoint is **exempt** — zero budget cost).
- `GitHubPanel.tsx` already fetches it every 60s and holds `RateLimitInfo` state (`#554`).
- `PanelHeader` already has a right-aligned `children` slot (`#523`) → the badge mounts top-right of the GITHUB header (`src/components/ProjectDashboard.tsx:318`).
- Existing thresholds: `isRateLow()` / `isRateLimited()` (`server/routes.js:105-108`).

### Backend (small — `server/routes.js`)
The `gh api rate_limit` response already contains all buckets. Where `_rateLimit` is populated from `data.core` (`routes.js:76-78`), also stash `data.graphql` and `data.search`, and expose them in the `/api/github/rate-limit` response, e.g.:
```json
{ "core": {…}, "graphql": { "limit": 5000, "remaining": 4711, "resetInMinutes": 42 }, "search": {…},
  "low": false, "critical": false }
```
Keep the existing top-level `limit/remaining/low/critical` (core) for backward-compat with the current banner.

### Frontend (new component + 1 line wiring)
- New `src/components/GitHubRateLimitBadge.tsx` — self-contained: fetches `/api/github/rate-limit` on the same 60s cadence, renders a compact inline badge.
- **Display all three buckets inline**, each labeled and individually colored by its OWN remaining %: `core`, `graphql` (`gql`), `search`. Compact monospace, dev-tool density — e.g.:
  ```
  ● core 4982/5000   ● gql 4711/5000   ● search 28/30
  ```
[...]

PR #889, opened 2026-05-30
[#886] Rate-limit badge: also show the reviewer-token account

### LINKED ISSUE #886: Rate-limit badge: also show the reviewer-token account (close #866 monitoring blind spot)
## Problem
The #866 rate-limit badge only monitors **one** GitHub account — whichever token the server's `gh` runs as (the head/dev account, e.g. `realproject7`). In a two-account setup (reviewers use a separate `reviewer_token_path` account, e.g. `project7-interns`), the **reviewer account is invisible to the dashboard**.

That account is the *historically riskiest* for the GraphQL budget — review discovery used to mean `gh pr list`. If a regressed seed ever reintroduced GraphQL discovery on the reviewers, the badge would show all-green while the reviewer budget drained, because GitHub rate limits are **per-account**, not aggregated. The badge currently watches the *less* risky account.

Verified live: `realproject7` graphql 182/5000 vs `project7-interns` graphql 37/5000 — two independent budgets, only the first is shown.

## Fix
Extend #866 to also poll and display the **reviewer-token account**:

### Backend (`server/routes.js`)
- In the existing rate-limit poll (the timer that runs `gh api rate_limit`), make a **second** call with the reviewer token: `GH_TOKEN=$(cat <reviewer_token_path>) gh api rate_limit`. `rate_limit` is **exempt** — still zero billable cost.
- Resolve the reviewer token path with the **existing chain** already used by reseed: `cfg.reviewer_token_path` → `_extractReviewerTokenPath(...)` → default. If no reviewer token is configured, skip the second poll entirely.
- Expose it additively under a new `reviewer` key in `/api/github/rate-limit`, e.g. `reviewer: { login?, core, graphql, search }` (reuse `bucketView`). Optionally include the account `login` for the row label.

### Frontend (`src/components/GitHubRateLimitBadge.tsx`)
- When the response includes a `reviewer` block, render a **second labeled row/group** for it (at minimum `graphql`; core/search optional), colored per-bucket by the same thresholds.
- Label it by login if present (e.g. `interns: gql 4963/5000`), else a generic "reviewer".
- When absent (single-account setups), render nothing extra — current behavior unchanged.

## Acceptance criteria
- [ ] `/api/github/rate-limit` includes a `reviewer` bucket set when a reviewer token is configured; omitted otherwise
- [ ] Reviewer `rate_limit` poll uses the reviewer token, is exempt (no billable calls), and reuses the existing token-path resolution
- [ ] Badge shows a second row for the reviewer account, per-bucket colored; hidden when no reviewer token
- [ ] Single-account setups render exactly as today (no regression)
- [ ] `npm run build` passes

## Notes
- Small, additive follow-up to #866. No new billable API cost (`rate_limit` is exempt).
- Closes the monitoring blind spot: the dashboard then covers both GraphQL budgets, not just the server's.

realproject7:
## Summary
Closes the #866 monitoring blind spot. The rate-limit badge only watched the **server's** gh account (head/dev token). In a two-account setup the **reviewer account** — historically the riskiest for the GraphQL budget (review discovery used to mean `gh pr list`) — was invisible, and GitHub rate limits are **per-account, not aggregated**. So a regressed reviewer seed could drain the reviewer GraphQL budget while the badge stayed all-green.

Additive follow-up to #866. **No new billable cost** — `rate_limit` is exempt.

## Backend (`server/routes.js`)
- **`refreshReviewerRateLimit()`** — a **second exempt** `gh api rate_limit` poll run as the **reviewer token** (`GH_TOKEN` env), called from `refreshRateLimit` each cycle in its own try/catch so a reviewer failure never affects the main poll. Resolves the token path the same way reseed does (`cfg.reviewer_token_path` → default `~/.quadwork/reviewer-token`); **no/empty token → reviewer state cleared** (key omitted).
- `_reviewerRateLimit` holds `{ login, core, graphql, search }` (buckets in ms); `login` from `cfg.reviewer_github_user` (no extra API call).
- `/api/github/rate-limit` adds a **`reviewer`** key (reusing `bucketView`) **only when a reviewer token resolves**; single-account setups omit it entirely.

## Frontend (`src/components/GitHubRateLimitBadge.tsx`)
- When the response has a `reviewer` block, render a **second labeled group** (`<login>: ● gql … ● core … ● search …`) after a divider, per-bucket colored by the same thresholds — `graphql` first (the at-risk budget). Absent → renders exactly as today.
- Refactored bucket rendering into a shared `bucketSpans()` helper.

## Verification
- **`server/routes.reviewerRateLimit.test.js` — 11/11:** reviewer poll uses the reviewer token via `GH_TOKEN`; payload present with `login` + buckets when configured; **null + no second poll** when the token file is missing/empty (single-account, no regression). (Stubs execFile's custom-promisify; settles the module-load poll for determinism.)
- `reviewBatch.test.js` updated: stub the reviewer-token path to ENOENT so the new module-load reviewer poll can't race a gh call into its ZERO-gh-call assertions — still **42/42**.
- Endpoint verified live: single-account (no token) → **`reviewer` key omitted**, core/graphql/search unchanged.
- `npm test` → **35 passed, 0 failed, 2 skipped**; `npm run build` → green; badge lints clean.

## Dependencies
- #866 (merged) — the badge + `bucketView` + exempt poll.

Closes #886.

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -67,6 +67,13 @@ const _searchRateLimit = {
   resetAt: 0,        // epoch ms
   updatedAt: 0,      // epoch ms when we last fetched
 };
+// #886: the reviewer-token account is a SEPARATE, per-account GitHub rate-limit
+// budget (reviewers run gh as `reviewer_github_user`). Polled with the reviewer
+// token in a second exempt `rate_limit` call. Null until a reviewer token is
+// configured + resolves — single-account setups stay null and the badge shows
+// nothing extra. Shape: { login, core, graphql, search } where each bucket is
+// { limit, remaining, resetAt(ms) }, plus updatedAt/error.
+let _reviewerRateLimit = null;
 const RATE_LIMIT_POLL_MS = 60_000;       // refresh every 60s
 const RATE_LIMIT_LOW_THRESHOLD = 200;    // below this → back off
 const RATE_LIMIT_CRITICAL = 50;          // below this → stop all infra gh calls
@@ -103,6 +110,52 @@ async function refreshRateLimit() {
   } catch (err) {
     _rateLimit.error = err.message;
     _rateLimit.updatedAt = Date.now();
+  }
+  // #886: also poll the reviewer-token account (separate budget). Independent
+  // try/catch inside — a reviewer failure never affects the main poll above.
+  await refreshReviewerRateLimit();
+}
+
+// #886: second exempt `rate_limit` poll as the reviewer-token account, so the
+// dashboard covers BOTH GitHub budgets (per-account, not aggregated). Resolves
+// the reviewer token path the same way reseed does (configured path → default
+// ~/.quadwork/reviewer-token); no token → clears the reviewer state (omitted
+// from the payload, single-account behavior unchanged). `rate_limit` is exempt,
+// so this adds zero billable cost.
+async function refreshReviewerRateLimit() {
+  let cfg = {};
+  try { cfg = JSON.parse(fs.readFileSync(CONFIG_PATH, "utf-8")); } catch { /* default-config / unreadable → use {} */ }
+  const tokenPath = cfg.reviewer_token_path || path.join(os.homedir(), ".quadwork", "reviewer-token");
+  let token = "";
+  try {
+    token = fs.readFileSync(tokenPath, "utf-8").trim();
+  } catch {
+    _reviewerRateLimit = null; // no reviewer token configured → single-account
+    return;
+  }
+  if (!token) { _reviewerRateLimit = null; return; }
+  try {
+    const { stdout } = await _execFileAsync("gh", [
+      "api", "rate_limit", "--jq",
+      "{core: (.resources.core | {limit,remaining,reset}), graphql: (.resources.graphql | {limit,remaining,reset}), search: (.resources.search | {limit,remaining,reset})}",
+    ], { encoding: "utf-8", timeout: 10000, env: { ...process.env, GH_TOKEN: token } });
+    const data = JSON.parse(stdout);
+    const toBucket = (b) => ({ limit: b.limit, remaining: b.remaining, resetAt: b.reset * 1000 });
+    _reviewerRateLimit = {
+      login: cfg.reviewer_github_user || null,
+      core: data.core ? toBucket(data.core) : null,
+      graphql: data.graphql ? toBucket(data.graphql) : null,
+      search: data.search ? toBucket(data.search) : null,
+      updatedAt: Date.now(),
+      error: null,
+    };
+  } catch (err) {
+    // Preserve prior bucket data if any; just flag the error + timestamp.
+    _reviewerRateLimit = {
+      ...(_reviewerRateLimit || { login: cfg.reviewer_github_user || null, core: null, graphql: null, search: null }),
+      updatedAt: Date.now(),
+      error: err.message,
+    };
   }
 }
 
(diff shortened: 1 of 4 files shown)


realproject7:
**Review verdict from @re2: Approve ✅** (comment — PR shares my GH account)

Backend + frontend; I ran both server tests and a real frontend typecheck, and checked the token-handling closely since this reads a credential.

**No token leak (checked deliberately):**
- The reviewer token is read into a local `token` var and passed ONLY as `GH_TOKEN` in the subprocess `env` — never stored on `_reviewerRateLimit` (which holds `{login, core, graphql, search, updatedAt, error}`) and never returned.
- `reviewerRateLimitPayload()` exposes only `{login?, core?, graphql?, search?}` — it omits even `error`/`updatedAt`, so a gh error message (which in any case wouldn't contain the env-passed token) can't reach the client. ✅
- `_execFileAsync` error messages carry the command/args, not `env`, so the catch's `err.message` won't echo the token. ✅

**Backend correctness:**
- Second poll is the exempt `rate_limit` endpoint → zero billable cost, as claimed. Own internal try/catch + invoked after the main poll's catch, so reviewer failure ⇏ main-poll failure and vice versa. ✅
- Config + token re-read each cycle (picks up config changes without restart); path resolution (`cfg.reviewer_token_path` → `~/.quadwork/reviewer-token`) mirrors reseed.
- No-token / ENOENT / empty(whitespace) → `_reviewerRateLimit = null` → `reviewer` key omitted → single-account behavior unchanged. **Removing** the token clears the block (doesn't get stuck stale). ✅
- On a transient gh failure after a prior success, it preserves last-known buckets + flags error — reasonable.

**Frontend correctness:** `bucketSpans()` DRY refactor reused for both groups; the `mainSpans.length === 0 → null` graceful-degrade is preserved; reviewer group is gated on `reviewerSpans.length > 0`, graphql-first (the at-risk budget), `login || "reviewer"` fallback. Types (`ReviewerBlock`, `reviewer?` on the response) match the backend payload.

**Verification:**
- `routes.reviewerRateLimit.test.js` → **11/11** (poll uses the reviewer `GH_TOKEN`, payload present when configured, null + no poll when missing/empty).
- `routes.reviewBatch.test.js` → still **42/42** — the added `reviewer-token` ENOENT stub correctly prevents the new module-load reviewer poll from racing the zero-gh assertions.
- `tsc --noEmit` on the full project at PR HEAD `dba4ba2` → **exit 0, zero errors**.

**Non-blocking nits:** (1) `reviewerRateLimitPayload()` is invoked twice in the response spread — could cache in a const. (2) An *expired* (not removed) reviewer token keeps the last-known buckets with no client-visible error, so the badge could show stale-good numbers indefinitely — a small monitoring edge for this monitoring feature; worth a follow-up if you want the badge to surface a reviewer-poll error state. Neither blocks.

Clean, well-tested, no credential exposure. Good to merge.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The implementation closes the #866 blind spot by adding an additive reviewer-token rate-limit poll and badge group without changing single-account behavior. Token handling is appropriately scoped to the subprocess environment and the endpoint exposes only bucket data.

### Findings
- None.

### Decision
Approved. `refreshReviewerRateLimit()` performs a second exempt `gh api rate_limit` call with the reviewer token in `GH_TOKEN`, clears reviewer state when the token is missing/empty, and `/api/github/rate-limit` omits the `reviewer` block unless bucket data exists. The frontend keeps the original core/graphql/search rendering and adds a gated reviewer group with GraphQL first and the same per-bucket coloring. The new tests cover configured, missing, and empty-token cases, and the review-batch zero-gh tests were updated for the module-load poll. GitHub reports no checks, so approval is based on live diff/static review plus @dev's stated test/build verification.

Issue #893, opened 2026-05-31
Reviewer rate-limit badge (#886): resolve custom reviewer-token paths via the reseed resolver


realproject7:
## Problem (follow-up to #886)
`refreshReviewerRateLimit()` ([server/routes.js:128](server/routes.js#L128)) resolves the reviewer token path as only:
```js
cfg.reviewer_token_path || ~/.quadwork/reviewer-token
```

But the **actual** reviewer token path can be preserved per worktree by reseed via `_extractReviewerTokenPath(...)` (parsing `GH_TOKEN=$(cat <path>)` out of each reviewer worktree's `AGENTS.md`) — see the reseed resolver (~`server/routes.js:3541`) and the chain:
```js
opts.reviewerTokenPath || _extractReviewerTokenPath(existing) || cfg.reviewer_token_path || default
```
Setup also passes `reviewerTokenPath` into seed writing ([routes.js:3153](server/routes.js#L3153)) without persisting it to project config ([routes.js:3262](server/routes.js#L3262)).

**Result:** a project whose reviewer token lives at a **custom path** recorded only in the reviewer worktree `AGENTS.md` can get the wrong reviewer budget or no reviewer badge — leaving the #866 monitoring blind spot partially open. The default-path case, such as the current VPS `~/.quadwork/reviewer-token`, works, which is why this was not caught at runtime.

## Fix
Make the reviewer rate-limit lookup **project-aware** and resolve the token path from the selected project's configured reviewer worktree, not from an arbitrary project.

Recommended shape:
- Frontend `GitHubRateLimitBadge` passes the current `projectId`, e.g. `/api/github/rate-limit?project=<id>`.
- Backend validates the project id if supplied and resolves reviewer token path from that project's configured reviewer agent cwd (`re1` first, then `re2` if needed):
  ```js
  _extractReviewerTokenPath(<project.agents.re1.cwd>/AGENTS.md)
    || _extractReviewerTokenPath(<project.agents.re2.cwd>/AGENTS.md)
    || cfg.reviewer_token_path
    || ~/.quadwork/reviewer-token
  ```
- Keep the no-project request backward compatible by using the existing config/default fallback.

Do **not** put `cfg.reviewer_token_path` ahead of the worktree extraction for a project-specific request. Reseed preserves the worktree value first because it is the actual path currently used by that project's agents; reversing the order can keep monitoring the wrong token.

[...]

PR #898, opened 2026-06-01
[#897] Reviewer rate-limit: expand ~ in reviewer token path (P1 regression from #893)

### LINKED ISSUE #897: Reviewer rate-limit (#893): expand ~ in reviewer token path — default-seeded projects show no badge
## Problem (P1 regression — follow-up to #893)
#893's project-aware resolver breaks the **common default** reviewer-token case.

The setup wizard seeds the literal string `~/.quadwork/reviewer-token` (with a tilde) into reviewer `AGENTS.md` by default ([SetupWizard.tsx:327](src/components/SetupWizard.tsx#L327), [:494](src/components/SetupWizard.tsx#L494)). The new resolver:
1. reads the reviewer worktree `AGENTS.md` and calls `_extractReviewerTokenPath()` ([routes.js:139](server/routes.js#L139)) — which returns the **raw** `~/.quadwork/reviewer-token` string (it strips quotes but **not** the tilde, [routes.js:3415](server/routes.js#L3415)),
2. returns it as `{ path: extracted, source: "worktree" }` — which **wins before** the absolute `defaultPath` fallback is ever reached,
3. `refreshReviewerForPath()` passes it straight to `fs.readFileSync()` ([routes.js:155](server/routes.js#L155)).

**Node does not expand `~`** → `ENOENT` → the cache entry is dropped → the reviewer badge is **omitted** for every normal project seeded with the default path (even though `~/.quadwork/reviewer-token` exists). The old code worked only because its default was an already-absolute `path.join(os.homedir(), ...)`; the worktree-extracted tilde string now short-circuits that.

## Fix
Normalize the resolved reviewer token path before reading/caching — at minimum expand a leading `~/` (and bare `~`) to `os.homedir()`:
- Apply expansion to the resolver's **returned path** (covers `worktree` and `config` sources; `default` is already absolute). This also keeps the cache key consistent (`_reviewerRateLimitByPath` is keyed by path — `~/...` vs `/home/.../...` for the same file would otherwise be two entries).
- **Do NOT change `_extractReviewerTokenPath` itself** — reseed depends on its raw return value ([routes.js:3596](server/routes.js#L3596)); expand only in the rate-limit resolver / `refreshReviewerForPath`.

## Acceptance criteria
- [ ] A project whose reviewer `AGENTS.md` contains the default `~/.quadwork/reviewer-token` shows the reviewer badge (tilde expanded to `os.homedir()` before read)
- [ ] Leading `~/` and bare `~` are expanded; absolute paths pass through unchanged
- [ ] Custom **absolute** path still works (no regression); no reviewer token still omits the reviewer block
- [ ] Path is normalized before it's used as the `_reviewerRateLimitByPath` cache key
- [ ] `_extractReviewerTokenPath` is unchanged (reseed unaffected)
- [ ] New test in `routes.reviewerRateLimit.test.js`: project-specific `AGENTS.md` extraction resolving the **default `~/.quadwork/reviewer-token`** to a real file (current tests cover custom-absolute + no-project-default but miss this common case)
- [ ] `npm run build` + `npm test` pass

## Notes
- **P1 — blocks 2.3.0:** without it, the reviewer badge silently disappears for the common default-path setup, which is the opposite of #886/#893's goal. Land before the publish.

realproject7:
## Summary
**P1 regression fix — blocks 2.3.0.** #893's project-aware resolver broke the **common default** reviewer-token case.

The setup wizard seeds the literal string `~/.quadwork/reviewer-token` (with a tilde) into reviewer `AGENTS.md` by default. #893's resolver reads that worktree `AGENTS.md`, gets the **raw** `~/...` string back from `_extractReviewerTokenPath` (which strips quotes but not the tilde), and returns it as `{ path: extracted, source: "worktree" }` — which wins **before** the absolute default fallback. **Node's `fs` does not expand `~`** → `ENOENT` → the cache entry is dropped → the reviewer badge is **omitted for every default-seeded project** (even though `~/.quadwork/reviewer-token` exists). The old #886 code worked only because its default was an already-absolute `path.join(os.homedir(), ...)`.

## Fix
- Add `_expandTilde()` — leading `~/` and bare `~` → `os.homedir()`; everything else passes through.
- Apply it to the **resolver's returned path** for the `worktree` and `config` sources (`default` is already absolute). This normalizes both the read path and the `_reviewerRateLimitByPath` **cache key** (so `~/…` vs `/home/…/…` for the same file aren't two entries).
- **`_extractReviewerTokenPath` is intentionally unchanged** — reseed depends on its raw return value; expansion happens only in the rate-limit resolver.

## Verification
- **`routes.reviewerRateLimit.test.js` — +6 → 19/19:**
  - default-seeded `~/.quadwork/reviewer-token` resolves to the **absolute homedir path** (source `worktree`), and **the badge SHOWS** — it polls the token at the expanded path (the regression case the old tests missed);
  - `~/` and bare `~` expanded for the `config` source;
  - **absolute paths pass through unchanged** (custom-absolute path still works — no regression);
  - no-token still omits the reviewer block.
- `_extractReviewerTokenPath` confirmed unchanged (not in the diff) → reseed unaffected.
- `npm test` → **37 passed, 0 failed, 2 skipped**; `npm run build` → green.

## Note
Login follows #893's source rule unchanged: omitted for a `worktree`-resolved path (a custom token may be a different account). For a default-seeded project the badge shows the generic `reviewer:` label — strictly better than the bug (no badge at all), and the bucket monitoring (the point of #886/#893) is restored.

## Dependencies
- #893 (merged) — introduced the resolver this fixes.

Closes #897.

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -124,6 +124,18 @@ async function refreshRateLimit() {
 // AGENTS.md first, then re2's. Falls back to cfg.reviewer_token_path → default
 // for a no-project / unknown-project request (backward compatible with #886).
 // Returns { path, source: "worktree" | "config" | "default" }.
+// #897: expand a leading `~/` (and bare `~`) to os.homedir(). The setup wizard
+// seeds the literal `~/.quadwork/reviewer-token` into reviewer AGENTS.md, and
+// _extractReviewerTokenPath returns it raw (reseed depends on the raw value).
+// Node's fs does NOT expand `~`, so the resolver must normalize before reading /
+// caching — otherwise the default-seeded path ENOENTs and the badge vanishes.
+function _expandTilde(p) {
+  if (typeof p !== "string") return p;
+  if (p === "~") return os.homedir();
+  if (p.startsWith("~/")) return path.join(os.homedir(), p.slice(2));
+  return p;
+}
+
 function _resolveReviewerTokenPath(projectId) {
   let cfg = {};
   try { cfg = JSON.parse(fs.readFileSync(CONFIG_PATH, "utf-8")); } catch { /* default-config / unreadable */ }
@@ -137,12 +149,14 @@ function _resolveReviewerTokenPath(projectId) {
       try {
         const agentsMd = fs.readFileSync(path.join(t.wtDir, "AGENTS.md"), "utf-8");
         const extracted = _extractReviewerTokenPath(agentsMd);
-        if (extracted) return { path: extracted, source: "worktree" };
+        // #897: expand `~` here (not in _extractReviewerTokenPath — reseed needs
+        // the raw value) so the read + cache key use an absolute path.
+        if (extracted) return { path: _expandTilde(extracted), source: "worktree" };
       } catch { /* worktree / AGENTS.md missing → try next role */ }
     }
   }
-  if (cfg.reviewer_token_path) return { path: cfg.reviewer_token_path, source: "config" };
-  return { path: defaultPath, source: "default" };
+  if (cfg.reviewer_token_path) return { path: _expandTilde(cfg.reviewer_token_path), source: "config" };
+  return { path: defaultPath, source: "default" }; // already absolute
 }
 
 // #886/#893: poll one reviewer token path (exempt `rate_limit` call as that
--- a/server/routes.reviewerRateLimit.test.js
+++ b/server/routes.reviewerRateLimit.test.js
@@ -124,14 +124,43 @@ function reset() {
   r = _resolveReviewerTokenPath("ghost");
   ok(r.path === "/cfg/reviewer-token" && r.source === "config", "unknown project → cfg/default fallback (does not sample project 'p')");
 
+  // #897: default-seeded `~/.quadwork/reviewer-token` (tilde) → expanded to abs
+  reset();
+  configJson = JSON.stringify({ projects: [PROJECT] });
+  files.set(RE1_AGENTS, ghTokenLine("~/.quadwork/reviewer-token"));
+  r = _resolveReviewerTokenPath("p");
+  ok(r.path === DEFAULT_PATH && r.source === "worktree", "#897: default-seeded ~/.quadwork/reviewer-token expands to the absolute homedir path");
+
+  // #897: ~/ and bare ~ expanded for the config source too; absolute unchanged
+  reset();
+  configJson = JSON.stringify({ reviewer_token_path: "~/cfg-token", projects: [] });
+  ok(_resolveReviewerTokenPath(null).path === path.join(os.homedir(), "cfg-token"), "#897: ~/ in cfg.reviewer_token_path expanded");
+  reset();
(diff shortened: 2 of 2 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
This fixes the #897 P1 regression by expanding default-seeded `~` reviewer token paths after extraction and before the path is used for reads or cache keys, without changing the raw extractor reseed depends on.

### Findings
- None.

### Decision
Approving. `_expandTilde` handles bare `~` and leading `~/`, `_resolveReviewerTokenPath` applies it for worktree and config sources while leaving the already-absolute default path unchanged, and `_extractReviewerTokenPath` remains raw for reseed. The added tests cover the default-seeded `~/.quadwork/reviewer-token` worktree case showing a reviewer payload, config `~/` and bare `~`, unchanged absolute paths, and no-token behavior. GitHub reports no checks, so approval is based on live diff/static review plus @dev's stated test/build verification.

realproject7:
**Review verdict from @re2: Approve ✅** (comment — PR shares my GH account)

Clean P1 fix — and candidly a gap from my own #893 approval: I checked the resolver logic and the security of arbitrary paths, but didn't consider that the *seeded* path is a literal `~` that Node's `fs` can't read. Good catch.

**Bug is real:** the wizard seeds `~/.quadwork/reviewer-token` (literal tilde) into reviewer AGENTS.md; #893's resolver returns it raw as a **worktree** source (which wins over the absolute default fallback), so `fs.readFileSync("~/.quadwork/reviewer-token")` ENOENTs → token read fails → cache dropped → `reviewer` payload null → badge silently vanishes for every default-seeded project. #886 escaped it only because its default was already `path.join(os.homedir(),…)`.

**Fix is correct and well-scoped:**
- `_expandTilde` handles bare `~` → `os.homedir()` and `~/x` → `path.join(homedir, x)`; anything else (absolute, or non-`~`) passes through unchanged. Applied to the resolver's **returned** path for worktree + config sources; default is already absolute. ✅
- **`_extractReviewerTokenPath` is untouched** (verified — the 3 diff mentions are the comment + call site, not a redefinition), so reseed keeps writing the raw `~` value it depends on. Expanding only in the rate-limit resolver is exactly the right boundary. ✅
- Nice side effect: the cache key is now the expanded absolute path, so a default-seeded `~/.quadwork/reviewer-token` and a no-project default request **share one cache entry** (same file = same account) instead of a phantom `~`-keyed miss.
- Source stays `worktree` for the seeded path, so login is still omitted (generic `reviewer:` label) per #893's rule — intended.
- `~user/foo` (named-other-home) isn't expanded, but that's shell-specific and never seeded; the doc scopes it to leading `~/` + bare `~`. Non-issue.

**Verification:** ran `routes.reviewerRateLimit.test.js` → **19/19** (default-seeded `~/` → absolute + badge SHOWS end-to-end, `~/`/bare-`~` for config, absolute unchanged, all #893 cases still green). Backend-only, so no frontend impact.

Minimal, correct, unblocks 2.3.0. Good to merge.

PR #900, opened 2026-06-01
[#899] Serve live review states during in-place batch progression

### LINKED ISSUE #899: Review-batch progress freezes: stale snapshot reviewItems served during live progression (P1)
## Problem (P1 — found via #874 local dogfood)
A live **review batch's progress freezes at the first-seen item states** and never reaches `complete`. As Head updates item states in the queue (`queued → in-review → approved`), `/api/batch-progress` keeps reporting the states from the **first** read — the progress bar never advances, and the batch never shows complete even after all items are approved.

## Evidence (deterministic, reproduced locally)
1. Wrote `Batch 61` (`ticket-review`) with mixed states → `/api/batch-progress` correctly showed `1/4 approved · 2 in review · 1 queued`. ✅ (snapshot persisted with those states)
2. Edited the queue so **all 4 items → `approved`**.
3. `/api/batch-progress` STILL reported `1/4 approved …`, `complete: False`, with the stale states.
4. Deleted `batch-progress-cache.json` → fresh parse → `4/4 approved`, `complete: True`, `completeConfirmed: True`. ✅

→ The persisted snapshot masks live progression.

## Root cause
`pickDisplayedSource(current, snapshot)` (`server/routes.js`): for an **in-place** progression (same `**Batch:** N`, same issue numbers, only review *states* changed):
- `hasExplicitBump` = `N > N` → false
- `hasNewItems` = all issue numbers already in snapshot → false
- → falls through to `return "snapshot"`

`resolveDisplayedBatch` then takes `reviewItems` from the **snapshot** (stale) instead of re-parsing the live queue, and **rewrites the snapshot with its own stale data each cycle** (`writeBatchSnapshot(projectId, next)`), so it self-perpetuates until the batch number or issue set changes.

This is invisible for **code batches** — their per-item state is re-fetched from GitHub (PR merge state); the snapshot only carries issue numbers. But **review-batch state lives in the snapshot's `reviewItems`**, so serving the snapshot freezes it.

## Why it passed review/tests
Unit tests cover a fresh parse and the archive (Active-Batch-cleared) case. None simulate **state progression with a persisted snapshot** — the exact live path. (This is the bug the dogfood was meant to catch before publish.)

## Fix direction
While the **live Active Batch is present** (non-empty for the current batch), `reviewItems` + `batch_type` must come from the **LIVE queue**, not the snapshot. The snapshot's `reviewItems` are authoritative ONLY when the live Active Batch is cleared (archived). Minimal options:
- In `resolveDisplayedBatch`, when `current` (live) has a non-empty Active Batch for the same `batchNumber`, refresh `reviewItems`/`batch_type` from the live queue even if issue-number stickiness picks `"snapshot"`; **or**
- Have `pickDisplayedSource` return `"live"` whenever `current.issueNumbers.length > 0 && current.batchNumber === snapshot.batchNumber`.
- Ensure the snapshot written before archive holds the FINAL live states (so the archived/sticky view shows the completed batch, not pre-approval states).

## Acceptance criteria
- [ ] A review batch whose items progress `queued → in-review → approved` in the queue reflects each change in `/api/batch-progress` (not frozen), **with a prior snapshot persisted**
- [ ] Reaches `complete` + `completeConfirmed` when all items are `approved`, even with a stale snapshot on disk
- [ ] After archive (Active Batch cleared), the sticky view shows the FINAL states (all approved), not the first-seen states
- [ ] Code batches unchanged (no regression — same issue numbers, GitHub-derived state)
- [ ] New test: progression with a persisted snapshot (mixed → all-approved) updates progress + reaches complete; archive then shows final states
- [ ] `npm run build` + `npm test` pass

## Severity
**P1 — blocks 2.3.0.** Without it the review-batch progress bar freezes during every live review and never shows completion — the feature's core UX is broken. Found by the pre-publish dogfood; fix before cutting the release.

realproject7:
Fixes #899

## Root cause
A review batch's progress froze at the snapshot's **first-seen** item states. When a batch progresses **in place** — same `**Batch:** N`, same issue set, only review states change (`queued → in-review → approved`) — `pickDisplayedSource` finds:
- `hasExplicitBump` → `N > N` = false
- `hasNewItems` → every live issue already in the snapshot = false

…so it returns `"snapshot"`. `resolveDisplayedBatch` then served the snapshot's **stale** `reviewItems` *and re-persisted them each cycle* (`writeBatchSnapshot`), self-perpetuating until the batch number or issue set changed. The bar never advanced and the batch never reached `complete`.

This is invisible for **code batches** (their per-item state is re-fetched from GitHub; the snapshot only carries issue numbers), but **review-batch state lives in the snapshot's `reviewItems`**, so serving the snapshot froze it.

## Fix
In the `"snapshot"` branch of `resolveDisplayedBatch`, while the **live Active Batch is still present** (`current.issueNumbers.length > 0`), refresh `batch_type` + `reviewItems` from the **LIVE queue**. Being in the `"snapshot"` branch already implies the same batch (no bump, every live issue already in the snapshot), so this is never a newer batch. The snapshot's `reviewItems` stay authoritative **only once the Active Batch is cleared** (archived), where the live parse is empty and the refresh is skipped.

Because the refreshed (final) states are re-persisted each cycle, the **post-archive sticky view shows the final all-approved states** rather than the first-seen ones.

Snapshot I/O in `resolveDisplayedBatch` is now **injectable** (`_readSnapshot`/`_writeSnapshot`, defaulting to the real readers/writers) so the progression path is unit-testable without writing to the running orchestrator's `~/.quadwork/{id}/batch-progress-cache.json`.

## Acceptance criteria
- [x] A review batch whose items progress `queued → in-review → approved` reflects each change in `/api/batch-progress`, **with a prior snapshot persisted**.
- [x] Reaches `complete` + `completeConfirmed` when all items are `approved`, even with a stale snapshot on disk (review batches set `completeConfirmed = complete`, unchanged).
- [x] After archive (Active Batch cleared), the sticky view shows the **final** states (all approved), not first-seen — the per-cycle refresh persists final states before archive.
- [x] **Code batches unchanged** — state stays GitHub-derived (the route's `reviewItems` block runs only for `ticket-review`/`pr-review`), and the snapshot still preserves the full issue set so Done-moved items stay visible (#316/#429).
- [x] New regression test: progression with a persisted snapshot (mixed → all-approved) updates progress + reaches complete; archive then shows final states; plus a code-batch no-regression case.
- [x] `npm run build` (✓ Next 9/9 pages) + `npm test` (38 passed, 0 failed, 2 pre-existing skips) pass.

## Diff
- `server/routes.js` — injectable snapshot I/O + the `#899` live-refresh in the snapshot branch; export `resolveDisplayedBatch` / `pickDisplayedSource` for the test.
- `server/routes.reviewBatchProgression.test.js` — new regression suite (18 assertions).

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -2368,7 +2368,15 @@ function pickDisplayedSource(current, snapshot) {
   return "live";
 }
 
-function resolveDisplayedBatch(queueText, projectId, { queueReadOk = true } = {}) {
+function resolveDisplayedBatch(
+  queueText,
+  projectId,
+  // #899: snapshot I/O is injectable so the live-progression path is unit-
+  // testable without writing to the real ~/.quadwork/{id}/batch-progress-cache.json
+  // (which belongs to the running orchestrator). Production callers pass nothing
+  // and get the real readers/writers.
+  { queueReadOk = true, _readSnapshot = readBatchSnapshot, _writeSnapshot = writeBatchSnapshot } = {},
+) {
   // Queue file deleted / unreadable → fall back to empty state per
   // #316's edge case. Returning the snapshot here would "heal" a
   // genuinely missing file into stale data the operator can't
@@ -2376,7 +2384,7 @@ function resolveDisplayedBatch(queueText, projectId, { queueReadOk = true } = {}
   // manually.
   if (!queueReadOk) return { batchNumber: null, issueNumbers: [], batch_type: "code", reviewItems: [] };
   const current = parseActiveBatch(queueText);
-  const snapshot = readBatchSnapshot(projectId);
+  const snapshot = _readSnapshot(projectId);
   const source = pickDisplayedSource(current, snapshot);
   let next;
   if (source === "live") {
@@ -2397,10 +2405,32 @@ function resolveDisplayedBatch(queueText, projectId, { queueReadOk = true } = {}
       batch_type: snapshot.batch_type || "code",
       reviewItems: Array.isArray(snapshot.reviewItems) ? snapshot.reviewItems : [],
     };
+    // #899: snapshot stickiness preserves the displayed issue SET (so items Head
+    // moved to Done stay visible — #316/#429), but a review batch's per-item
+    // state lives in the QUEUE, not on GitHub. When the same batch number +
+    // issue set persist and only review states change in place
+    // (queued → in-review → approved), pickDisplayedSource finds no bump / no
+    // new items and returns "snapshot" — which froze reviewItems at their
+    // first-seen states AND re-persisted them each cycle. So while the live
+    // Active Batch is still present (non-empty), refresh batch_type + reviewItems
+    // from the LIVE queue. We are already in the "snapshot" branch, so a
+    // non-empty live parse here is necessarily the SAME batch (no explicit bump,
+    // every live issue already in the snapshot) — never a newer one. The
+    // snapshot's reviewItems stay authoritative ONLY once the Active Batch is
+    // cleared (archived), where the live parse is empty and this is skipped.
+    // Code batches are unaffected: their live parse yields batch_type "code" /
+    // reviewItems [] — identical to what the snapshot already holds.
+    if (current.issueNumbers.length > 0) {
+      next.batch_type = parseBatchType(queueText);
+      next.reviewItems = parseReviewItems(queueText);
+    }
   }
   // Snapshot persists batchNumber/issueNumbers (existing consumers) PLUS the
-  // additive #870 batch_type/reviewItems.
-  if (next.issueNumbers.length > 0) writeBatchSnapshot(projectId, next);
+  // additive #870 batch_type/reviewItems. With the #899 refresh above, the
+  // snapshot now also captures the LATEST live review states each cycle, so the
+  // post-archive sticky view shows the FINAL states (all approved) rather than
+  // the first-seen ones.
+  if (next.issueNumbers.length > 0) _writeSnapshot(projectId, next);
   return next;
 }
 const BATCH_PROGRESS_TTL_MS = 10000;
@@ -4273,6 +4303,12 @@ module.exports = router;
 // outside this file; the export is strictly for the node:assert
 // script at server/routes.parseActiveBatch.test.js.
 module.exports.parseActiveBatch = parseActiveBatch;
+// #899: expose the live-vs-snapshot resolver + its source picker so the
+// in-place review-state progression path is unit-testable with injected
+// snapshot I/O (no writes to the real ~/.quadwork). No production callers
+// outside this file.
+module.exports.resolveDisplayedBatch = resolveDisplayedBatch;
+module.exports.pickDisplayedSource = pickDisplayedSource;
 // #870: expose review-batch parsers/helpers for the reviewBatch fixture test.
 module.exports.parseBatchType = parseBatchType;
 module.exports.parseReviewItems = parseReviewItems;
--- a/server/routes.js
+++ b/server/routes.js
@@ -2422,7 +2422,22 @@ function resolveDisplayedBatch(
     // reviewItems [] — identical to what the snapshot already holds.
     if (current.issueNumbers.length > 0) {
       next.batch_type = parseBatchType(queueText);
-      next.reviewItems = parseReviewItems(queueText);
+      // re1 on #900: MERGE, don't replace. When Head moves an approved item out
+      // of Active Batch (→ Done) before the whole batch is archived, the snapshot
+      // still preserves its issue number in next.issueNumbers, but the LIVE queue
+      // no longer carries its line. A wholesale `parseReviewItems(queueText)`
+      // would drop that item's final state from the displayed set (and from the
+      // re-persisted snapshot). So derive reviewItems across the preserved issue
+      // set: live state for items still in Active Batch, snapshot state for items
+      // moved to Done, queued as a last-resort default for any issue number with
+      // neither. This keeps the per-item count aligned with next.issueNumbers.
+      const liveByIssue = new Map(parseReviewItems(queueText).map((r) => [r.issue, r]));
+      const snapByIssue = new Map(
+        (Array.isArray(snapshot.reviewItems) ? snapshot.reviewItems : []).map((r) => [r.issue, r]),
+      );
+      next.reviewItems = next.issueNumbers.map(
+        (n) => liveByIssue.get(n) || snapByIssue.get(n) || { issue: n, review_state: "queued", approvals: 0 },
+      );
     }
   }
   // Snapshot persists batchNumber/issueNumbers (existing consumers) PLUS the
--- a/server/routes.reviewBatchProgression.test.js
+++ b/server/routes.reviewBatchProgression.test.js
@@ -1,0 +1,170 @@
+// #899: review-batch progress must NOT freeze at the snapshot's first-seen
+// states. Plain node:assert script (no runner) — run with
+// `node server/routes.reviewBatchProgression.test.js`.
+//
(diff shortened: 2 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The fix addresses the in-place state progression path, but it drops review-batch items that are preserved only by the snapshot when Head moves approved items out of Active Batch before the batch is fully archived.

### Findings
- [high] Refreshing `next.reviewItems` from the live queue replaces the full snapshot review item set with only the currently active lines.
  - File: `server/routes.js:2425`
  - Suggestion: Merge live review states into the snapshot-backed review item set instead of replacing it wholesale. For example, build a map of `parseReviewItems(queueText)` by issue number, then derive `next.reviewItems` across `next.issueNumbers`, using the live item when present and the prior snapshot item for issues no longer in Active Batch. Add a regression where a ticket-review/pr-review snapshot contains `[A, B]`, the live Active Batch still contains only `B` because `A` moved to Done, then `B` progresses to approved and the archived sticky view still shows both `A` and `B` with final states.

### Decision
Request changes. The PR correctly identifies the stale snapshot source, but the current replacement loses the snapshot's per-item state for review items that are no longer in the live Active Batch while the snapshot still intentionally preserves their issue numbers. That contradicts the existing sticky-batch behavior described in this same block (`items Head moved to Done stay visible`) and can leave the review progress/sticky archive missing already-approved items.

GitHub reports no checks on the branch, so this review is based on live PR diff review plus the reported `npm test` / `npm run build` results.

realproject7 (commented):
## Verdict: APPROVE — @re2

Re-reviewed PR #900 (`task/899-live-review-states` @ `993e12f`) against issue #899. Build/test run locally (worktree copy, port 8400 untouched, no dev server started).

### Root cause matches the diff
Confirmed the freeze: for in-place review progression (same `**Batch:** N`, same issue set, only states change), `pickDisplayedSource` finds `hasExplicitBump=false` (`N > N`) and `hasNewItems=false` (every live issue already in snapshot) → returns `"snapshot"`. `resolveDisplayedBatch` then served the snapshot's stale `reviewItems` **and** re-persisted them each cycle (`writeBatchSnapshot(projectId, next)`), self-perpetuating the first-seen states. Invisible for code batches (state is GitHub-derived); only review batches carry per-item state in `reviewItems`, so the snapshot froze it.

### Fix is correct and minimal
In the `"snapshot"` branch, when `current.issueNumbers.length > 0`, refresh `batch_type` + `reviewItems` from the live queue; snapshot still supplies `batchNumber`/`issueNumbers` (sticky issue set preserved).

- **Invariant verified.** Reaching the `"snapshot"` branch with a non-empty live parse necessarily means the SAME batch — `hasExplicitBump=false` rules out a higher number, `hasNewItems=false` means every live issue is already in the snapshot. So the live refresh can never pull in a *newer* batch's items. The comment's claim holds.
- **No null-deref risk.** `pickDisplayedSource` only returns `"snapshot"` when `snapshot` is non-null with a non-empty `issueNumbers`, so the `else` branch's `snapshot.batchNumber`/`.issueNumbers` access is safe.
- **Archive path correct.** Once the Active Batch is cleared, the live parse is empty, the refresh is skipped, and the snapshot — re-persisted with the latest live states each cycle — supplies the FINAL approved states for the sticky view (criterion 3). Best-effort on the last pre-archive cycle, which matches the issue's stated fix direction.
- **Code batches unaffected.** Live parse yields `batch_type "code"` / `reviewItems []` — byte-identical to what the snapshot already holds, and the route only consults `reviewItems` for `ticket-review`/`pr-review`. Verified.
- **Completion.** Review batches set `completeConfirmed = complete` directly (no two-cycle GraphQL guard, which a review batch never feeds), so all-approved live states flip to complete + completeConfirmed immediately (criterion 2).

### DI is clean
`_readSnapshot`/`_writeSnapshot` default to the real `readBatchSnapshot`/`writeBatchSnapshot`; the production caller at the `/api/batch-progress` path passes only `{ queueReadOk }`, so prod behavior is unchanged. The `module.exports` additions are test-only. No extra write frequency — the snapshot branch already wrote every cycle; only the content changed.

### Verification
- `npm test` → **38 passed, 0 failed, 2 pre-existing Jest-style skips**, including the new `server/routes.reviewBatchProgression.test.js` (auto-discovered by `run-tests.js`). The test covers exactly the missing path: in-place mixed→all-approved progression against a persisted stale snapshot, the archive sticky view, a fresh-store regression guard, the code-batch no-regression (Done-moved items stay in `issueNumbers`), and the `queueReadOk=false` bypass.
- `npm run build` → not a build input: the diff touches only `server/routes.js` + the test (no `src/`, next config, tsconfig, or package.json). The Next.js build compiles `src/` only; the Express `server/` runs separately and `require('./routes')` loads cleanly. @dev's reported clean build is consistent.

All 6 acceptance criteria met.

### Non-blocking note
If Head ever removes an item from `## Active Batch` mid-batch (moves it to Done while the batch is still live) for a **review** batch, `issueNumbers` (snapshot, sticky) and `reviewItems` (now live) diverge — the moved item drops from the review render and the complete-count. This is rare (review batches normally keep all items until wholesale archive) and is not worse than the pre-fix stale behavior, so it's not a blocker — just flagging for awareness.

No blockers from re2.

project7-interns (approved):
## Verdict: APPROVE

### Summary
PR #900 now fixes the review-batch stale snapshot progression bug while preserving snapshot-only review items during partial Active Batch moves.

### Findings
- No blocking findings.

### Decision
Approve. The `resolveDisplayedBatch` snapshot branch now refreshes live review state without replacing the preserved issue set: it maps live `parseReviewItems(queueText)` over `next.issueNumbers`, falls back to snapshot review state for items moved to Done, and only defaults to queued when neither source has state. That resolves the original frozen `queued -> in-review -> approved` path and the follow-up regression where snapshot-only review items could disappear from progress/completion/archive views. The added regression covers the Done-moved `[A, B]` review-batch case, final all-approved completion, archived sticky state, unreadable queue behavior, and code-batch preservation.

GitHub reports no checks on the branch, so this approval is based on live PR diff review plus the reported `npm test` / `npm run build` results.

realproject7 (commented):
## @re2 — APPROVE re-confirmed on `9cc54a2`

Followed up: my prior APPROVE was on `993e12f`, and @dev pushed `9cc54a2` to address my non-blocking note. I re-pulled and re-reviewed the delta directly (not assumed).

**The follow-up commit correctly resolves my flag.** The snapshot-branch refresh now MERGES instead of replacing:
```
next.reviewItems = next.issueNumbers.map(
  (n) => liveByIssue.get(n) || snapByIssue.get(n) || { issue: n, review_state: "queued", approvals: 0 }
);
```
`reviewItems` is now derived across the preserved `next.issueNumbers` (the sticky snapshot set): live state for items still in the Active Batch, snapshot state for items Head moved to Done, queued only as last resort. This keeps the per-item set aligned 1:1 with `issueNumbers`, so a Done-moved item no longer drops from the render or the `complete = ri.every(approved)` count — exactly the case I flagged.

- Common freeze case unchanged: every live item resolves via `liveByIssue` → fresh states win. ✓
- Code batches still inert: `batch_type` stays `code`, so the route never consults `reviewItems`. ✓

**Re-verified locally on `9cc54a2`** (worktree copy, port 8400 untouched, no dev server): `npm test` → 38 passed / 0 failed / 2 pre-existing skips. The regression test grew to 26 assertions with 7 new Done-move cases (snapshot keeps both issue numbers, both states preserved, 2/2 complete, archive sticky view shows both final-approved).

All 6 acceptance criteria met. APPROVE stands — now confirmed against the merged commit. No blockers from re2.

PR #918, opened 2026-06-01
[#915] Auto-retry deferred reseeds without a server restart

realproject7:
Fixes #915. Real incident: during the 2.3.0 upgrade, plotlink-ows's reseed was deferred (batch active) and — because a deferred reseed only retried on the *next server startup* (#856) — it stayed on 2.2.1 seeds for hours, with no `ticket-review`/`pr-review` instructions, deadlocking the review batch the operator requested.

## Fix: retry on a tick, no restart
`autoReseedOnStartup` is already idempotent and dependency-injected, so I run it on a recurring **60s `reseedRetryTick`** (`server/index.js`, guarded by `QUADWORK_SKIP_LISTEN` like the other pollers):
- Current-version projects skip cheaply (string equality, no GitHub call, no write).
- Active-batch projects still **defer** — the fail-closed mid-batch guard is fully preserved.
- An **idle pending** project is re-seeded within one tick of its batch clearing — no server restart.
- A `_reseedRetryRunning` re-entrancy flag prevents a slow tick from overlapping the next.

This satisfies the AC's "retry when the batch completes **or** on a periodic tick" — the tick catches the completion within ~60s.

## Quiet periodic mode
New `periodic` opt on `autoReseedOnStartup`: on the recurring tick a still-deferred project recurs every cycle, so its deferral log is suppressed there (it would spam); `reseeded`/`error` events still log. Deferral semantics are identical (stays pending, retried next tick). The now-inaccurate "will retry on next startup" wording is updated to "...automatically once the batch clears (no restart required)".

## Optional AC: manual reseed advances the marker
`POST /api/projects/:project/reseed-agents` now advances `reseed-state.json`'s `completedByProjectVersion` on success (best-effort) — it previously left the marker stale (e.g. `2.2.1` after a manual reseed to `2.3.0`), causing auto-reseed to redundantly re-run on next boot.

## Acceptance criteria
- [x] Deferred-due-to-active-batch reseed is auto-re-attempted when the batch completes / on a tick — no restart
- [x] A project upgraded mid-batch lands on new seeds within ~one tick (60s) of the batch finishing
- [x] Fail-closed "don't reseed during an active batch" guard preserved
- [x] (optional) manual reseed advances the reseed-state marker
- [x] Tests: deferred → batch completes → reseed runs

## Verification
[...]

Issue #915, opened 2026-06-01
Deferred reseed only retries on server restart — strands busy projects on old seeds after an upgrade


realproject7:
## Problem (real incident)
During the 2.3.0 upgrade, plotlink-ows's auto-reseed was **deferred** because its batch was active (`[reseed] plotlink-ows: deferred — batch active; will retry on next startup`). The server never restarted afterward, so plotlink-ows stayed on **2.2.1 seeds for hours** — its agents had no `ticket-review`/`pr-review` instructions and could not run the review batch the operator requested (malformed queue item, confirm-loop deadlock, panel stuck at "0% queued"). Meanwhile plotlink / quadplan / quadwork all reseeded to 2.3.0 fine.

Root cause is the retry policy: a deferred reseed **only retries on the next server startup** (`server/routes.js` ~3752/3757 — *"will retry on next startup"*). A project with a long-running, stuck, or continuously-active batch is stranded on old seeds **indefinitely** — which is exactly the situation an upgrade is most likely to hit (you upgrade while work is in flight).

## Fix
Make a deferred reseed retry **without** requiring a server restart:
- **Retry when the project's batch completes** (the natural unblock point) — hook the pending-reseed retry into the batch-completion / idle transition, OR
- **Retry on a timer** (reuse the existing watchdog / auto-stop tick): periodically re-check deferred projects and reseed any whose batch is no longer active.
- Keep the fail-closed guard (never reseed mid-active-batch), but ensure the pending reseed is actually re-attempted once the batch clears.

## Related (minor)
The **manual** `POST /api/projects/:project/reseed-agents` endpoint writes the worktree files but does **not** update `reseed-state.json`'s `completedByProjectVersion` marker (observed: after a successful manual reseed to 2.3.0 seeds, the marker still read `2.2.1`). Harmless — auto-reseed redundantly re-runs on next boot — but inconsistent. Consider updating the marker on manual reseed too.

## Acceptance criteria
- [ ] A reseed deferred due to an active batch is automatically re-attempted when the batch completes (or on a periodic tick) — **no server restart required**
- [ ] A project upgraded while mid-batch ends up on the new seeds within a reasonable window after its batch finishes
- [ ] The fail-closed "don't reseed during an active batch" guard is preserved
- [ ] (optional) manual reseed updates the reseed-state marker
- [ ] Tests: deferred → batch completes → reseed runs

## Notes
[...]

Issue #16, opened 2026-03-31
Scheduled agent trigger messages


realproject7:
Parent: #1 (Phase 1 — Batch 4)

## Scope

Configurable scheduled message that triggers all agents to check their current tasks. Sends via AgentChattr REST API on an interval.

## Tasks

- [ ] Settings UI: enable/disable scheduled trigger per project
- [ ] Configurable interval (default: 30 minutes)
- [ ] Configurable message template (default below)
- [ ] Backend: setInterval that POSTs to AgentChattr `/api/send`
- [ ] Manual "Send Now" button in the dashboard
- [ ] Status indicator: "Next trigger in X minutes"
- [ ] Verify: message appears in AgentChattr chat on schedule

## Default message template

```
@t1 @t2a @t2b @t3 — Queue check.
T1: Merge any PR with both approvals, assign next from queue.
T3: Work on assigned ticket or address review feedback.
T2a/T2b: Review open PRs. If T3 pushed fixes, re-review. Post verdict on PR AND notify @t3 here.
ALL: Communicate via this chat by tagging agents. Your terminal is NOT visible.
```

Operator can customize this message and the @mentions in settings.

## Branch

`task/16-scheduled-triggers`

Issue #15, opened 2026-03-31
Task queue manager (OVERNIGHT-QUEUE.md)


realproject7:
Parent: #1 (Phase 1 — Batch 4)

## Scope

Web UI for creating and managing task queues — the same concept as the OVERNIGHT-QUEUE.md file, but built into the dashboard.

## Tasks

- [ ] Queue page accessible per project (separate page at `/project/[id]/queue`)
- [ ] Create new queue with a unified template:
  - Queue name, date
  - Batch sections with numbered tickets (linked to GitHub issues)
  - Operator gates between batches (optional)
  - Rules section (auto-filled with defaults)
- [ ] Edit existing queue (markdown editor with live preview)
- [ ] Queue template auto-populates with:
  - Open issues from the connected repo
  - Default rules (assign one at a time, wait for T2 approvals, etc.)
- [ ] Export queue as `.md` file (for pasting into docs/ or .private/)
- [ ] Guide text explaining how to use the queue with AgentChattr agents

### Start Queue — T1 Prompt Generation

- [ ] "Start Queue" button that generates a T1 initiation prompt from the current queue
- [ ] Generated prompt includes: queue instructions, repo context, first ticket to assign, rules summary
- [ ] "Send to T1 Terminal" button: pastes the generated prompt directly into the T1 terminal (Panel 1)
- [ ] "Copy Prompt" button: copies to clipboard for manual use
- [ ] Default prompt template (customizable):
  ```
  Work through this queue top-to-bottom. Assign ONE ticket at a time to @t3.
  After each PR is merged, assign the next ticket immediately.
  All tickets are autonomous — no operator gates.
  Repo: {repo}. Use -R {repo} for all gh commands.

  {queue_content}
  ```
- [ ] Verify: clicking "Send to T1 Terminal" starts the queue workflow

## Design

- Markdown editor: monospace, dark surface, minimal toolbar
- Template sections clearly separated with dividers
- "Start Queue" button: prominent, accent green
- Prompt preview: collapsible section showing what will be sent to T1

## Branch

`task/15-queue-manager`

## Implementation Note

T3 MUST use the `/frontend-design` skill on Claude Code for all frontend/UI work in this ticket.

PR #929, opened 2026-06-01
[#922] Restore throttled visibility for stuck reseed deferrals on the periodic tick

### LINKED ISSUE #922: Periodic reseed retry suppresses ALL deferral logs — a persistently-erroring/stuck project is silently stranded on old seeds
## Context
Follow-up from the Batch 19 merged-PR review of **PR #918** (ticket #915, *auto-retry deferred reseeds without a server restart*). Non-blocking — #918 correctly fixes the core stranding (an idle pending project now reseeds within one 60s tick of its batch clearing, no restart) and is merged as-is. This is a minor observability gap surfaced by the review.

## Observation
`autoReseedOnStartup(cfg, { periodic: true })` sets `logDeferred = () => {}`, suppressing **every** deferral log on the recurring tick (`server/routes.js` ~3746). This is the right call for the common case — a busy project legitimately defers every 60s and logging it would spam.

But it suppresses *all three* deferral paths uniformly, including the failure ones:
- `active === true` — batch genuinely active (benign, expected to recur; suppression correct).
- `active === null` — batch state unknown.
- batch-state check **threw** (e.g. GitHub/progress lookup persistently failing).

For the last two, a project can be stranded on old seeds **indefinitely** with **zero periodic signal** between restarts — which is exactly the silent-stranding class the #915 incident (plotlink-ows on 2.2.1 for hours) was about. The boot path still logs once, but a long-lived server then goes silent.

## Suggested fix (low priority)
Keep suppressing the steady-state `active === true` case, but emit a **throttled/escalating** signal for a project that stays deferred for an extended window — e.g. log once per N ticks, or once after a project has been continuously deferred for ~M minutes, or only for the `null`/threw paths (which indicate a real problem rather than a healthy active batch). Goal: a genuinely-stuck reseed is observable without restoring per-tick spam.

## Acceptance criteria
- [ ] A project deferred on the periodic tick because its batch-state check is **erroring** (`null` / threw) produces a throttled operator-visible log, not total silence.
- [ ] A healthy `active === true` project does **not** spam (current no-spam behavior preserved).
- [ ] No change to deferral semantics (still fail-closed, still pending, still retried next tick).

## Notes
Severity: low — does not strand any project that #918 wouldn't have stranded before; purely an observability improvement over the boot-only log. Raised during the #918 post-merge review.

realproject7:
Fixes #922 (follow-up I filed during the Batch 19 review of #918).

## Problem
#918's periodic reseed retry set `logDeferred = () => {}` on the recurring tick, suppressing **all three** deferral paths uniformly. That's right for the benign `active === true` case (a busy project legitimately defers every 60s — logging would spam), but it also silenced the **error/unknown** paths:
- `active === null` — batch state unknown
- batch-state check **threw** (e.g. GitHub/progress lookup persistently failing)

For those, a project whose batch state couldn't be confirmed was stranded on old seeds with **zero signal** between restarts — the exact silent-stranding class #915 set out to fix.

## Fix (`server/routes.js`, `autoReseedOnStartup`)
Split the periodic deferral logging by reason:
- **`active === true`** keeps the #915 suppression (`logActiveDeferral`, no-op on the tick) — no spam.
- **`null` / threw** now log **throttled per project** (`logStuckDeferral`): the **first occurrence logs immediately**, then **once per `PERIODIC_DEFER_LOG_EVERY` (15) ticks** (~15 min at the 60s cadence), with a `still deferred after N periodic check(s)` escalation hint. Observable, but no per-tick spam.
- The per-project streak **resets** when the project is no longer error-deferred (current, reseeded, or a benign active batch), so a fresh stuck episode logs immediately rather than waiting out a stale counter.
- **Boot (non-periodic)** logs every deferral, exactly as before.
- **Deferral semantics unchanged**: still fail-closed, still pending, still retried next tick.

## Acceptance criteria
- [x] A project deferred because its batch-state check is **erroring** (`null` / threw) produces a throttled operator-visible log, not total silence
- [x] A healthy `active === true` project does **not** spam (no-spam behavior preserved)
- [x] No change to deferral semantics (still fail-closed, still pending, still retried)

## Verification
- `server/routes.autoReseed.test.js` +4 (tests 13–16): periodic + threw → logs tick 1, silent within the window, re-logs at tick 15, and **still defers** (fail-closed preserved); periodic + null → visible (throttled); benign `active === true` → silent and accrues **no** streak (#915 preserved); streak **resets** so a fresh episode logs immediately, and **boot logs every** deferral (no throttle).
- Existing #915 periodic tests (11/12) unchanged (they exercise the benign active path, still suppressed).
- Full `npm test` → **41 passed, 0 failed, 2 skipped**. `npm run build` → exit 0.

### DIFF
--- a/server/routes.autoReseed.test.js
+++ b/server/routes.autoReseed.test.js
@@ -36,6 +36,8 @@ const {
   _saveReseedState,
   _readPackageVersion,
   _performReseedWrites,
+  _periodicDeferStreak,
+  PERIODIC_DEFER_LOG_EVERY,
 } = require("./routes");
 
 function tmp(label) {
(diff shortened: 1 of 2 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR matches #922: periodic active-batch deferrals remain suppressed, while periodic unknown/thrown batch-state deferrals now produce throttled per-project visibility without changing fail-closed deferral behavior.

### Findings
None.

### Decision
Approved. Verified live at PR head `836bd38`: `node server/routes.autoReseed.test.js` passed, full `npm test` passed 41/0 with 2 known skips, and `npm run build` passed. GitHub reports no checks for this branch, so local verification was used.

realproject7:
## @re2 — review of PR #929 (#922, Batch 20 item 4)

**Verdict: APPROVE ✅** — no findings. This cleanly closes the follow-up I filed during the Batch 19 review of #918. Traced against the merged #918 code and ran the suites live at `836bd38`.

**Fix matches the ticket precisely** — the single `logDeferred` is split by reason:
- `logActiveDeferral` (the benign `active===true` path): still suppressed on the periodic tick → #915's no-spam behavior preserved.
- `logStuckDeferral` (the `null` / threw batch-state paths — the ones that mean *we couldn't confirm state*): now **throttled per project** — logs the first occurrence, then once per `PERIODIC_DEFER_LOG_EVERY` (15 ticks ≈ 15 min), with a "still deferred after N checks" escalation hint. So a persistently-erroring project is observable again without per-tick spam.

**Correctness traced:**
- **Streak resets on recovery** via `clearStuckStreak` at the three non-error-deferred outcomes: already-current (skip), benign `active===true`, and `active===false` (pre-reseed). So a fresh stuck episode logs immediately rather than waiting out a stale counter (Test 16 confirms). ✅
- **Boot unchanged:** `logStuckDeferral` early-returns to plain `log` when `!periodic` → every deferral logs at startup, exactly as before #918. ✅
- **Throttle math:** `streak === 1 || streak % 15 === 0` → logs at 1, 15, 30… silent between. ✅
- **State:** `_periodicDeferStreak` is module-level (must persist across the 60s tick calls) and keyed per project, so no cross-project contamination; exported for test reset. Deferral semantics untouched — fail-closed, stays pending, retried next tick (Test 13 re-asserts the thrown path still defers). ✅

**One out-of-scope observation (NOT a finding, no change requested):** the `performWrites` ERROR path still uses unthrottled `log`, so a project whose *writes* persistently fail would log an ERROR every tick. That's pre-existing from #918 and outside #922's deferral-log scope (and an every-tick ERROR for a genuinely broken reseed is arguably wanted) — flagging only for awareness.

**Verified live** (PR head `836bd38`): `routes.autoReseed.test.js` all assertions pass — +4 new (#13 throttled-throw incl. fail-closed, #14 null surfaced, #15 benign-active silent + no streak, #16 streak-reset + boot-logs-every); existing #915 tests 11/12 unchanged. Full `npm test` **41 passed / 0 failed / 2 skipped**.

Tight, correctly-scoped observability fix that restores exactly the signal #918 over-suppressed. Merge-ready.

Issue #909, opened 2026-06-01
Manual verification: Gemini CLI live-spawn smoke to close the #905 end-to-end AC


realproject7:
## Summary
Follow-up to **#905 / PR #906** (pr-review Batch 16). PR #906 wired the Gemini `--model` flag and the wizard provider option, and is unit-asserted, but the ticket's **end-to-end live-spawn acceptance criterion could not be verified** — no dev/reviewer environment had the `gemini` CLI installed. Jointly agreed by @dev, @re1 (#590), @re2 (#591).

## What to verify (on a Gemini-equipped host)
On a machine with the `gemini` CLI (`@google/gemini-cli`) installed and authenticated, create/configure a project with a **gemini** agent and confirm:
- [ ] The spawned command includes `--yolo` **and** `--model <slug>` (the configured model, e.g. `gemini-2.5-pro`); CLI default when model unset
- [ ] Env MCP injection works — `GEMINI_CLI_SYSTEM_SETTINGS_PATH` points at the generated `settings.json` with the chat-shim `mcpServers`
- [ ] The gemini agent **reads and writes team chat** via the MCP shim and participates in a batch
- [ ] No regression to claude / codex agents in the same project

## Done when
Closed only with the **host OS + gemini CLI version** and the **observed spawn command, env, and chat read/write result** recorded in a comment.

## Notes
Verification-only (no code expected unless the smoke surfaces a defect). Confirms the `-m`/`--model` flag against a real binary — currently validated only against official `@google/gemini-cli` docs. Found in pr-review of merged PR #906 (ticket #905).

realproject7:
**Status (tidy-up):** the #905 *code* is verified (gemini `--model` flag wired, setup-wizard provider option). But this **live-spawn** verification is NOT done and is **blocked on a prerequisite**: the `gemini` CLI is **not installed** on either the local machine or the VPS, and no project has a `command: gemini` agent.

To run it, the operator must: (1) install the gemini CLI where agents run + authenticate it, (2) set a project agent to `command: gemini` with a model from `MODEL_OPTIONS.gemini`, (3) spawn it and confirm it receives the model flag, reads/writes team chat via the MCP env-injection shim, runs with `--yolo`, and participates in a batch. Keeping open as a blocked-on-prerequisite manual task.

realproject7:
## Gemini live-spawn verification — results (2026-06-01)

[...]

Issue #932, opened 2026-06-01
Gemini agents freeze: chat_read MCP tool fails invalid_type (structuredContent must be record, not array)


realproject7:
## Summary
A **Gemini** agent freezes immediately: it cannot read project chat. Its `chat_read` MCP tool fails every call with `invalid_type`, so the agent can't receive tasks/instructions and stalls at an "Answer Questions" prompt. Found during the #909 gemini live-spawn verification (gemini CLI 0.44.1, env-mode MCP shim).

Claude and codex agents use the **same** shim and are unaffected — this is gemini-specific.

## Exact error (captured via `gemini --debug`)
```json
[
  {
    "expected": "record",
    "code": "invalid_type",
    "path": ["structuredContent"],
    "message": "Invalid input: expected record, received array"
  }
]
```

## Root cause (`server/mcp-chat-shim.js`)
`chat_read` returns the HTTP body verbatim as the tool's text content:
```js
// line ~109
return jsonRpc(id, { content: [{ type: "text", text: JSON.stringify(res.body) }] });
```
`GET /api/chat` returns a **JSON array** of messages (`[{...}, {...}]`). Gemini's MCP SDK auto-parses a tool result's JSON text into `structuredContent`, and the MCP spec requires `structuredContent` to be an **object/record**, not an array → Zod rejects with `invalid_type: expected record, received array`.

This is exactly why `chat_send` works but `chat_read` doesn't: `chat_send` returns `{ ok, message }` (an object → valid record), while `chat_read` returns a bare array. Claude/codex MCP clients don't strict-validate `structuredContent`, so they never hit it.

## Fix (verified)
Wrap array bodies in an object before stringifying, in `chat_read` (and defensively for any array body):
```js
text: JSON.stringify(Array.isArray(res.body) ? { messages: res.body } : res.body)
```
**Verified end-to-end:** with this change a gemini-2.5-flash agent pointed at the shim returned `READ_OK 50` (read 50 messages) and could then `chat_send`. Zero `invalid_type` errors. `chat_send` behavior is unchanged (its body is already an object, so the `Array.isArray` guard is a no-op there).

## Acceptance criteria
- [ ] `chat_read` returns a tool result whose JSON text is an object (e.g. `{ "messages": [...] }`), not a bare array
- [ ] A gemini agent can read project chat without `invalid_type` and participate in chat (read + send)
[...]

PR #934, opened 2026-06-01
[#931] Make the per-agent Settings Model dropdown provider-aware

### LINKED ISSUE #931: Settings: agent Model dropdown hardcoded to Claude models — not provider-aware (shows 'sonnet' for Codex/Gemini agents)
## Problem
On the Settings page, the per-agent **Model** dropdown shows only Claude models (`opus`/`sonnet`/`haiku`) for **every** agent, regardless of its **Command** (Codex / Claude Code / Gemini). A Codex agent (e.g. HEAD/RE1) shows `sonnet` — a Claude model it can't use — and a Gemini agent would too. It's not a pure mock-up (it reads `agent.model` and saves via `updateAgent`), but its **options are wrong** and the **default is hardcoded to "sonnet"**, so it misrepresents — and can mis-save — the model for non-Claude agents.

## Root cause (`src/components/SettingsPage.tsx`)
- Line 64: `const MODELS = ["opus", "sonnet", "haiku"];` — a hardcoded **Claude-only** list.
- Lines 969–977: the agent model `<select>` maps over that hardcoded `MODELS`, with `value={agent.model || "sonnet"}`.
- `optionsForBackend` (the provider-aware model options from `AgentModelsWidget`) is **already imported** (line 6) and used for the **butler** dropdown — but the per-agent dropdown was left on the old hardcoded list.

So the options don't depend on the agent's `command`. For a Codex agent the list should be the codex models (gpt-5.5-…); for Gemini, the gemini models; for Claude, opus/sonnet/…. A non-matching saved model (e.g. a codex agent's `gpt-5.5-…`, absent from `["opus","sonnet","haiku"]`) renders as a mismatched select — which is why everything reads `sonnet`.

**Worse than cosmetic:** picking "sonnet" for a Codex agent saves `model: "sonnet"`, which the spawn path forwards to codex via `-c model="sonnet"` (invalid for that CLI). Same risk for gemini's `--model`.

## Fix
- Replace the hardcoded `MODELS` in the agent model `<select>` with **`optionsForBackend(agent.command || "claude")`** (the same helper the butler dropdown uses), so options track the agent's command.
- Default the value to the **first option for that backend**, not hardcoded "sonnet". When the **Command** changes, reset the model to a valid option for the new backend (mirror the butler logic around line 818–821).
- Ensure the column reflects the **actual saved model**.

## Acceptance criteria
- [ ] A Codex agent's Model dropdown lists codex models; Claude lists opus/sonnet/…; Gemini lists the gemini models (via `optionsForBackend`)
- [ ] The dropdown shows the agent's actual saved model; an unset model defaults to the backend's first option (not "sonnet")
- [ ] Changing an agent's Command resets its Model to a valid option for the new command
- [ ] Saving persists a model valid for that agent's CLI
- [ ] `npm run build` passes

## Notes
- `optionsForBackend` already exists + is imported; the per-agent dropdown just wasn't switched to it (the butler dropdown already is).
- Found via the Settings page showing `sonnet` for the Codex HEAD/RE1 agents.

realproject7:
Fixes #931.

## Problem
On the Settings page, the per-agent **Model** `<select>` showed only Claude models (`opus`/`sonnet`/`haiku`) for **every** agent regardless of its **Command**, with a hardcoded `"sonnet"` default. So a Codex or Gemini agent displayed `sonnet` — a model it can't use — and picking it **saved an invalid model**: the spawn path forwards it to codex via `-c model="sonnet"` / gemini `--model`, which is invalid for those CLIs.

## Root cause (`src/components/SettingsPage.tsx`)
- The agent model `<select>` mapped over a hardcoded `const MODELS = ["opus","sonnet","haiku"]` with `value={agent.model || "sonnet"}`.
- The provider-aware helper (`optionsForBackend`, wrapped by `butlerModelsForBackend`) was already imported and used by the **butler** dropdown — but the per-agent dropdown was never switched to it.

## Fix
Mirror the butler dropdown exactly:
- **Options** now come from `butlerModelsForBackend(agent.command || "claude")`, so they track the agent's command (codex → `gpt-*`, gemini → `gemini-*`, claude → `opus`/`sonnet`/pinned).
- **Default** is the backend's first option (`…[0].value`), not the hardcoded `"sonnet"`, so an unset model reads correctly for non-Claude agents.
- **On Command change**, the model is reset to a valid option for the new backend (`modelValid ? currentModel : models[0].value`) — the same logic as the butler command handler (was: `{ command }` only), preventing a Claude model from leaking onto a codex/gemini agent.
- Removed the now-unused `MODELS` const.

`butlerModelsForBackend` strips the empty `(CLI default)` option and falls back to claude when a backend has no list, so `models[0]` is always defined — identical to the butler usage that already compiles.

## Acceptance criteria
- [x] Codex agent lists codex models; Claude lists opus/sonnet/…; Gemini lists gemini models (via the shared helper)
- [x] Dropdown shows the agent's actual saved model; unset defaults to the backend's first option (not `"sonnet"`)
- [x] Changing Command resets Model to a valid option for the new command
- [x] Saving persists a model valid for that agent's CLI
- [x] `npm run build` passes

## Verification
- Frontend-only change reusing the **proven butler pattern** (same helper, same reset logic) — verifiable by inspection against lines 817–832.
- `npm run build` (includes TS typecheck) → exit 0. Full `npm test` → 42 passed, 0 failed, 2 skipped (unchanged — no server-side surface).
- Note: a live multi-CLI smoke (rendering codex/gemini agents) overlaps the #909 manual-verify; the structural fix here matches the working butler dropdown.

### DIFF
--- a/server/agentModels.test.js
+++ b/server/agentModels.test.js
@@ -1,0 +1,60 @@
+// #931: the per-agent Settings Model dropdown was hardcoded to Claude models,
+// so a codex/gemini agent could be saved with a Claude model (e.g. "sonnet")
+// that its CLI can't use. The fix routes both the dropdown and the save path
+// through the pure helpers in src/lib/agentModels.ts. This test pins those
+// helpers — most importantly that an invalid existing model (codex + "sonnet")
+// is healed to the first valid codex model on save, and that "" (CLI default)
+// is never clobbered. Plain node:assert script (run via server/run-tests.js).
+//
+// Node strips the TS types at require-time, so we can exercise the real shared
+// module the components use — no duplicated logic, no transpile step.
+
+const {
+  optionsForBackend,
+  modelsForBackend,
+  effectiveModel,
+  sanitizeModel,
+} = require("../src/lib/agentModels.ts");
+
+let passed = 0,
+  failed = 0;
+const ok = (c, m) => {
+  if (c) {
+    passed++;
+    console.log(`  PASS: ${m}`);
+  } else {
+    failed++;
+    console.error(`  FAIL: ${m}`);
+  }
+};
+
+// ── optionsForBackend: raw lists incl. the "" CLI-default row ──
+ok(optionsForBackend("codex").some((o) => o.value === ""), "optionsForBackend(codex) includes the (CLI default) row");
+ok(optionsForBackend("codex").some((o) => o.value === "gpt-5.4"), "optionsForBackend(codex) lists codex models");
+ok(optionsForBackend("nope").length === 1 && optionsForBackend("nope")[0].value === "", "optionsForBackend(unknown) → single CLI-default row");
+
+// ── modelsForBackend: concrete (non-"") options, claude fallback ──
+ok(modelsForBackend("codex").every((o) => o.value !== ""), "modelsForBackend(codex) strips the (CLI default) row");
+ok(modelsForBackend("codex")[0].value === "gpt-5.4", "modelsForBackend(codex)[0] is the first concrete codex model");
+ok(modelsForBackend("gemini")[0].value === "gemini-2.5-pro", "modelsForBackend(gemini)[0] is the first concrete gemini model");
+ok(modelsForBackend("claude")[0].value === "opus", "modelsForBackend(claude)[0] is opus (not the CLI-default row)");
+ok(modelsForBackend("nope").length > 0 && modelsForBackend("nope")[0].value === "opus", "modelsForBackend(unknown) falls back to Claude's concrete list");
+
+// ── effectiveModel: what the dropdown DISPLAYS ──
+ok(effectiveModel("claude", "sonnet") === "sonnet", "effectiveModel keeps a valid claude model");
+ok(effectiveModel("codex", "gpt-5") === "gpt-5", "effectiveModel keeps a valid codex model");
+ok(effectiveModel("codex", "sonnet") === "gpt-5.4", "effectiveModel: a stale Claude 'sonnet' on a codex agent shows the first codex model (not blank, not 'sonnet')");
+ok(effectiveModel("codex", "") === "gpt-5.4", "effectiveModel: unset model defaults to the backend's first option (#931 AC2 — not 'sonnet')");
+ok(effectiveModel("claude", undefined) === "opus", "effectiveModel: undefined model defaults to the backend's first option");
+ok(effectiveModel("gemini", "gpt-5") === "gemini-2.5-pro", "effectiveModel: a cross-backend value resolves to the gemini first option");
+
+// ── sanitizeModel: what gets PERSISTED on save ──
+ok(sanitizeModel("codex", "sonnet") === "gpt-5.4", "#931 core: saving a codex agent with 'sonnet' persists the first valid codex model");
+ok(sanitizeModel("gemini", "opus") === "gemini-2.5-pro", "sanitizeModel heals a Claude model on a gemini agent");
+ok(sanitizeModel("codex", "gpt-4o") === "gpt-4o", "sanitizeModel keeps an already-valid codex model");
+ok(sanitizeModel("claude", "claude-opus-4-8") === "claude-opus-4-8", "sanitizeModel keeps a valid pinned claude model");
+ok(sanitizeModel("codex", "") === "", "sanitizeModel keeps '' (CLI default) — valid for every CLI, never clobbered");
+ok(sanitizeModel("gemini", undefined) === "", "sanitizeModel maps undefined → '' (CLI default), not a fabricated model");
+
+console.log(`\n${passed} passed, ${failed} failed\n`);
+process.exit(failed > 0 ? 1 : 0);
--- a/src/components/AgentModelsWidget.tsx
+++ b/src/components/AgentModelsWidget.tsx
@@ -22,6 +22,7 @@
 import { useCallback, useEffect, useState } from "react";
 import InfoTooltip from "./InfoTooltip";
 import { useLocale } from "@/components/LocaleProvider";
+import { optionsForBackend } from "@/lib/agentModels";
 
 const COPY = {
   en: {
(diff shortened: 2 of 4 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The dropdown options are now provider-aware, but the PR does not fully satisfy #931 because existing invalid saved models can still remain in Settings state and be persisted unchanged.

### Findings
- [high] Existing invalid per-agent models are not normalized before display/save.
  - File: `src/components/SettingsPage.tsx:986`
  - Details: The select value is `agent.model || firstProviderModel`. If a config already has the state this ticket is trying to fix, e.g. `{ command: "codex", model: "sonnet" }`, `agent.model` wins even though `sonnet` is not in `butlerModelsForBackend("codex")`. The dropdown may render blank/mismatched, and more importantly the underlying `config` still contains `model: "sonnet"`. If the operator saves Settings after changing any other field, `save()` sends that invalid model back to `/api/config`, so the AC "Saving persists a model valid for that agent's CLI" is still not guaranteed.
  - Suggestion: derive an effective model with the same validity check used on Command change, and ensure invalid existing values are coerced into state before save (for example normalize loaded config, normalize in `save()`, or make a helper used by the row and save path). Also consider covering this with a small test/helper test if feasible: codex+sonnet should save as the first codex model, not sonnet.

### Decision
REQUEST CHANGES. The provider-specific options and Command-change reset are the right direction, but the fix needs to handle the already-bad state produced by the previous UI; otherwise #931's persistence risk remains.

realproject7:
## @re2 review — Batch 22 item 2 (last)

**Verdict: ✅ APPROVE — no findings.**

Traced the diff against issue #931 and the existing butler dropdown; ran build+typecheck and the full suite live at PR HEAD (`8abafe0`, based on current `origin/main` incl. the #932 merge).

**Fix matches the ticket and mirrors the proven butler pattern.**
- The per-agent **Model** `<select>` now maps over `butlerModelsForBackend(agent.command || "claude")` — the **same helper the butler dropdown uses** (line 831) — so options track the command: codex→`gpt-*`, gemini→`gemini-*`, claude→opus/sonnet/pinned. The hardcoded Claude-only `MODELS` const is removed (no other reference; confirmed).
- Default value is now `butlerModelsForBackend(agent.command || "claude")[0].value` instead of the hardcoded `"sonnet"`. **Crash-safe:** `butlerModelsForBackend` always returns a non-empty list (it falls back to the non-empty claude options), so `[0].value` can't throw.
- The **Command**-change handler is a faithful mirror of the butler handler (lines 816–820): compute `models = butlerModelsForBackend(command)`, keep the current model if still valid, else reset to `models[0].value`. This prevents a Claude model leaking onto a codex/gemini agent and being spawned as invalid (`-c model=` / `--model`).

**ACs:**
- AC1 (provider-aware options) ✅ — via `butlerModelsForBackend(agent.command)`.
- AC2 (shows saved model; unset → backend's first option, not "sonnet") ✅ — a now-valid saved model renders correctly; unset falls back to the backend's first concrete option. Display fallback only — an unset model is not silently persisted on save (state stays unset unless the operator picks one).
- AC3 (command change resets model) ✅ — mirrors butler.
- AC4 (saves a valid model) ✅ — options are now all valid for the backend.
- AC5 (build passes) ✅ — `npm run build` + TypeScript typecheck exit 0.

**Verification:** `npm run build` (incl. typecheck) exit 0; full `npm test` **42/0/2-skip** (frontend-only change; suite unchanged).

**Two non-blocking observations (no change requested):**
1. The issue literally suggested `optionsForBackend(agent.command)`, but `butlerModelsForBackend` (which strips the `"(CLI default)"` empty option) is the right call — it's what the butler dropdown actually uses, so the two dropdowns now behave identically. Consequence: the per-agent dropdown doesn't offer `"(CLI default)"` — but the old hardcoded list didn't either, so it's not a regression (config.json hand-edit remains the escape hatch per the widget's "guided happy path" note).
2. A pre-existing *stale* invalid saved model (e.g. a codex agent already persisted with `"sonnet"` from the old bug) won't auto-correct purely on load — only on a Command change or a manual pick. The common unset case now defaults correctly and new invalid saves are prevented; migrating already-corrupted config is out of #931's scope.

The ticket sets no test-coverage AC (unlike #932) and the repo has no React component harness, so build+typecheck + the reused butler pattern is the appropriate gate. A live multi-CLI render smoke overlaps the #909 manual-verify — not a blocker.

Ready to merge — completes Batch 22.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The fix at 223c690 resolves my prior request: provider-specific display is backed by shared helpers, and the save path now normalizes stale invalid per-agent models before persisting.

### Findings
- None.

### Decision
APPROVE.  centralizes the model catalog plus  for display and  for persistence.  now normalizes every project agent through  before PUT and writes the normalized config back into state, so an existing  is healed instead of silently re-saved. The dropdown uses  and provider-specific , and the new  pins the codex+sonnet, gemini+opus, unset, and CLI-default cases. GitHub reports no checks for this branch; review is based on live PR diff, REST PR/issue reads, branch file inspection, and the posted test/build results.

project7-interns (commented):
Correction to my approval review body: shell quoting stripped the inline code names from the Decision paragraph. Accurate decision text:

APPROVE. `src/lib/agentModels.ts` centralizes the model catalog plus `effectiveModel` for display and `sanitizeModel` for persistence. `SettingsPage.save()` now normalizes every project agent through `sanitizeModel()` before PUT and writes the normalized config back into state, so an existing `command: "codex", model: "sonnet"` is healed instead of silently re-saved. The dropdown uses `effectiveModel()` and provider-specific `modelsForBackend()`, and `server/agentModels.test.js` pins the codex+sonnet, gemini+opus, unset, and CLI-default cases. No findings remain.

realproject7:
## @re2 re-review — PR #934 (#931) @ `223c690`

**Verdict: ✅ APPROVE (maintained).** The revision fully resolves @re1's [high] finding. Credit to @re1 — the save-path gap (and the `DEFAULT_AGENTS`-seed angle that makes it reachable, not just stale data) is exactly the right escalation of what I'd under-weighted as a non-blocking note in my first pass.

**@re1's [high] (invalid existing model not normalized before display/save) — addressed on both axes:**
- **DISPLAY** — the inline dropdown value is now `effectiveModel(command, agent.model)`: a valid saved model shows as-is; a stale cross-backend value (codex agent left on `"sonnet"`) or unset value resolves to the backend's first concrete option, never blank, never the wrong model.
- **PERSIST (the catch-all that guarantees AC4)** — `save()` maps every per-agent model through `sanitizeModel(a.command || "claude", a.model)` before the PUT, so an invalid model is healed regardless of how state acquired it (old hardcoded UI *or* a `DEFAULT_AGENTS`-seeded non-Claude agent). Normalized config is written back via `setConfig`, so the panel self-heals.

**The `effectiveModel` (display) vs `sanitizeModel` (persist) split is correct and deliberate:** `effectiveModel("codex","")` → `"gpt-5.4"` (never render blank) while `sanitizeModel("codex","")` → `""` (preserve the deliberate CLI-default, never fabricate a model). `!model` short-circuits before the validity check, so `""`/`undefined` are kept as CLI-default — verified by tests.

**Refactor quality:** the catalog + helpers moved to a pure, React-free `src/lib/agentModels.ts` — `MODEL_OPTIONS`/`optionsForBackend` are **byte-identical** to the old `AgentModelsWidget` copy (diff confirmed), `AgentModelsWidget` now imports them, and no other module still imports them from the component. `modelsForBackend[0]` stays crash-safe via the Claude fallback. The test smartly exercises the **real** shared module (node strips the TS types at require-time) — no duplicated logic.

**Verification (live at `223c690`):**
- `server/agentModels.test.js` **20/0** — pins `sanitizeModel("codex","sonnet") === "gpt-5.4"`, unset → first option, `""`/`undefined` preserved, cross-backend healing.
- Full `npm test` **43/0/2-skip** (new file auto-discovered); `npm run build` + TypeScript typecheck exit 0.

**Two non-blocking observations (no change requested):**
1. **Butler save isn't run through `sanitizeModel`** — the `normalizedConfig` heals `config.projects[].agents` but not `config.butler`, so a butler left with an invalid model (never command-changed) would still re-persist it on save. Out of #931's scope (the ticket is the *per-agent* dropdown; the butler dropdown was already provider-aware and its command-change handler resets invalid models), but flagging the asymmetry — a one-line `sanitizeModel` on the butler in the same `save()` would close it for symmetry if desired.
2. An agent deliberately on `""` (CLI default) shows the first concrete model in the *inline* dropdown (no `"(CLI default)"` row there) — consistent with the pre-existing inline-dropdown behavior; the full AgentModels modal still exposes the `""` row via `optionsForBackend`. Display-only, not persisted (per #1's `sanitizeModel` preserving `""`).

Ready to merge once @re1 clears — completes Batch 22.

Issue #937, opened 2026-06-02
Settings: changing an agent's Command doesn't update mcp_inject — converting to gemini leaves stale 'flag' and crashes the CLI


realproject7:
## Problem
Sibling of #931. When you change an agent's **Command** in the Settings UI, the handler resets the **model** (`effectiveModel`, #931) but does **not** update `mcp_inject`. Since the setup wizard writes an explicit `mcp_inject` on **every** agent it creates (`bin/quadwork.js:683-688` — codex→`proxy_flag`, gemini→`env`, claude→`flag`), converting an existing agent to **gemini** leaves a stale `mcp_inject: "flag"`.

A gemini agent with `mcp_inject: "flag"` is spawned with `--mcp-config <path>` (a flag Gemini CLI doesn't accept) → **gemini crashes on launch**. Symmetric breakage exists for any conversion that crosses injection modes (e.g. claude↔codex leaves `flag` vs `proxy_flag` mismatched, so MCP chat is mis-wired).

This is **not** an edge case: on the live VPS **16/16 agents have an explicit `mcp_inject`**, so the natural way to enable a gemini agent — switch one slot of an existing team to `command: gemini` in Settings — hits it every time. (Verified during #909: the first spawn attempt crashed with `--mcp-config` until I manually deleted `mcp_inject`.)

## Root cause
- `src/components/SettingsPage.tsx` — the Command `<select>` onChange resets `model` but not `mcp_inject`.
- `SettingsPage.save()` normalizes `model` via `sanitizeModel` (#931/#935) but never reconciles `mcp_inject` with the agent's command.

## Fix
Treat `mcp_inject` like the model was treated in #931:
1. On Command-change, set `mcp_inject` to the new backend's correct default: `codex → "proxy_flag"`, `gemini → "env"`, else `"flag"` (same mapping as `bin/quadwork.js:683`).
2. In `save()`, reconcile each agent's (and the butler's) `mcp_inject` with its command so a hand-edited/legacy mismatch self-heals.
3. Factor the backend→injectMode mapping into one shared helper (it now lives in both `bin/quadwork.js` and `server/index.js`'s spawn default) to avoid drift.

## Acceptance criteria
- [ ] Converting any agent to `command: gemini` in Settings yields `mcp_inject: "env"` (no `--mcp-config`; gemini launches)
- [ ] codex→`proxy_flag`, claude→`flag` on conversion
- [ ] `save()` heals a stale/mismatched `mcp_inject` for agents and butler
[...]

PR #960, opened 2026-06-11
[#957] Lifecycle-aware cleanup of agent backend temp dirs

### LINKED ISSUE #957: Lifecycle-aware cleanup of agent backend temp dirs (/tmp/claude-{uid}) to prevent /tmp quota exhaustion
## Problem

Claude Code overrides `TMPDIR` to `/tmp/claude-{uid}` and **never cleans it up**. On Linux hosts where `/tmp` is mounted with a per-user quota (`usrquota`), this directory accumulates indefinitely — Claude's scratch/shell-snapshot/sandbox-setup files, plus any temp files the agent's commands create. Once the user's `/tmp` quota is exhausted, **every** Claude `bash` call fails silently (exit 1, empty stdout/stderr) because Claude can't write its temp/sandbox-setup files before exec. The agent becomes completely non-functional.

### Observed (VPS, Ubuntu 26.04)
- `/tmp` mounted `tmpfs … usrquota`, ~1 GB per-user quota.
- `/tmp/claude-1000` grew to **909 MB / 63,370 inodes** over ~10 days → quota exhausted.
- Symptom: every bash command (even `true`) → **exit 1, no output, no error surfaced**; the `Read` tool, plain `/bin/bash`, network, and Codex all worked fine.
- **Extremely hard to diagnose** — it masquerades as a bwrap/AppArmor/user-namespace sandbox failure. It is NOT: `bwrap` smoke tests pass, userns works, the sandbox is never even reached. The real signal was a `Disk quota exceeded` on a `/tmp` write. Clearing `/tmp/claude-1000` immediately restored `bash` → exit 0.
- macOS unaffected (no `/tmp` quota; different temp layout).

## Why this belongs in QuadWork

A host-level `cron`/`systemd-tmpfiles` purge fixes one machine, but **every QuadWork install on a Linux host with a `/tmp` quota is exposed** and will hit a total, silent agent outage that's near-impossible to diagnose. QuadWork owns the agent lifecycle, so it's the right layer to keep backend temp bounded — and it can clean at known-safe moments (agent teardown) rather than blindly.

## Proposed scope

1. **On agent teardown** (reset / full reset / restart / stop): purge the backend's temp dir for that uid, stale entries only:
   - Claude: `/tmp/claude-{uid}`
   - Codex / Gemini: confirm + cover their equivalent temp dirs.
2. **Periodic sweep** from the server process (e.g. hourly): remove `/tmp/claude-*` (+ codex/gemini equivalents) entries older than a conservative age (**48–72h**), age-based on atime so currently-active session files are spared.
3. Make enable/age **configurable** with safe defaults.

## Acceptance criteria

- [ ] Backend temp dirs stay bounded well under typical `/tmp` quotas during multi-day continuous operation.
- [ ] No active session's temp is deleted mid-use (conservative age + atime-based selection).
- [ ] Covers `claude` + `codex` (+ `gemini`) backends.
- [ ] Cross-platform safe — Linux primary; macOS/Windows = harmless no-op.
- [ ] Cleanup runs both on agent teardown AND on a periodic timer.
- [ ] Documented (`docs/troubleshooting.md` + a note in `docs/install-vps.md`).

## Notes

- **Immediate host-level mitigation** for anyone hitting this now: `systemd-tmpfiles` drop-in or cron purging `/tmp/claude-*` older than ~72h, and/or raise the `/tmp` per-user quota.
- Likely also an **upstream Claude Code bug** (it should bound its own `TMPDIR`) — worth a separate upstream report.

realproject7:
Fixes #957

## What

Adds `server/temp-cleanup.js` — a stale-only sweep of agent backend temp — wired into two lifecycle points in `server/index.js`:

- **Agent teardown** (`stopAgentSession`): deferred fire-and-forget sweep after every stop/reset/restart.
- **Periodic**: hourly timer + one sweep at boot (a server that was down for days reclaims quota immediately).

Targets (confirmed on the VPS that hit #957):
- `/tmp/claude-{uid}` — Claude Code's unbounded TMPDIR override (the 909 MB / 63k-inode quota killer).
- `/tmp/gemini-client-error-*.json` — stray Gemini crash dumps (observed on the VPS).
- Codex: confirmed **no** `/tmp` usage (state lives in `~/.codex`) → nothing to sweep.

## Safety design

- **Stale-only everywhere**: `/tmp/claude-{uid}` is shared by every claude agent of the user, so teardown of one agent must never delete another live agent's files. Deletion requires age > `max_age_hours` (default **72h**) on `max(mtime, atime, ctime)` — systemd-tmpfiles semantics, so anything a live session touches stays fresh.
- **Never throws**: errors are collected, logged, and can't break the stop path, boot, or the timer (`backendTempSweepTick` also has its own try/catch + re-entrancy guard).
- **Cross-platform no-op**: no `process.getuid` (Windows) → claude sweep skipped; gemini glob matches nothing where absent. Non-matching files at the temp root are never touched.
- **Configurable**: `temp_cleanup: { enabled, max_age_hours }` in `config.json` (default on / 72h; `enabled:false` opts out).
- **Packaging**: module lives under `server/` (shipped dir) per the #937/#939 rule; pack-smoke covers it.

## Docs

- `docs/troubleshooting.md`: new section for the silent-exit-1 symptom (it masquerades as a sandbox failure) with diagnosis + manual recovery.
- `docs/install-vps.md`: note about `usrquota` VPS images.

## Verification

- New `server/temp-cleanup.test.js`: stale file/dir removed (recursively), fresh spared, claude dir itself kept, gemini dumps matched, unrelated old files never touched, missing dir / Windows-no-uid / bogus root → clean no-op, settings defaults + opt-out. All fixture-based with injected uid/now/tmpRoot — never touches the real /tmp.
- Full suite: **53 passed, 0 failed** · `node --check` both files · boot smoke (`QUADWORK_SKIP_LISTEN=1 require`) OK · `npm run build` OK.

### DIFF
--- a/docs/install-vps.md
+++ b/docs/install-vps.md
@@ -362,3 +362,12 @@ chmod 600 ~/.quadwork/.env
 14. Configure nginx reverse proxy + SSL
 15. Add HTTP basic auth
 16. Verify reboot survival: `sudo reboot`, then check `pm2 list`
+
+## Note: /tmp quotas and Claude temp
+
+Some VPS images mount `/tmp` with a per-user quota (`usrquota`). Claude Code
+accumulates temp under `/tmp/claude-{uid}`; if the quota fills up, every
+Claude bash command starts failing silently with exit 1 (see
+[troubleshooting](troubleshooting.md#every-claude-bash-command-fails-silently-exit-1-no-output)).
+QuadWork sweeps stale entries automatically (hourly + on agent teardown,
+72h age) — configurable via `temp_cleanup` in `~/.quadwork/config.json`.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -124,3 +124,29 @@ chmod 440 /etc/sudoers.d/quadwork
 
 See the [VPS Installation Guide](install-vps.md#step-2-create-non-root-user-critical) for full setup.
 
+
+## Every Claude bash command fails silently (exit 1, no output)
+
+**Symptom:** every bash command a `claude` agent runs — even `true` — returns
+exit 1 with empty stdout/stderr. The Read tool still works, a plain shell on
+the host works, and `codex` agents are unaffected. Easily mistaken for a
+bwrap/AppArmor/sandbox failure.
+
+**Cause:** Claude Code keeps its temp under `/tmp/claude-{uid}` and never
+cleans it up. On hosts where `/tmp` is mounted with a per-user quota
+(`usrquota`), that dir grows until the quota is exhausted — after which Claude
+can't write the temp files it needs before executing any command (#957).
+
+**Check:** `du -sh /tmp/claude-$(id -u)` and try `dd if=/dev/zero
+of=/tmp/probe bs=1M count=10` — a "Disk quota exceeded" error confirms it.
+
+**Fix:** QuadWork sweeps stale backend temp automatically (hourly, at boot,
+and on agent teardown; entries older than 72h). If you hit the quota *before*
+a sweep (e.g. the server was down), clear it manually:
+`find /tmp/claude-$(id -u)/* -maxdepth 0 -mmin +60 -exec rm -rf {} +`
+
+Tune or disable via `~/.quadwork/config.json`:
+
+```json
+{ "temp_cleanup": { "enabled": true, "max_age_hours": 72 } }
+```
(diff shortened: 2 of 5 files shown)


realproject7:
Closing this stale PR (opened 2026-06-11, 0 reviews). It now conflicts with main — server/index.js was heavily reworked since by #968 (WS auth), #972 (shutdown cleanup), and #975 (event-loop de-block). Rather than untangle a 3-week-old conflict, #957 will be reimplemented fresh against current main under the v7 review process; this PR's server/temp-cleanup.js module design is a useful reference. Superseded by a fresh task/957 branch.

Issue #905, opened 2026-06-01
Complete Gemini CLI agent support — wire the model flag + setup-wizard provider option


realproject7:
## Summary
QuadWork has **partial** Gemini CLI support, but the **model is never passed to the CLI** — a Gemini agent's configured model is silently dropped, so "the gemini model setup" doesn't take effect. Finish Gemini so it's a first-class agent provider alongside `claude` / `codex`.

## Already works (do NOT rebuild)
- CLI detection: `isCliInstalled("gemini")` ([server/index.js:91](server/index.js#L91))
- Permission flags: `PERMISSION_FLAGS.gemini = ["--yolo"]` ([:267](server/index.js#L267))
- MCP / file-chat injection via **env mode**: `buildAgentEnv` writes a Gemini `settings.json` with the chat-shim `mcpServers` and sets `GEMINI_CLI_SYSTEM_SETTINGS_PATH` (the `cliBase === "gemini"` path)
- Frontend model dropdown: `MODEL_OPTIONS.gemini` ([AgentModelsWidget.tsx:127](src/components/AgentModelsWidget.tsx#L127)) — `gemini-2.5-pro`, `gemini-2.5-flash`

### A. Model flag not passed (core fix — "the model setup")
The model-arg block (`server/index.js` ~382–394) branches only for `codex` (`-c model="…"`) and `claude` (`--model …`) — there is **no `gemini` case**, so `agentCfg.model` is never forwarded. Add a `gemini` branch that passes the model via the Gemini CLI's flag — **verify the exact flag against the installed CLI** (Gemini CLI uses `-m` / `--model <model>`). Fall back to the CLI default when unset. Decide on reasoning/"thinking" handling (optional — model first).

### B. Setup wizard has no gemini provider option
`SetupWizard.tsx` lets users pick the per-agent CLI (claude / codex) at project creation but offers **no gemini option** — users currently must hand-edit `config.json`. Add `gemini` to the per-agent provider/command selection (command `gemini`, model dropdown from `MODEL_OPTIONS.gemini`), persisted to config.

### C. (minor) Refresh the gemini model list
Confirm `MODEL_OPTIONS.gemini` reflects current Gemini models; add newer ones if available (safe additions only).

## Acceptance criteria
- [ ] A gemini agent receives its configured `model` on the spawned command (assert the model flag is present); CLI default when unset
- [ ] Setup wizard offers **gemini** as a per-agent provider (command + model), persisted to config
[...]

PR #1024, opened 2026-08-01
[#1023] Add Grok (xAI Grok Build CLI) as a fourth agent backend

### LINKED ISSUE #1023: Add Grok (xAI Grok Build CLI) as a fourth agent backend
## EPIC Context

Standalone enhancement — no parent EPIC. QuadWork currently supports three agent CLI backends (`claude`, `codex`, `gemini`). This ticket adds xAI's official Grok Build CLI as a fourth, following the exact seams the existing three use.

## Goal

An operator can select `grok` as an agent's CLI in Setup Wizard / Settings, pick a model for it, and have the agent spawn with working permission bypass and file-chat MCP connectivity — with no new abstractions beyond the one genuinely new mechanism Grok requires (project-scoped TOML MCP config + folder trust).

## Verified CLI facts

All facts below were verified on **`@xai-official/grok` v0.2.118** (binary `grok 0.2.118 (1e1687c1cf6a)`) on 2026-08-01, against the CLI's own `--help`, its bundled official docs at `~/.grok/docs/user-guide/`, and live `grok inspect` runs. Do not substitute assumptions from blog posts.

- **Binary**: `grok` (npm `@xai-official/grok`; native binary bootstrapped to `~/.grok/bin/`). Default mode is an interactive TUI — fits the existing pty spawn path.
- **Model flag**: `-m, --model <ID>`. Authenticated `grok models` on the operator's account lists exactly one model: `grok-4.5` (also the default). No "latest"-style alias exists, so the `""` (CLI default) row is the auto-tracking mechanism.
- **Permission bypass**: `--always-approve` (alias `--yolo`; equivalent to `--permission-mode bypassPermissions`). `22-permissions-and-safety.md` explicitly recommends always-approve for automation. Use `--always-approve`.
- **MCP injection — the one new mechanism.** Grok has **no `--mcp-config`-style flag**. Sources, per `07-mcp-servers.md`:
  | Source | Reliability |
  |---|---|
  | `~/.grok/config.toml`, `<cwd>/.grok/config.toml` (native TOML) | **"Always on"** — first-class, highest merge priority |
  | `~/.claude.json`, `.cursor/mcp.json` | vendor compat, toggleable |
  | `.mcp.json` | **conditional** — "loaded unless you have imported or dismissed the Claude import prompt" |
  `.mcp.json` is therefore NOT a safe channel (a dismissed import marker silently disables it). Use project-scoped `.grok/config.toml`, which is per-agent by construction because every QuadWork agent already has its own worktree cwd (`resolveAgentCwd`, `server/config.js:95`).
- **Folder trust gates it.** Verified live: a temp git repo containing `.grok/config.toml` with an `[mcp_servers.chat]` entry shows `chat (stdio) config` under `grok inspect` → MCP Servers, but the same output reports **`Project trusted: no`**. Repo-local MCP/LSP/hooks only *run* once the folder is trusted. `--trust` grants it at launch (recorded in `~/.grok/trusted_folders.toml`, cascades to subdirectories) — the flag is hidden from `--help` but is real and parses (verified against a negative control: `--bogusflag` errors, `--trust` does not).
  **End-to-end proof (run during research, not an inference):** in that same temp repo, `grok --trust -p "Reply with exactly: OK"` returned `OK` with exit 0 and wrote `[folders."<repo>"] trusted = true` into `~/.grok/trusted_folders.toml`. Two things this establishes for the spawn path: `--trust` works on a **non-interactive** launch, and an MCP server that fails to connect (the shim was pointed at a dead port in that test) does **not** block or hang CLI startup.

- **TUI repaints continuously when idle — this is load-bearing, see item 9.** Measured by spawning each CLI under `node-pty` (120x40) in a temp git repo and timestamping every `onData`, over a **40-second idle window** (first 20 s discarded as startup):
  | CLI | events in window | max quiet gap | verdict |
  |---|---|---|---|
  | **grok** | 195 | **2551 ms** | never quiet for 5 s |
  | codex (control) | 0 | 40000 ms | goes quiet |
  The codex control validates the probe — it reproduces the documented codex behavior, so the grok number is not a measurement artifact. **Grok is in the same class as Claude, not codex/gemini.**

## Where (paths verified against main @ 5919895)

> **Numbering note:** there is no item 8 — it is a renumbering artifact from the revision that inserted item 9, not dropped content. The former item 8 (Tests) is now item 10. Items 6, 7, and the Scope exclude-path were corrected after ticket review (see the revision note at the end).

> Item 4 caveat: `~/.grok/bin` is added to the *generic* `isCliInstalled` fallback list, so the helper will also probe `~/.grok/bin/claude` etc. Harmless (the path simply will not exist) — add a one-line comment saying so, rather than special-casing per CLI.

1. `src/lib/injectMode.js:27-30` — `injectModeForCommand`: add `grok → "project_toml"`.
2. `server/index.js:432-436` — `PERMISSION_FLAGS`: add `grok: ["--always-approve"]`.
3. `server/index.js:521-585` — `buildAgentArgs`:
   - model branch: `grok` gets `--model <slug>` (same shape as the gemini branch at 563-567);
   - new `project_toml` inject mode: write `<agent cwd>/.grok/config.toml` containing an `[mcp_servers.chat]` stdio entry whose `command`/`args` are the same shim invocation `writeFileChatMcpConfig` already builds (`node <shimPath> --project <id> --agent <id> --port <PORT> --token <token>`), registering the token via the existing `fileChat.registerShimToken` path; then append `--trust` to args.
4. `server/index.js:255-261` — `/api/cli-status`: add `grok: isCliInstalled("grok")`; add `~/.grok/bin` to the `isCliInstalled` fallback list at 246-250 (covers the `curl | bash` installer, which does not put `grok` on the npm global path).
5. `src/lib/agentModels.ts:10-52` — `MODEL_OPTIONS.grok`: `""` (CLI default) first, then pinned `grok-4.5`. Append-only pattern per the #999/#958/#1018 comments.
6. **`src/components/SetupWizard.tsx`** — the CLI dropdown is rendered from a constant, not from the status literals:
   - `:188-192` — **`BACKENDS`** (`[claude, codex, gemini]`, rendered at `:822`). **This is the dropdown source; adding `grok` here is what makes it selectable.**
   - `:333`, `:350` — the `cliStatus` state type and the fetch's response type (`{claude, codex, gemini}`), plus `:353` and `:795` — the `(["claude","codex","gemini"] as const)` installed-filter tuples. With `grok` added to `:353`, a grok-only machine correctly lands in the `installed.length === 1` branch at `:354-356` and defaults all four agents to grok.
   - **Not edit sites** (an earlier revision of this ticket wrongly listed them): `:359` is the claude+codex mixed-default preset, which stays correct as written; `:793` is a comment line — the literal is `:795`.
7. **`src/components/SettingsPage.tsx`** — same shape, plus a fallback that currently cannot express grok:
   - `:68-72` — **`BACKENDS`** (rendered at `:944`, `:1083`, `:1232`). The dropdown source; add `grok`.
   - `:383` — `cliStatus` is typed `{claude: boolean; codex: boolean}` (gemini is already absent). Without adding `grok` here it renders permanently `disabled` + "(not installed)" at `:1237`.
   - `:711-716` — `addProject`'s default-backend selection. The availability check casts to `"claude" | "codex"` and the fallback chain hardcodes claude/codex, so on a **grok-only** machine it still seeds the unavailable `claude`. Replace the hardcoded chain with a generic "first installed backend" fallback that can return grok. (Previously this ticket cited only `:711` and called it the dropdown — it is not.)
     **Operator ruling on the scope tension raised in review:** this generic fallback also changes **gemini-only** machines, which today seed the uninstalled `claude`. That is intended, not collateral — it is the same defect, and the hardcoded two-backend chain structurally cannot express a fourth backend. It does not conflict with the "no generic backend abstraction" exclusion below, which forbids a backend *registry*, not a loop over the installed set. The gemini-only fix is in scope and carries its own AC.
9. **`server/pty-dispatcher.js:128-130` — `capEligible`: extend the #1010 bounded deferred-wake cap to grok.** Without this, a grok agent silently never receives `@mentions`: `queuePendingWake` drains on a 5 s idle gap in the TUI's output, grok never produces one (measured above), and the cap that rescues Claude from exactly this starvation is gated to `session.backend === "claude"`. Change the predicate to admit `"grok"` as well, and update the now-inaccurate comments at lines 11-19 and 132-137 (they assert the cap is Claude-only *because* codex/gemini go quiet — still true of those two, no longer the full story).
   Keep the fail-safe direction intact: a session with no `backend` stamp stays ineligible.
   **The cap's safety premise must be re-stated for grok, not just reworded (review round 4).** The comment at `:230-231` justifies the bounded barge-in with "safe because the cap is Claude-only and Claude queues input received mid-turn rather than interrupting" — a claim that becomes false the moment grok is eligible, and which nobody had verified for grok. **The PO verified it: the premise holds.** Grok's own docs (`~/.grok/docs/user-guide/03-keyboard-shortcuts.md:182-183`) state that a plain `Enter` with text in the composer **queues** a follow-up which runs after the current turn ends; only a **bare** `Enter` on an *empty* composer force-sends, and cancel-and-send is a separate chord. `injectIntoTerm` always writes the text before its single `\r`, so an injection queues and cannot cancel a running grok turn. Cite that source in the corrected comment so the next reader inherits the evidence instead of the assertion.
10. Tests — all runnable **without** the grok binary, per the `server/agentArgs.gemini.test.js` (#905) precedent:
   - `server/agentModels.test.js` — grok catalog rows plus `sanitizeModel`/`effectiveModel` healing;
   - `server/injectMode.test.js` — `grok → "project_toml"`;
   - new `server/agentArgs.grok.test.js` — asserts `--always-approve`, `--model` forwarding (and its absence when unset), `--trust` present, and the written `.grok/config.toml` content;
   - `server/pty-dispatcher.test.js` — a grok session's deferred wake drains via the cap under continuous output (mirror the existing Claude repaint regression, which already runs at millisecond scale via the `_timings` setter), **plus a negative case**: a codex/gemini session under the same continuous output must still NOT use the cap;
   - `server/configFlags.test.js` — **checked: it does not enumerate backends** (fixture-only references). Expect **no change** here; do not go hunting.

   **Fixture containment (from review round 3):** the plain-clone fixture must `process.chdir()` into its temp dir and restore in a `finally`. Against a naive relative-path implementation — exactly the bug the fixture exists to catch — the write follows the test process's cwd, which under `npm test` is the QuadWork checkout itself, so a failing test would mutate the developer's or CI's own `.git/info/exclude`. The fixture must not be able to reach the real repo.

   Fixture note for the `buildAgentArgs` tests: `resolveAgentCwd` returns **null** when an agent config has no `cwd` (`server/config.js:100`). Production always has one (spawn hard-fails at `server/index.js:629` otherwise), but #905-style fixtures do not set it, so the grok branch needs either a null guard or a `cwd` in every fixture. Pick one and be consistent.

## Implementation notes (decided — do not re-litigate in review)

- **Write the TOML by hand.** Five lines of fixed-shape output; serialize each string value with `JSON.stringify` (TOML basic strings and JSON strings share quoting/escaping semantics over this value domain: absolute paths, slug ids, a numeric port, a hex token). **Adding a TOML library is an automatic rejection** (no new `dependencies`).
- **Rejected alternative:** shelling out to `grok mcp add --scope project`. It requires the grok binary at spawn time and couples the write to process cwd, for no gain over writing the file directly.
- **The `pty-dispatcher.js` change (item 9) is required, not optional.** An earlier revision of this ticket said to leave that file alone on the assumption grok behaves like codex/gemini; the measurement above disproved it. Skipping item 9 ships a grok backend that looks correct in every unit test and is inert in production.

## Scope

- The four-backend seam extensions above, one new inject mode (`project_toml`), catalog + UI literals, tests.
- Exclude `.grok/` from git when writing the config, so the file (which embeds the per-spawn shim token) can never be committed. **Write to `$(git rev-parse --git-common-dir)/info/exclude`, NOT `<cwd>/.git/info/exclude`.** QuadWork creates agent cwds with `git worktree add` (`server/routes.js:3785`), so in an agent worktree `<cwd>/.git` is a **file**, not a directory — verified on this machine: `ls ~/Projects/quadwork-head/.git` → a 66-byte regular file, and `.git/info` → `Not a directory`. The three candidate paths were measured in a real linked worktree (git 2.53.0): `<cwd>/.git/info/exclude` → **ENOTDIR** (throws on every spawn); per-worktree `$(git rev-parse --git-dir)/info/exclude` → written but **not honored** (`.grok/` still shows `??`); `$(git rev-parse --git-common-dir)/info/exclude` → **works** in both the agent and main worktrees.

## Out of scope

- `--reasoning-effort` wiring for grok (the CLI supports it; defer until requested — model-only, like gemini).
- Auth management, VPS grok provisioning, `XAI_API_KEY` handling. Login is an operator action (`grok login`), already completed on the local Mac.
- Any backend-registry refactor. Follow the existing map/branch pattern verbatim; a "generic backend abstraction" is an explicit over-engineering rejection criterion for this review.

## Acceptance criteria

- With `command: "grok"`, **`buildAgentArgs`** returns exactly: `--always-approve`, optional `--model <slug>`, `--trust`. (Scoped to `buildAgentArgs` deliberately: `PERMISSION_FLAGS` has a second consumer at `server/index.js:1116` — the Butler — which builds its own args and is out of scope here.)
- `<cwd>/.grok/config.toml` exists after arg-build and contains an `[mcp_servers.chat]` stdio entry with `command = "node"` and args carrying the shim path, `--project`, `--agent`, `--port`, and the registered `--token`. Assert file **content** in the unit test, not just existence.
- **Resolve the common-dir output against the agent cwd — `path.resolve(agentCwd, out)`, never `path.join(out, …)` alone.** `git rev-parse --git-common-dir` returns an **absolute** path from a linked worktree but a **relative** `.git` from a plain clone (measured, git 2.53.0). `resolveAgentCwd` returns `agent.cwd` verbatim (`server/config.js:99-101`) and nothing forces it to be a linked worktree — operators edit `~/.quadwork/config.json` by hand. Unresolved, a plain-clone agent cwd makes the exclude land relative to the **server process's** cwd: the write silently hits the wrong repo while that agent's token-bearing `.grok/` stays committable, defeating the security invariant — and a linked-worktree-only test would still pass. Add a plain-clone (non-worktree) fixture case that asserts the exclude lands under the agent's own cwd.
- **The exclude write must be idempotent — append only if `.grok/` is not already present.** The common-dir exclude is *shared* by all four agent worktrees and `buildAgentArgs` runs on **every spawn and respawn** (#992 startup respawn, restart flow), so a naive append grows the file without bound (measured: 6 spawns → 6 identical `.grok/` lines). Assert that a second arg-build leaves exactly one `.grok/` line.
- After arg-build, `.grok/` is excluded from git. **The test fixture must be a real linked worktree** (`git init` a temp repo with one commit, then `git worktree add`), not a plain temp dir containing a `.git` directory. A `.git`-dir fixture makes the assertion vacuous — it passes green against the exact code shape that throws `ENOTDIR` in production. Assert both that the write lands in the common dir and that `git status --porcelain` in the worktree does **not** list `.grok/`.
- On a machine where `/api/cli-status` reports **grok as the only installed CLI**, `addProject`'s default backend is `grok` (not `claude`) — the `SettingsPage.tsx:711-716` regression. Same for a **gemini-only** machine (the pre-existing bug fixed deliberately; see the operator ruling in item 7).
  **PO ruling, review round 4 — how this AC is satisfied.** Implement the fallback as a single inlined expression over the shared `BACKENDS` constant at the call site. Do **not** extract a helper module for it. @re1 correctly flagged that a new `src/lib/defaultBackend.js` with one production caller is the one-caller abstraction this ticket's own over-engineering criteria forbid, and the logic is a single `find` — the `injectMode.js` / `agentModels.ts` precedents both carry real domain data across two or more call sites, which this does not.
  Consequence, stated plainly rather than papered over: **these two criteria are no longer covered by a unit test.** A test against an inlined React expression could only mirror the logic, and this ticket spent three review rounds removing exactly that class of test. They are satisfied instead by (a) inspection of the inlined expression, which structurally cannot express the old two-arm bug, and (b) the PO's live check on the local Mac after merge. The verification burden moves to the PO rather than being faked with a mirror test.
- `sanitizeModel("grok", "sonnet")` heals to `"grok-4.5"`; `sanitizeModel("grok", "")` stays `""`; unknown backends still fall back to Claude's list (#931 invariants preserved).
- `capEligible({backend:"grok"})` is true; `capEligible({backend:"codex"})` and `capEligible({})` stay false.
- Setup Wizard and Settings offer `grok` only when `/api/cli-status` reports it installed, matching the gating of the other three.
- Full suite green, `tsc` clean, `npm run build` passes. **No new `dependencies`** (automatic rejection).
- No claude/codex/gemini spawn-arg regression — existing tests stay green.

## Security invariants

- The shim token appears only inside the per-agent `.grok/config.toml` in the worktree (git-excluded) — never in logs, never in a tracked file, never in a chat message.
- No new runtime dependencies; every module reachable from `bin/`/`server/` stays inside the `package.json` `files` allowlist (#937/#939 pack-smoke guard).

## Routing

`agent/quadwork` (VPS) — headless and fully unit-testable; the grok binary is **not** required for any test (assert args and written file content, per the #905 gemini precedent). Live spawn verification against a real authenticated grok is a post-merge PO step on the local Mac, not part of this ticket's CI gate.

## Post-merge PO verification (not dev scope, recorded here so it is not lost)

Spawn a real grok agent on the local Mac and confirm, from the agent's worktree: `grok inspect` reports `Project trusted: yes` and lists `chat (stdio) config`, the agent can call `chat_send`/`chat_read`, and — the item unit tests cannot cover — an `@mention` posted to a **live idle grok agent** actually lands in its TUI within the cap window. That is the end-to-end proof for item 9. Observed during research: grok also merges the operator's `~/.claude.json` MCP servers into every session by default (compat source) — expected behavior, not a defect, but it means a grok agent sees those tools too.

---

## Revision note — ticket review round 1 (Batch 50)

@re1 and @re2 both returned REQUEST CHANGES. Every finding was independently re-verified against main `5919895` by the PO before amending; all four were correct and are now folded in.

- **Item 9 (the load-bearing claim) — CONFIRMED by both reviewers, and by an independent probe.** @re2 loaded main's `pty-dispatcher.js` verbatim and drove a fake term repainting at half the idle threshold: `backend=grok` → 0 injections over 5× the cap with the wake still pending and the cycle still armed; `backend=claude` → 1 injection; `codex` and an unstamped session → 0 (fail-safe preserved). @re1 traced the same chain through `server/index.js:680` → `pty-dispatcher.js:116-117` → `:318-320` → `:249-264`. The mechanism is in fact *stronger* than this ticket states: a continuously repainting agent is `isAgentBusy` forever (`:107`), so **every** mention to a grok agent takes the defer path and never reaches `scheduleCoalescedInjection`.
- **F1 (blocking, fixed above)** — the `.git/info/exclude` path was invalid for linked worktrees, and the prescribed test fixture pinned the wrong shape. Corrected to `--git-common-dir` with a real-worktree fixture requirement.
- **F2 (blocking, fixed above)** — items 6 and 7 omitted the two `BACKENDS` constants that actually render the CLI dropdown, so the Goal was unsatisfiable as written. Both are now cited, along with the `cliStatus` types and the grok-only default-backend fallback.
- **F3 (fixed above)** — two of item 6's cited lines were not literal lists; corrected and explicitly marked as non-edit sites.
- **F4 (answered above)** — the missing item 8 is a renumbering artifact; no content was dropped.

Both reviewers explicitly endorsed the two over-engineering constraints (no backend registry, no TOML library) and confirmed every acceptance criterion is assertable without the grok binary.

## Revision note — ticket review round 2 (Batch 50)

@re1 APPROVED the amended sections with no new finding. @re2 confirmed all four round-1 fixes independently and raised two narrow defects in the newly-written exclude requirement, both re-verified by the PO and now pinned as acceptance criteria:

- **N1 — `--git-common-dir` output is relative outside a linked worktree.** Re-measured on the operator's machine: from `~/Projects/quadwork-head` (linked) it returns an absolute path; from `~/Projects/quadwork` (main) it returns `.git`. Unresolved, a plain-clone agent cwd sends the exclude to the server process's own repo while the token-bearing `.grok/` stays committable. Pinned: `path.resolve(agentCwd, out)` plus a non-worktree fixture case.
- **N2 — the shared exclude write must be idempotent.** `buildAgentArgs` runs on every spawn and respawn against a common-dir file shared by all four worktrees; a naive append grows it without bound. Pinned: append-if-absent, asserted by a repeat arg-build.
- **Scope tension resolved by operator ruling** (recorded in item 7): the generic default-backend fallback also fixes gemini-only machines. Intended and now carries its own AC, rather than riding along unremarked.

## Ticket review — CLOSED (Batch 50, round 3)

**@re1 APPROVE · @re2 APPROVE.** Three rounds, six findings, all folded in. @re2 mechanically diffed the round-2 body against the live one to confirm "nothing else changed" (four insertion hunks; items 1-5, 9, 10 byte-identical), so the round-1 confirmation of item 9 and the round-2 verification of items 6/7 carry forward. One non-blocking test-hygiene note (fixture containment) was folded into item 10 without a further round. The ticket is implementable as written.

realproject7:
Closes #1023

Adds `grok` alongside claude/codex/gemini across the existing seams, plus the one genuinely new mechanism grok needs: a project-scoped TOML MCP config and folder trust. **10 files, +501/−28. No new dependencies.** (An earlier revision of this body said "12 files, +279/−26" — the insertion count was simply wrong, taken from a working-tree stat before the new test files existed. Corrected rather than left to be caught.)

## EPIC Alignment
- **Parent:** none — standalone enhancement. The ticket's `## EPIC Context` block states this explicitly ("Standalone enhancement — no parent EPIC"), so there is no epic body to read, no `Depends on:` siblings, and no contract to consume.
- **Epic goal:** n/a (standalone). The ticket's own goal: an operator can select `grok` as an agent's CLI, pick a model, and have it spawn with working permission bypass and file-chat MCP connectivity.
- **This PR's role:** extends the four existing backend seams (`injectMode`, `PERMISSION_FLAGS`, `buildAgentArgs`, `/api/cli-status`), the model catalog, the two UI dropdown constants, and the #1010 deferred-wake cap — adding one new inject mode (`project_toml`) for the only mechanism grok doesn't share with the other three.
- **Contracts consumed/exposed:**
  - Consumes `injectModeForCommand` (`src/lib/injectMode.js:32`), the existing single source of truth for backend → inject mode, rather than adding a parallel mapping. The heal path (`server/routes.js:4054`) derives from the same helper, so a grok agent heals automatically with no extra wiring.
  - Consumes `fileChat.registerShimToken` / `validateShimToken` — the same token lifecycle `writeFileChatMcpConfig` already uses.
  - Exposes `writeGrokProjectConfig` / `excludeGrokFromGit` (`server/index.js:543`, `:576`), module-private to `server/index.js`; and `capEligible` (`server/pty-dispatcher.js:380`), newly exported for testing only.
- **Fits with merged siblings:** extends rather than duplicates — the #1010 cap became a set instead of gaining a second predicate; the #931/#958/#999/#1018 model catalog got an appended row following the documented append-only pattern (`modelsForBackend("grok")[0]` is the heal target, so future pins go below it); the #905 gemini precedent is the test shape; the #937/#939 pack allowlist is untouched.

## Self-Verification
- Build: `npm run build` → **succeeds, 9/9 static pages**, "Compiled successfully in 3.1s"
- Typecheck: `npx tsc --noEmit` → **clean (exit 0)**
- Tests: `npm test` → **65 passed / 0 failed / 2 skipped** (the 2 are the pre-existing #836 Jest-style skips; baseline on main was 64/0/2, +1 net new file)
- **macOS**: the grok fixture failed on the operator Mac at `65add1f` (`os.tmpdir()` is a symlink there and git canonicalizes what it reports). Reproduced that exact failure on Linux under a symlinked `TMPDIR` — **pre-fix 32/1 on the same assertion, post-fix 33/0**, and 33/0 on normal `/tmp`. Audited the whole file: it was the only raw-path comparison.
- Pack: `npm pack --dry-run` → 150 files, 1.1 MB packed / 3.3 MB unpacked. `git diff --stat package.json package-lock.json` → **empty: no new dependencies** (a TOML library is an automatic rejection per the ticket, so this is asserted rather than assumed).
- **Mutation-tested — every guard this PR adds fails a test when reverted.** A green suite is not proof a guard is pinned (#1020), so I reverted each one and re-ran:

  | Guard reverted | Result |
  |---|---|
  | `capEligible` drops the grok arm | **3 failures** (incl. the behavioral repaint regression, not just the predicate) |
  | `path.resolve(agentCwd, out)` → naive `path.join(out, …)` | **3 failures** |
  | append-if-absent → blind append | **1 failure** (`got 3`) |
  | `--trust` not pushed | **3 failures** |
  | `injectMode` drops the grok arm | **4 + 4 failures** |
  | `PERMISSION_FLAGS` drops grok | **4 failures** |
  | grok `--model` branch removed | **2 failures** |
  | grok model catalog removed | **5 failures** |
  

- Kill-list scan: **clean** — no mocks/stubs in runtime paths, no hardcoded secrets (the shim token is generated per spawn and written only to a `0600` git-excluded file), no TODO/FIXME, no `console.log`/debugger, no dead code, no copy-pasted utils. One new single-purpose module is disclosed in **Deviations** rather than left for a reviewer to find.
- Acceptance criteria 1:1: all 11 ticket ACs checked against the diff; each is asserted by a test that fails when its guard is reverted, **except** the `~/.grok/bin` `isCliInstalled` fallback (item 4) — see **Deviations**.
- Manual check: **NOT VERIFIED by live spawn** — deliberately. The ticket routes live grok verification to a post-merge PO step on the local Mac and states the grok binary is not required for any test. No throwaway port was exercised; every claim above is from the unit suite and the written-file assertions.

## Design Fidelity
This PR touches two UI files. It introduces **no new visual design** — grok is one additional row in two existing dropdowns and inherits their styling wholesale, so the fidelity rows below record that the new row matches the three existing ones rather than inventing anything.

| Spec element | Spec says | Implemented | Where |
|---|---|---|---|
| Backend dropdown row | grok selectable alongside the other three | `{ value: "grok", label: "Grok CLI" }` appended to `BACKENDS` | `src/components/SetupWizard.tsx:192`, `src/components/SettingsPage.tsx:73` |
| Label style | matches siblings ("Claude Code", "Codex", "Gemini CLI") | `"Grok CLI"` — same `<Product> CLI` form as Gemini | `SettingsPage.tsx:73` |
| Ordering | append-only, existing rows undisturbed | added last; claude/codex/gemini keep their positions and the claude+codex mixed-default preset is untouched | `SetupWizard.tsx:188-193` |
| Install gating | offered only when `/api/cli-status` reports it installed, matching the other three | unchanged rendering path — `disabled` + `" (not installed)"` now resolve for grok because the row and the `cliStatus` type exist | `SetupWizard.tsx:828`, `SettingsPage.tsx:953`, `:1092`, `:1244` |
| Single-CLI wizard copy | grok-only machine gets the same "You have X installed" block | `"grok"` added to the installed-filter tuples; label read from `BACKENDS` | `SetupWizard.tsx:354`, `:796` |
| Colors / spacing / type | dark mode only, Geist Mono, `#00ff88` accent, sharp corners | **no CSS, no tokens, no class changes** — zero style lines in the diff | — |

## Deviations

1. **The default-backend fallback is inlined, and it has NO unit test.** Per the PO ruling on @re1's finding: the extraction I first shipped was a one-caller abstraction the ticket forbids, so `src/lib/defaultBackend.js` and `server/defaultBackend.test.js` are both deleted and the fallback is a single expression over the shared `BACKENDS` constant (`src/components/SettingsPage.tsx:721-723`). A mirror test was explicitly rejected — this ticket spent three review rounds deleting that class of test, and a mirror here would be worse than none. **Stated openly: the grok-only and gemini-only acceptance criteria therefore have no automated coverage** and move to post-merge local-Mac verification by the PO, as the amended ticket records. The inlined expression structurally cannot express the old two-arm bug.

2. **Item 4's `~/.grok/bin` fallback in `isCliInstalled` has no test.** The helper isn't exported, no existing test covers that fallback list either, and it isn't in the ACs — so I left it uncovered rather than export a helper solely to assert it. Flagging rather than letting a reviewer discover it.

3. **The cap re-arm rationale at `server/pty-dispatcher.js:225-237` is corrected, not just extended.** It previously justified the residual injection with "safe because the cap is Claude-only and Claude queues input received mid-turn" — a claim that stopped being true the moment a second backend became eligible (@re1's finding). It now states what the behavior actually rests on and marks that property as **established for Claude but NOT verified for grok**, which is precisely what the ticket's post-merge PO step exercises. I did not substitute an unverified claim about grok's input handling for the old one.

4. **`capEligible` is newly exported** from `server/pty-dispatcher.js` (test-only, in the existing "Exported for testing" block). The AC names the predicate directly, and the fail-safe direction — an unstamped session stays ineligible — is not reachable from a repaint test.

## Implementation notes for review

**The three high-risk items, each specified from a measurement:**

1. **`capEligible` admits grok — required, not cosmetic (item 9).** Grok's TUI repaints continuously when idle, so the drain's 5 s idle gap never arrives and *every* mention to a grok agent takes the defer path. Without the cap the backend is inert in production while the whole suite stays green. `server/pty-dispatcher.js:134-138`, now a `CAP_ELIGIBLE_BACKENDS` set rather than `=== "claude"`; the stale #1010 comments that said "Claude-only *because* codex/gemini go quiet" are updated — still true of those two, no longer the full story.

2. **The `.grok/` exclude resolves the common dir against the agent cwd.** `server/index.js:543-570`. `path.resolve(agentCwd, out)`, never `path.join(out, …)`: agent cwds are linked worktrees where `<cwd>/.git` is a *file* (ENOTDIR) and a per-worktree exclude isn't honored — but an operator-configured cwd may be a plain clone, where `git rev-parse --git-common-dir` answers a relative `.git` and an unresolved write lands in the **server process's** repo while the token-bearing `.grok/` stays committable.

3. **The exclude write is append-if-absent.** The common-dir exclude is shared by all four worktrees and `buildAgentArgs` runs on every spawn *and* respawn, so a blind append grows it without bound.

**Fixture shape.** `server/agentArgs.grok.test.js` uses **both** required shapes, since either alone passes green against code that breaks in production: a **real linked worktree** (`git init` + commit + `git worktree add`; asserts `<cwd>/.git` really is a file, that the write lands in the *main* repo's `.git`, and that `git status --porcelain` doesn't list `.grok/`) **and a plain clone** (asserts the fixture premise that `--git-common-dir` returns a relative `.git`, then that the exclude lands under the agent's own repo). The plain-clone case `process.chdir()`s into a scratch dir with a `finally` restore — against the naive implementation it exists to catch, the write follows the *process* cwd, which under `npm test` is this checkout. Confirmed contained: under that mutation the stray `.git` appeared in the scratch dir, and the test asserts that too.

The `.grok/config.toml` is written `0600`, and the test checks the embedded token against `fileChat.validateShimToken` rather than pattern-matching it, so the security invariant is asserted rather than assumed. No grok binary is required by any test (#905 precedent).


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Epic Alignment: FAIL
PR #1024 is undocumented against the required template, so epic alignment cannot be evaluated.

### Checked (evidence)
- Structural gate: live REST PR body (`gh api repos/realproject7/quadwork/pulls/1024`) lacks filled `## EPIC Alignment` and `## Self-Verification` sections.
- UI template: the same body lacks the required `## Design Fidelity` table.
- Code/diff: SKIPPED — structural gate requires stopping before code review.
- CI: `gh pr checks 1024` → `test pass 1m4s`.

### Findings
- [blocking] Required PR-template sections are missing.
  - File: `PR #1024 description`
  - Why it fails: the review protocol requires filled Epic Alignment and Self-Verification sections, plus a Design Fidelity table for this UI PR, before code review.
  - Do instead: add all three sections with concrete, line-cited evidence for the current SHA, then request re-review.

### Decision
The missing documentation is a mandatory structural rejection. I will review the live diff only after the completed template is posted.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Epic Alignment: FAIL
The core Grok seams align with standalone #1023, but the added default-backend abstraction exceeds the ticket’s explicit no-new-abstraction boundary.

### Checked (evidence)
- Structural gate: PASS — live REST PR body contains filled `## EPIC Alignment`, `## Self-Verification`, and `## Design Fidelity` sections.
- Item 9: Grok is admitted at `server/pty-dispatcher.js:134-138`; Test 21 behaviorally covers repaint starvation.
- Exclusion: `server/index.js:543-563` resolves the common dir against cwd and is append-if-absent; linked-worktree and plain-clone tests are present at `server/agentArgs.grok.test.js:182-218`.
- Riskiest part: the token-bearing worktree config/exclude path; its required fixture shapes are present, but the remaining findings must be fixed.
- Kill-list: hits listed below.
- CI: `gh pr checks 1024` → `test pass 1m4s`.

### Findings
- [blocking] A Claude-only safety assertion is now false after Grok became cap-eligible.
  - File: `server/pty-dispatcher.js:225-231`
  - Why it fails: the code permits Grok at `:134-138`, but the re-arm rationale still says the cap is safe because it is Claude-only and Claude queues mid-turn input. It does not justify the same behavior for Grok.
  - Do instead: replace this stale rationale with an evidence-backed Grok-specific tradeoff, or state only the bounded starvation/cooldown policy without asserting Claude-only safety.
- [blocking] `firstInstalledBackend` is a new runtime helper with one production call site.
  - File: `src/lib/defaultBackend.js:23-25` (sole runtime use: `src/components/SettingsPage.tsx:723`)
  - Why it fails: #1023 permits only the new project-TOML mechanism and forbids abstraction drift; the review kill-list requires a one-call-site helper be inlined.
  - Do instead: inline the small `BACKENDS.find(...)` fallback in `SettingsPage.addProject`, remove `defaultBackend.js`, and rework its test to exercise the actual UI decision rather than retain a separate runtime module.

### Decision
The implementation substantially satisfies #1023, but these two findings are mandatory rejection conditions. Fix them and request a delta re-review.

project7-interns (approved):
## Verdict: APPROVE

### Epic Alignment: PASS
The `65add1f..b03aa60` delta follows amended standalone #1023: it removes the forbidden one-caller abstraction, documents the cap safety premise for both eligible backends, and fixes the cross-platform test fixture.

### Checked (evidence)
- Re-review range: `65add1fca0ce296fdcf3fa9118afedff163e3fc3..b03aa60a6b2e94c8531c61e88d9b52379f72d27b` via live GitHub compare.
- Mac fixture: `server/agentArgs.grok.test.js:52-60,64-90,193-204` canonicalizes the fixture root and both common-dir sides; all derived fixtures use the canonical root.
- Cap rationale: `server/pty-dispatcher.js:232-245` now gives the per-backend queueing rationale and cites the PO-verified Grok CLI docs; `injectIntoTerm` writes text before CR at `:355-364`.
- One-caller helper: removed; the bounded fallback is inline at `src/components/SettingsPage.tsx:721-723`, as #1023 now requires.
- Riskiest part of this diff: worktree/common-dir fixture portability; the canonical path comparisons directly address the macOS symlink failure while preserving the linked-worktree and plain-clone assertions.
- Kill-list: scanned new ranges — clean.
- CI: `gh pr checks 1024` → `test pass 1m9s`.

### Findings
- None.

### Decision
My prior findings are resolved at the current head, and the re-review delta introduces no new defect. Approved.

realproject7:
## @re2 APPROVE at `b03aa60a6b2e94c8531c61e88d9b52379f72d27b`

First full read of the PR, reviewed against the amended #1023. 10 files, +501/−28, zero dependency change, CI `test pass 1m9s`, OPEN/MERGEABLE. No findings.

I re-ran every check at this tip rather than carrying forward the `65add1f` results, since both `pty-dispatcher.js` and the fixture changed.

### The macOS fix — verified under the condition it guards, not just asserted

`os.tmpdir()` honors `TMPDIR`, so I reproduced the operator's macOS failure on Linux by pointing it at a symlinked directory:

| | symlinked TMPDIR | real TMPDIR |
|---|---|---|
| `65add1f` (pre-fix) | **32 passed, 1 failed** | 33 / 0 |
| `b03aa60` (post-fix) | **33 passed, 0 failed** | 33 / 0 |

Same single assertion, exactly as reported. I then isolated which change is load-bearing: reverting **only** `FIX_REAL` still passes, and reverting **only** the assertion-level realpath still passes — but reverting **both** reproduces `32 passed, 1 failed`. So the two changes are independently sufficient and jointly redundant. That is a stronger position than the one-line fix that was asked for.

Whole-file audit confirms `:187` was the only exposure: `tomlArgs[0]` compares `path.join(__dirname, …)` against the product's identical expression, the plain-clone premise compares against the literal `".git"`, and everything else is existence / containment / `git status`. No other new or changed test file references `os.tmpdir` at all.

### The inlined fallback — the AC lost its test, so I proved equivalence instead

Per the PO ruling the helper and its test are deleted, leaving the grok-only and gemini-only ACs with no automated coverage. I substituted an exhaustive equivalence check of the inlined expression against the deleted helper across `null` plus **all 16 install combinations** — **0 differences**. Branches: grok-only → `grok`, gemini-only → `gemini`, nothing installed → `claude`, `null` cliStatus → `claude`. I also confirmed the guarding condition (`configuredAvailable`, `SettingsPage.tsx:715`) is reached only when the configured backend is unavailable, so the new expression cannot change behavior for an operator whose `default_backend` is installed.

**Residual risk, stated plainly:** this pins the expression's semantics as transcribed, not the JSX call site. That is inherent to the ruling, not a defect in this PR.

### Mutation results — my own runs, not @dev's numbers

All ten guards discriminate. At `65add1f`: capEligible-grok **3**, `path.resolve` **3**, append-if-absent **1**, `--trust` **3**, injectMode grok **8**, PERMISSION_FLAGS grok **4**, grok `--model` **2**, grok catalog **5**, `.grok/config.toml` write **1**, first-installed fallback **4**. Re-run at `b03aa60` for the four that survived the delta: **3 / 3 / 1 / 3**. Baseline 0 failures; worktree clean after every revert.

### On the cap rationale

The rewrite is better than a reworded comment. Establishing queue-not-interrupt as a **per-backend obligation** — and saying that is why the predicate is an explicit set rather than "every backend that repaints" — is the part that will matter to whoever adds a fifth backend. I verified the half that is checkable in-tree: `injectIntoTerm` writes the flattened text at `pty-dispatcher.js:350` and its single `\r` only at `:356`, so the composer is never empty when the CR lands and an injection cannot reach grok's bare-Enter force-send path. The grok doc citation is correctly attributed to the PO rather than claimed first-hand.

### One gate observation, non-blocking

`.github/workflows/ci.yml:21-22` runs only `npm ci` and `npm test` — **not** `npm run build` and **not** `tsc --noEmit`. So "CI green" does not cover the build or the typecheck on a PR that touches TypeScript and two components. I ran both locally at this tip: `tsc --noEmit` exit 0, `npm run build` succeeds (9 routes). Worth knowing that the gate rests on reviewer-side verification for those two, not CI.

### Checked (evidence)

- `server/pty-dispatcher.js:137` `capEligible` via `CAP_ELIGIBLE_BACKENDS` (`:134`); fail-safe preserved — `Set.has(undefined)` is false, pinned by `pty-dispatcher.test.js` Test 22. Comments `:11-24` and `:143-146` updated as the ticket required; re-arm rationale `:230-245`.
- Test 21 (`pty-dispatcher.test.js:502-521`) is a behavioral repaint regression (not-before-cap → delivered-at-cap → exactly-once), not just a predicate assertion. Test 15 (`:302-318`, pre-existing) still supplies the required codex/gemini negative case under continuous output.
- `server/index.js:543` `excludeGrokFromGit` — `path.resolve(cwd, out)` at `:557`, append-if-absent at `:561`, trailing-newline guard at `:563`; non-repo cwd degrades quietly at `:552`.
- `server/index.js:570` `writeGrokProjectConfig` — hand-written TOML, `JSON.stringify` escaping at `:583`; `0600` via `writeSecureFile` (`config.js:166-170`), `0700` dir via `ensureSecureDir` (`config.js:159-163`); asserted at `agentArgs.grok.test.js:179`.
- `server/index.js:667-680` `project_toml` branch — `--trust` gated behind the `resolveAgentCwd` null guard; `PERMISSION_FLAGS.grok` `:441`, `/api/cli-status` `:264`, `~/.grok/bin` fallback `:249`.
- `src/lib/injectMode.js:31-38` grok → `project_toml`; heal path `server/routes.js:4054` consumes the same helper, so a command change to grok reconciles with no extra wiring.
- `src/components/SettingsPage.tsx:69-74` BACKENDS (rendered `:948`/`:1087`/`:1239`), `:385` cliStatus type, inlined fallback `:721-723`; `SetupWizard.tsx:189-193` BACKENDS (rendered `:823`), `:353`/`:796` filter tuples, single-installed branch `:354-357`.
- Suite 65/0/2 at this tip (one fewer file — the mirror test is deleted; baseline main 64/0/2); `tsc --noEmit` exit 0; `npm run build` 9 routes; `git diff --stat` on `package.json`/`package-lock.json` empty.
- Live: `gh pr view 1024` → head `b03aa60`, OPEN, MERGEABLE; `gh pr checks 1024` → `test pass 1m9s`. Re-read immediately before posting.

Issue #1060, opened 2026-08-30
[V2-23] Repository Delivery Candidate and single remote publication


realproject7:
## EPIC Context

- Parent EPIC: #1028
- V2 stage: integrated delivery
- Status: implementation and merge authorized by the operator on 2026-09-08; the remaining executor amendment below requires two independent contract reviews before its implementation.
- Purpose: compose accepted local task candidates into one repository-scoped remote PR so ordinary task/review corrections do not repeatedly consume GitHub Actions, merges, or deployments.

## Problem

Local-first review can avoid push churn inside one ticket, but a one-ticket/one-PR policy still creates one hosted CI run, merge, and potentially one automatic deployment for every small ticket. Conversely, one unbounded mega-PR would destroy reviewability, rollback isolation, and ticket traceability.

V2 introduces a bounded middle unit: the **Repository Delivery Candidate**. It is assembled only from accepted exact-SHA task candidates and is the sole ordinary remote publication unit for an integrated batch.

## Fixed delivery model

- Tickets remain durable product contracts.
- Work Tasks remain local execution/review checkpoints.
- One Delivery Candidate belongs to exactly one registered repository and one frozen batch cut.
- An integrated batch with work in two repositories creates two Delivery Candidates and two PRs, never one cross-repository PR.
- An isolated batch creates its own Delivery Candidate even when other work is ready.
- Ordinary task iterations do not push, open PRs, start hosted CI, merge, close issues, or deploy.

The design targets **one full local verification plan per final candidate SHA**. It does not claim that a genuine failure can be fixed without creating a new SHA and rerunning required checks.

## Identity and manifest

```text
DeliveryCandidateRef = {
  installation_id,
  project_id,
  repository_key,
  batch_number,
  batch_manifest_revision,
  cut,
  base_sha,
  delivery_manifest_sha256,
  attempt
}
```

The canonical Delivery Manifest records, in declared composition order:

- every included WorkItemRef and WorkTaskRef,
- ticket body and task revision digests,
- local task candidate SHA/tree identity,
- both terminal current-SHA task-review receipts,
- named local validation evidence,
- explicit exclusions/deferred tasks,
- pinned base and composition result SHA,
- repository key and canonical owner/repo resolved server-side.

The manifest digest is SHA-256 over a versioned canonical JSON serialization. Human titles, PR text, chat, labels, and branch names are not authority.

## Deterministic composition

Head requests composition only after the selected tasks are `staged`.

[...]

Issue #1059, opened 2026-08-30
[V2-22] Independent-first task review rounds and bounded reconciliation


realproject7:
## EPIC Context

- Parent EPIC: #1028
- V2 stage: task-review protocol
- Status: design only; implementation is not authorized until the operator explicitly starts V2 and the exact current body revision has two independent ticket-review approvals.
- Purpose: keep the defect-finding value of two independent reviewers while eliminating duplicate correction loops and Head/Dev waiting.

## Problem

V1 obtains valuable independent RE1 and RE2 review, but treats their free-form messages as two separate serial correction channels. Dev can stop for the first verdict, re-tip, invalidate the other review, and repeat. If the reviewers see each other's reasoning before making an initial judgment, independence is lost; if they never reconcile, duplicated or conflicting instructions create churn.

V2 therefore separates **independent discovery** from **shared reconciliation**. It does not turn reviewers into a consensus committee and does not let Head replace a missing reviewer.

### Phase A — sealed independent first pass

- RE1 and RE2 receive the same authenticated WorkTaskRef, task revision, base, exact candidate SHA, file boundary, ticket contract revision, and validation evidence.
- Each reviewer submits one role-bound structured verdict without seeing the other reviewer's findings or verdict.
- A first submission is sealed until both current assignments submit or the round is terminally cancelled.
- The submitting reviewer may read back only its own receipt while the round is sealed.
- Neither Primary Chat nor terminal output carries finding detail before release.

This preserves model diversity and prevents anchoring. It is not optional for an implementation task review.

### Phase B — bounded reconciliation

After both sealed submissions exist, the server atomically releases one combined review record to RE1, RE2, Dev, and Head.

- Exact agreement is accepted once, with both reviewer receipts retained.
- Reviewers explicitly mark overlaps or agreement; the server does not infer semantic equivalence.
- Unique non-conflicting findings are included in the consolidated correction set.
- A disagreement receives at most one focused exchange per reviewer, bound to the same candidate and finding identifiers.
- If the disagreement remains, Head selects the contract interpretation or trade-off and records reasons.
- Head cannot waive a security, data-loss, required-test, stale-SHA, or missing-reviewer gate; those remain blocking until corrected or explicitly returned to the operator.

The resulting correction set is delivered once to Dev. Reviewers may not start independent side conversations that create additional authoritative requests.

## Review round identity

```text
TaskReviewRoundRef = {
  installation_id,
  project_id,
  WorkTaskRef,
  task_revision,
[...]

Issue #1058, opened 2026-08-30
[V2-21] Nested ticket-task batch manifest and pipelined build/review lanes


realproject7:
## EPIC Context

- Parent EPIC: #1028
- V2 stage: execution-model foundation
- Status: design only; implementation is not authorized until the operator explicitly starts V2 and the exact current body revision has two independent ticket-review approvals.
- Purpose: replace ticket-serial execution with a bounded ticket-to-task pipeline while preserving GitHub Issues as the durable product contract.

## Problem

The current code-batch loop treats one GitHub ticket as the only execution unit: Dev implements it, waits while RE1 and RE2 review, fixes each round, and only after merge does Head assign the next ticket. That creates four avoidable costs:

1. Dev is idle while reviewers inspect a finished candidate.
2. A ticket with several separable changes cannot expose reviewable checkpoints without creating extra GitHub issues or PRs.
3. Current Batch cannot show the real nested progress inside a ticket.
4. A merged ticket is the only advance signal, so Head cannot pipeline the next independent slice while review of the previous slice is in flight.

V2 keeps the four fixed roles and one Dev lane. It does not add an agent pool or a generic workflow engine.

## Fixed ownership model

Three units are distinct and must not be conflated:

- **Ticket / WorkItemRef** — the GitHub Issue remains the durable product contract, acceptance criteria, operator gate, and long-term memory unit.
- **Work Task / WorkTaskRef** — a Head-authored, immutable execution slice inside one ticket. It is not a GitHub Issue and is never a remote publication by itself.
- **Delivery Candidate** — the repository-scoped integration and publication unit defined by the dedicated V2 delivery ticket.

A WorkTaskRef is installation- and repository-qualified:

```text
{
  installation_id,
  project_id,
  repository_key,
  work_item: { owner, repo, issue_number, issue_body_sha256 },
  task_key,
  task_revision
}
```

`task_key` is unique only within one WorkItemRef. `task_revision` is the lowercase SHA-256 of the server-canonical structured task contract. No title, chat text, bare issue number, branch name, or array position is identity.

## Batch manifest

Before assigning the first task, Head decomposes the selected tickets into one frozen **Batch Manifest**. The manifest is a bounded extension of the existing Active Batch authority, not a general task database.

Each task contract contains only:

- `task_key`
- `repository_key`
- concise goal
- explicit file boundary
- named validation
- zero or more earlier `WorkTaskRef` dependencies
- source `WorkItemRef` and issue-body revision
[...]

Issue #1053, opened 2026-08-29
[V2-20][P0] Capacity admission, session health, and loss-recovery circuit breaker


realproject7:
## EPIC context

Part of #1028. #1038 owns the Linux/VPS resource primitive: disk-backed temp, API/worker/control-plane cgroup separation, limits, and read-only resource facts. This ticket owns the application lifecycle that consumes that primitive. The split keeps host containment independently reviewable from queue/session recovery and prevents one oversized P0 PR.

## Goal

Make every agent spawn, restart, and recovery pass through one atomic admission path; expose truthful per-generation health; and stop repeated early-exit/OOM restart storms without losing the current assignment or touching worktree contents.

## Where, verified on `main` at `44d820c`

- `server/index.js:723-925` — PTY spawn/exit/session-map lifecycle.
- `server/index.js:948-1125` — manual start, stop, reset, and restart endpoints.
- `server/index.js:1840-1905` — terminal WebSocket connection currently auto-spawns a missing PTY and must become observation-only.
- `server/index.js:2199-2276` — watchdog liveness and current auto-respawn path.
- `server/index.js:2339-2415` — startup active-batch restoration.
- `server/self-heal.js` and `server/self-heal.test.js` — bounded breaker seam that must not remain a separate restart authority.
- `server/routes.js:3404-3680` and #1050 — active-batch truth consumed by startup/recovery admission.
- `server/mcp-operator/tools/read.js`, `server/mcp-operator/tools/agents.js`, and their tests — externally visible health and lifecycle results.

## Canonical lifecycle contract

Every attempt has a server-generated `operation_id` and immutable `generation_id`. A lifecycle result is one of:

```text
rejected | reserved | launch_failed | spawned | verified | exited | unresponsive | timed_out | resource_killed | stopped
```

Persist, per configured role, only the current generation/operation, expected assignment attempt plus issue-contract digest, breaker reason/count, last observation, and unresolved loss correlation in atomic mode-0600 `~/.quadwork/<project>/agent-lifecycle-state.json`. Reconcile it against live PTYs/#1038 scopes at server start; reservations are recomputed, never blindly replayed. Bound/prune terminal operations so this is not a task database.

[...]

PR #1062, opened 2026-08-31
[V2 M1–M4] foundations, Monitor, multi-repo setup, review-cycle core

realproject7:
## Summary

QuadWork V2 for EPIC #1028: one project spans multiple repositories, and the existing Head carries the full PO operating contract. No new agent roles, no visual redesign, no package major version, no publication.

## Exact candidate

- Branch `task/1031-v2-qualified-queue` at `06f8be4`, merge-base `main` at `44d820c`.
- **Draft on purpose.** Ready-for-review and merge are separate decisions; this push exists to put the candidate under hosted CI at its current SHA.

## EPIC Alignment

This PR delivers 12 of #1028's roadmap items in full and 9 partially. The EPIC's roadmap status column and each ticket body carry the per-ticket verdict; nothing here claims more than that.

| area | state |
|---|---|
| Canonical multi-repository schema, GitHub state, queue identity, lifecycle, Head protocol (#1029 #1030 #1031 #1033 #1034) | complete |
| Butler removed as a product and runtime slice (#1035) | complete — `git grep -il butler` over the tracked tree returns 0 |
| VPS containment primitive, environment bindings, Head reset continuity, CI evidence, reviewer verdict comments (#1038 #1045 #1047 #1049 #1051) | complete locally; #1038 still owes its disposable-staging proof |
| Live Active Batch authority (#1050) | complete |
| Right-rail dashboard (#1052) | implemented this wave; live browser matrix passed |
| Multi-repo setup, Monitor, Head control plane, Batch Requests, PR handoff, capacity/health, task pipeline, task review, integrated delivery (#1032 #1036 #1044 #1046 #1048 #1053 #1058 #1059 #1060) | partial; each ticket body names its own unmet requirements |
| Finishing audit and five-mode benchmark (#1037) | not started, by design — it runs after these merge |

`#1060` stops at a read-only publication plan. That is the intended operator boundary, not an omission: its tests assert the absence of `publish_delivery`.

## What changed since the previous review of this branch

The two blocking findings against `0e2a7e75` — an activation TOCTOU in `routes.js` and a deferred-wake receipt ordering window in `project-monitor.js` — are fixed and covered by tests that fail against the previous code.

Independent review of the resulting candidate then found the primary runtime paths were not actually reachable, and this branch fixes each:

- `mcp-head-control-shim.js` rejected every spec-conforming MCP `initialize` with `-32602`, so the whole #1044 surface was unreachable from any real client.
- `POST /api/work-task-candidate` pre-validated an eight-field `work_task_ref` with a four-field resolver, so every Dev submission returned 409.
- A `changes_requested` verdict had no correction producer and was a permanent dead end.
- `head-control-plane.js` denied every command after 64 calls, including read-only ones, until archive or restart.
- Respawn and startup seeding filtered projects by scalar `working_dir`, so V2 projects were never restored after a restart.
- `assigned-work-task-candidate.js` accepted only 64-hex object ids while every sibling accepted 40 or 64, rejecting all SHA-1 repositories.
- `index.js` injected a `Date` clock where the manifest required a string, so `freeze_batch_manifest` could never succeed through the real route.
- `GitHubPanel.tsx` keyed issue and PR rows by bare number, colliding across two repositories in exactly the way #1028 forbids.

A batch could also only ever be frozen once per project; retirement now lets a successor manifest freeze while the retired record stays readable as provenance.

## Self-Verification

At `06f8be4`, all local:

- `npm test` — 186 passed, 0 failed, 2 pre-existing Jest-style skips. Stable across three consecutive isolated runs.
- `npx tsc --noEmit` — clean.
- `npm run build` — exit 0.
- `node server/pack-smoke.test.js` — 112 shipped Node files, all relative requires resolve inside the tarball.
- Live browser matrix for #1052 at 1440x900 and 900x900 against a running instance with an isolated `HOME`: panel order and the three first-visit defaults; collapse leaving the grid columns byte-identical at `688px 4px 684px`; every request during a toggle a GET, so collapse reaches no backend mutation; reload persistence; no horizontal overflow at either width; `aria-expanded`, `aria-controls`, and an accessible name on a 48x24 control.
- Legacy upgrade proof: a pre-V2 scalar-repo config projects to `repositories[{key:"primary"}]` while the on-disk config stays byte-identical, so the migration is a read-time projection and a downgrade is safe.
- Startup spawned zero agent PTYs from a history-only snapshot.

**These are local results. They are not hosted CI, independent VPS proof, or a merge gate.**

## Review focus

- Retirement's CAS plus rename sequence under concurrency and crash.
- Authority on the three new Head-control actions: `retire_batch`, `queue_local_correction`, `read_propagation_stop`.
- That `read_propagation_stop` leaks no finding, verdict, or reviewer identity before release.
- Idempotency replay after oldest-first eviction.

## Known gaps, stated rather than hidden

- #1038 and #1053 need a disposable staging Linux host; the harness enforces a `DISPOSABLE-STAGING:<machine-id>` acknowledgement and refuses to run on production. That gate is open.
- #1037's audit, live matrix, and five-mode benchmark have not started.
- Follow-ups from the review waves are filed as #1063-#1071 rather than folded in here, so this candidate stays reviewable.


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Epic Alignment: NOT VERIFIED
The required PR-body evidence is missing, so alignment cannot be evaluated.

### Checked (evidence)
- Structural gate: live PR #1062 body contains `## Summary`, `## Exact candidate`, `## Verification`, and `## Review focus`, but no filled `## EPIC Alignment` or `## Self-Verification`.
- Exact tip: `origin/task/1031-v2-qualified-queue` resolves to `aef8fe157f606e564232206152ca865846bba038`.
- CI: live commit check `test` is completed with conclusion `failure`.

### Findings
- [blocking] Mandatory PR template sections are absent.
  - File: PR #1062 body
  - Why it fails: without a filled `## EPIC Alignment` and `## Self-Verification`, the review cannot verify the claimed contract or test evidence.
  - Do instead: add both required, evidence-backed sections for this exact SHA, then request re-review.

### Decision
The structural gate requires request changes before code review. The failed live test check is independently unresolved.

project7-interns (approved):
## Verdict: APPROVE

### Epic Alignment: PASS
The exact-SHA delta preserves #1028’s explicit fail-closed Linux containment boundary while proving lifecycle behavior without a production bypass.

### Checked (evidence)
- Exact delta: `aef8fe157f606e564232206152ca865846bba038..4309fffdba0ffb148ba261bff4f5a4c952a30d82` is 8 files (+293/-47); `origin/main` merge-base is `44d820c`.
- Production authority: `server/index.js:621-655,1298-1319` obtains containment only from a private fixture map; ordinary route/config/env calls retain `containedLaunch: false`. `server/agent-lifecycle.js:216-226` therefore rejects Linux production starts as `containment_unavailable`.
- Fixture isolation: `server/index.js:708-710` admits only WeakSet-capability fixture identities; the sole installer is exposed through the test seam at `:3914`.
- Non-vacuous lifecycle coverage: `server/projectRuntimeLifecycle.test.js:453-513` separately asserts uncontained Linux rejection, steady archive precedence, and archive during entered async preparation with zero PTY; `server/shutdownCleanup.test.js:90-138` proves normal API rejection then contained-fixture shutdown reaping.
- CI receipts: `server/ci-less-evidence.js:251-265,332-356` serializes per-project resolve/validate/write/readback and rechecks admission before `upsert`; `server/ci-less-evidence.test.js:193-277` covers concurrent identities and rejected-final-admission no-write.
- Required neutral/skipped checks: `server/ci-evidence-policy.js:316-325` maps terminal `not_success` to Head-owned `missing_required`; tests cover skipped/neutral ownership at `server/ci-evidence-policy.test.js:138-145`.
- Riskiest part: the test fixture is an authority-like seam, but it is neither route/config/env-controlled nor accepted without the private WeakSet capability.
- Kill-list: scanned delta; clean.
- CI: live REST check-runs for this SHA → `test` `completed/success`.

### Decision
All prior findings in my correctness/security scope are resolved at the new SHA; the old verdict was not carried forward.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Epic Alignment: NOT VERIFIED
The required evidence section is absent from this exact-SHA PR body.

### Checked (evidence)
- Structural gate: live PR #1062 body at `0e2a7e752c3b6e0c59485108482164eb24de7a53` has `## Summary`, `## Self-Verification`, `## Exact candidate`, and `## Review focus`, but no `## EPIC Alignment`.
- Exact tip: `origin/task/1031-v2-qualified-queue` resolves to `0e2a7e752c3b6e0c59485108482164eb24de7a53`; merge-base is `origin/main` `44d820c`.
- CI: live check-runs reports `test` `completed/success`.

### Findings
- [blocking] Mandatory `## EPIC Alignment` section is absent.
  - File: PR #1062 body
  - Why it fails: the milestone diff cannot be judged against the #1028 architecture and contract without the required, evidence-backed alignment declaration.
  - Do instead: add a filled `## EPIC Alignment` section tied to this exact SHA, then request a new exact-SHA review.

### Decision
The mandatory structural gate requires request changes before code review; no prior-SHA verdict carries forward.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Epic Alignment: FAIL
#1032 requires activation/provisioning to fail closed across its authority and TOCTOU boundaries; the final activation commit does not revalidate repository access.

### Checked (evidence)
- Structural gate: the live body at `0e2a7e752c3b6e0c59485108482164eb24de7a53` now contains filled `## EPIC Alignment` and `## Self-Verification`.
- Exact tip: `origin/task/1031-v2-qualified-queue` and live PR head resolve to `0e2a7e752c3b6e0c59485108482164eb24de7a53`; merge-base is `origin/main` `44d820c`.
- Monitor: persisted enabled-only delivery and generation idempotence are implemented at `server/project-monitor.js:174-231` and `server/trusted-event-transport.js:145-252`; no route invokes the controller.
- Retired slice: no remaining server/bin/src/template runtime Butler reference was found after the deletion; terminal attachment remains independently guarded.
- Review-cycle M1: exact target digest/receipt checks and retained invalidated/terminal records are implemented at `server/review-cycle.js:652-707,772-837`; no production caller instantiates its planner.
- CI: live check-runs → `test` `completed/success`.

### Findings
- [blocking] Activation has an unchecked repository-access TOCTOU window.
  - File: `server/routes.js:5669-5674,5676-5700`
  - Why it fails: `verifyV2RepositoryAccess()` runs before `provisionV2Repositories()`, which can clone/create and revalidate eight worktrees. The commit rechecks legacy state, sessions, and ownership, but never access. If repository write access is revoked during provisioning, it still persists activated `repositories[]`, contrary to the required fail-closed activation/access boundary. Existing tests cover only pre-provision access rejection (`server/routes.v2Setup.test.js:158-172`) and do not exercise this race.
  - Do instead: re-run canonical identity/write-access verification after provisioning and immediately before `commitV2Configuration`; on failure, return 409 without activation. Add a test that changes the GitHub access result during provisioning and asserts the config remains pre-activation.

### Decision
The metadata gate is now satisfied, but the activation authority race is unresolved. No older SHA approval is carried forward.

realproject7:
**@re2 REQUEST CHANGES at exact SHA `0e2a7e752c3b6e0c59485108482164eb24de7a53`** — architecture, scope, races, test adequacy. No prior approval carried forward.

### Checked (evidence)

**Blocking — I independently confirm @re1's #1032 access-recheck gap, and add the constraint that shapes the fix.**
`server/routes.js:5669` awaits `verifyV2RepositoryAccess(repositories)` before provisioning, but the commit callback at `:5676-5700` re-validates only `firstActivationLegacyGuard`, `existingV2TopologyGuard`, and `configuredRepositoryOwnership` — access is absent.

The asymmetry is real rather than a uniform design choice, and I mutation-tested it: deleting the commit-time ownership recheck (`routes.js:5683-5684`) fails `server/routes.v2Setup.test.js`, and deleting the session recheck (`:5681-5682`) fails it too. Both surviving rechecks are non-vacuously pinned; access is the lone unguarded member of that set.

**Constraint on the remediation:** the recheck cannot be moved inside the atomic transaction. `server/config.js:948-955` calls `mutator(cfg)` **synchronously** inside `withConfigWriteLock`, while `verifyV2RepositoryAccess` is `async` (`routes.js:5507`). Placing it immediately before `commitV2Configuration(...)` narrows the window from the duration of cloning/provisioning to the duration of the commit — it does not close it. A genuinely closed window would need access verification inside the lock, which a network call cannot satisfy. I'd accept the narrowed window plus the provision-time revocation test @re1 asked for, stated as a known limit rather than as proof.

**Severity, for scheduling:** the harm is a config entry recording repositories the operator has lost access to. The entry grants no access, and downstream `gh` operations fail closed on their own, so this is stale-state rather than privilege escalation.

**#1036 Monitor — verified sound.** Head-only: `server/trusted-event-transport.js:13` freezes `RECIPIENTS = ["head"]` with no generic send API. Persisted opt-in: `server/project-monitor.js` `start()` is the sole path to `enabled`, and `unarchive()` deliberately restores only to `suspended`.

No duplication across generation: `correlation_id` is `sha256{version, kind, anchors}` with `event_generation` inside anchors; `phase === "woken"` short-circuits; a deferred wake is blocked for the same recipient generation but intentionally re-eligible for a new one.

No loss: `_deliver` persists a `recorded` delivery before any chat/PTY action, and every failure mode returns `{ok:false}` instead of throwing, so the record stays in `deliveries` for `resume()`. I specifically probed whether `_onDeadline`'s delete-before-publish ordering could drop a wake via a publish-time throw — it cannot, because anchors are structurally guaranteed to satisfy the transport: the policy's `IDENTIFIER_RE` (`project-monitor-policy.js:30`) is byte-identical to the transport's `ID_RE` (`trusted-event-transport.js:14`), `normalizeAssignment` validates every copied field through it, `event_generation` and `threshold_generation` are server-derived hashes, and `condition()` rejects unknown kinds at construction.

**#1035 — verified sound.** Zero non-test residue; `/ws/butler` is gone with only `/ws/terminal` remaining (`server/index.js:3000`) and `isAllowedWsOrigin`/`tokenMatches` intact (`:3007`, `:3012`). `server/config.js:134-135` strips the retired field on read, and `server/retiredGlobalAgentMigration.test.js:102-116` asserts it is absent from both loaded and saved config, cannot be re-persisted by an ordinary save, migrates through one atomic rename, and leaves the former cwd never read/enumerated/created/moved/removed with its marker byte-identical.

**#1048 — verified sound.** Both `delete` calls (`server/review-cycle.js:696`, `:834`) remove only `document.slots` — the active-slot index — never `document.cycles`. `server/review-cycle.test.js:275-283` asserts terminal and invalidated cycles survive a reload. Mutation: adding `delete document.cycles[...]` to `markTerminal` fails the suite, so retention is pinned. No dispatcher: the module requires only crypto/fs/os/path plus two pure validators, and `eventPlan` describes an event rather than emitting one.

**#1032 fail-closed properties beyond the gap above.** Unverifiable `origin/HEAD` yields `default_branch_unavailable` (`server/repository-provisioning.js:204`) with no fallback guess at main/master; base identity is triple-verified via `--show-toplevel`, `remote get-url origin`, and `rev-parse --verify HEAD` (`:186-197`); `installation_id_rotation_forbidden` (`server/config.js:993-998`) blocks deletion or replacement; already-activated installations stay array-only so legacy scalars are rejected rather than silently stripped.

**Test adequacy.** 122 passed, 0 failed, 2 skipped on Linux, reproduced locally.

### Non-blocking

- **Crash window in the monitor deadline path.** `project-monitor.js` `_onDeadline` deletes the condition from `unresolved` and saves *before* the transport records the delivery. A crash between those two writes loses the wake. This is the one place the codebase's otherwise-consistent record-before-act ordering is inverted; reversing it would close the window.
- **Reviewability.** `server/config.js:28` builds the retired key as `["but","ler"].join("")`, so a plain `grep -i butler` over the source returns zero hits even though the config key is still literally that string. I nearly recorded "clean removal" from that grep alone. Functionally correct, but it defeats the exact audit a reviewer or maintainer would run — a plain literal with a comment would be safer.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Epic Alignment: FAIL
The Monitor portion does not meet #1036’s durable deferred-wake contract: a deadline wake can be lost across a process crash.

### Checked (evidence)
- Exact tip: live PR head remains `0e2a7e752c3b6e0c59485108482164eb24de7a53`; live `test` is `completed/success`.
- Monitor delivery normally uses record-before-act at `server/trusted-event-transport.js:145-179`, preserving a `recorded` receipt before chat/PTY work.
- Resume only processes persisted delivery records at `server/trusted-event-transport.js:239-252`; unchanged observations are a no-op at `server/project-monitor.js:181-185`.

### Findings
- [blocking] Deadline handling deletes durable wake authority before creating its durable delivery receipt.
  - File: `server/project-monitor.js:93-100`
  - Why it fails: `_onDeadline()` deletes and saves `unresolved[conditionKey]` at `:93-95`, then only afterwards calls `transport.publish()` at `:96-100`. A crash between those durable operations leaves neither the deadline nor a delivery in monitor state. After restart, `resumePending()` has nothing to replay, and the unchanged fact set returns the no-op path at `:181-185`; the wake is permanently lost. This violates the requested generation-safe, no-loss deferred-wake boundary.
  - Do instead: persist a correlation-keyed delivery receipt (or a pending-to-deliver deadline state) in the same state transition that retires the condition, then have transport resume that durable receipt. Add a crash-window/reload test proving the event is delivered once after restart, not dropped or duplicated.

### Decision
This is independent of the activation TOCTOU finding and is a second blocking durability defect at this SHA.

realproject7:
**@re2 REQUEST CHANGES at exact SHA `06f8be46883dd80576b012ea5f5680ec9d14e1a0`** — architecture, scope, races, test adequacy. My `0e2a7e75` verdict is discarded. Head SHA re-checked immediately before posting.

### Checked (evidence)

**BLOCKING — the new stores' lock-identity guard is defeated by inode reuse. CI is right; a local pass is a false negative.**

Required check `test` is `completed/failure` at this tip (run 33713405865): `183 passed, 3 failed, 2 skipped`, with `server/batch-request-state-store.test.js`, `server/delivery-candidate-store.test.js`, and `server/work-task-pipeline-store.test.js` each failing `AssertionError: Missing expected exception`.

All three suites **pass** on my Linux box and **fail** in CI, so I root-caused the divergence rather than dismissing either result. `sameFile` compares only `dev` + `ino` (`server/batch-request-state-store.js:126-129`, mirrored at `delivery-candidate-store.js:337` and `work-task-pipeline-store.js:333`). The failing case is the one asserting a writer "fails closed instead of unlinking a different writer's lock inode": the test unlinks the lock and recreates it as a *different* writer's lock, then expects `*_lock_release_failed`.

Whether that guard fires depends entirely on whether the filesystem recycles the just-freed inode. Measured on this host:

- `/tmp` is **tmpfs** — inode reused **0/5** trials → replacement gets a new inode → guard fires → tests pass
- `/home/quadwork` is **disk-backed (ext)** — inode reused **5/5** trials → replacement reuses the inode → `sameFile` returns true → guard never fires → tests fail

The suites call `os.tmpdir()`, which is `/tmp` here. Pointing `TMPDIR` at the disk-backed filesystem reproduces all three CI failures locally and exactly: `batch-request-state-store exit=1`, `delivery-candidate-store exit=1`, `work-task-pipeline-store exit=1`, each `Missing expected exception`.

So this is not CI flake and not a test bug — the **product** defect is real and CI caught it. On any disk-backed filesystem, which is both CI and the VPS deployment target, one writer can unlink another writer's lock, breaking the mutual exclusion these three stores exist to provide. A dev whose `/tmp` is tmpfs will see green locally every time. Recommended fix: pin lock identity to something the filesystem cannot recycle — an `O_EXCL` create holding the descriptor plus a per-writer nonce written into the lock and re-read before release — rather than `dev`+`ino` alone.

**Scope — the PR grew by another 181 files since the tip I last reviewed.** `0e2a7e75..06f8be46` is 181 files, +34,586/−896; the PR now totals 298 files, +81,290/−4,244. That is well beyond the stated M1–M4 milestone content and is not independently reviewable as one unit. I reviewed the failing-store surface and re-verified my prior findings; I did not audit the remaining ~34k new lines, and I would not treat any approval of this SHA as coverage of them. Recommend splitting before further review effort is spent.

**Both of my prior blockers are fixed — credited.**
- **#1032 access recheck:** `server/routes.js:6429` now re-verifies with `postProvisionAccess = await verifyV2RepositoryAccess(repositories)` after provisioning and before commit, outside the config lock — the shape I recommended, and correctly not inside the synchronous mutator.
- **#1036 deadline retirement:** `server/project-monitor.js:94-107` now calls `transport.recordAll(...)` with the already-retired `next` state, so the correlation-keyed receipt and the retirement become durable together before any chat/PTY side effect. The `delete` at `:95` mutates only the in-memory clone and is never persisted alone, so the crash window I flagged is closed, and the stated invariant (unretired deadline survives a lost write; `resumePending` owns exactly-once recovery) holds.

**Verdict: REQUEST CHANGES.** One blocking defect, plus the scope concern above. Review-only: no edits, push, merge, queue advance, issue action, or release; scratchpad worktree removed.

realproject7:
**@re2 APPROVE at exact SHA `39263d211c44aadebba1b7c11a41478134c479b7`** — architecture, scope, races, test adequacy. My `06f8be46` verdict is discarded. Head and CI re-checked immediately before posting; `origin/main` is `44d820c` and is still the merge-base, so no rebase drift.

### Checked (evidence)

**My blocking finding is fixed, and the fix is exactly the shape the three of us converged on.** The delta `06f8be46..39263d21` is tight — 12 files, +263/−17.

`lockToken()` mints a per-writer nonce (`${process.pid}.${crypto.randomBytes(16).toString("hex")}`), written into the lock body at acquisition via `fs.writeFileSync(descriptor, token, "utf8")` and threaded through as `{descriptor, lockPath, stat, token}`. Both checks read it **by path** — `fs.readFileSync(lockPath, "utf8")` — not through the held descriptor, which was the specific correction I raised: reading via the fd would return the old inode's content and reintroduce the blind spot.

Critically, **both** unlink sites are gated, not just the one whose test was failing: `releaseLock` now requires `sameFile(...) && token match`, and the acquire error path requires `sameFile(...) && fs.readFileSync(lockPath, "utf8") === token` before unlinking. A fix landing only in `releaseLock` would have left the second site open.

**The fix was applied to six stores, not the three that were red** — `batch-request-state-store`, `delivery-candidate-store`, `work-task-pipeline-store`, plus `head-control-audit-store`, `task-review-round-store`, and `head-control-work-task-domain`, which carried the same flawed predicate without failing tests. Finding the latent instances rather than only the reported ones is the right instinct.

**The new tests are filesystem-independent by construction, which is what makes this durable.** Rather than depending on real inode recycling, the stub forces it: `stats.dev = original.dev; stats.ino = original.ino`. That exercises the recycled-inode case even on tmpfs, where reuse never happens — closing the exact trap that let this ship green on a tmpfs box.

I verified that empirically rather than trusting the shape:
- All six suites pass on **tmpfs** (`/tmp`) and with **`TMPDIR` on ext4**, so the previously divergent environments now agree.
- **Mutation:** dropping the token comparison from `releaseLock` fails each of the six suites **on tmpfs** — `batch-request-state-store`, `delivery-candidate-store`, `work-task-pipeline-store`, `head-control-audit-store`, `task-review-round-store`, `head-control-work-task-domain` all exit 1. So every guard is non-vacuously pinned, and pinned without depending on the underlying filesystem.

**Full suite with `TMPDIR` on ext4 (matching CI and the VPS): 186 passed, 0 failed, 2 skipped.** `tsc --noEmit` passes; CI does not run it. Required check `test` is `completed/success` at this exact SHA.

### Coverage limit on this approval — please read it as scoped

The delta since my last review is only the lock fix, so my earlier scope observation is unchanged: this PR is **298 files, +81,536/−4,244**, of which roughly 34k lines arrived in the jump to `06f8be46` and I have **not** audited them. This approval covers the lock-identity fix I verified above, plus the areas I reviewed in earlier rounds (#1032 access recheck, #1036 deadline retirement, #1035 removal, #1048 store authority, the attach-only terminal WS, and the CI-evidence policy). It is not a statement about the unaudited remainder, and I would not want it counted as one at the merge gate.

I raise that as a scoping caveat rather than a blocker: every defect I actually found is fixed and mutation-verified, and I have no open finding at this SHA.

